PR #227, opened 2016-04-15
Added __iter__ to Function, BasicBlock

withzombies:
`Function` and `BasicBlock` have `__iter__` support now, allowing more pythonic (and LLVM IR-like) iteration over functions.

```python
for func in bv.functions:
    print repr(func)
    for block in func.low_level_il:
        print "\t{0}".format(block)

        for insn in block:
            print "\t\t{0}".format(insn)
```

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -1932,7 +1932,7 @@ class Function:
 			return result
 		elif name == "low_level_il":
 			return LowLevelILFunction(self.arch, core.BNNewLowLevelILFunctionReference(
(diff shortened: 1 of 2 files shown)


withzombies:
I'd also like to discuss adding an `MCInstruction` class (or equivalent). Right now this just returns the instruction text if you iterate over a regular BasicBlock. It should return a `MCInstruction` class instance, similar to how an instance of `ILInstruction` is returned when iterating over an `ILBasicBlock`.

D0ntPanic:
Merged into dev after integrating into new IL basic block API

PR #477, opened 2016-09-16
Add the color support to export_svg.py

rootbsd:
I add the support of the built-in color of the blocks.


psifertex:
Thanks! I'll probably add in an extra line to default to the old color scheme in the instance where the lookup fails, but it looks good.

psifertex:
Heh, nevermind I see the default now. :-)

PR #556, opened 2016-11-22
documenting Type.int and Type.function

lucasduffey:


lucasduffey:
still says changes requested but pretty sure it's fixed

psifertex:
Thanks!

PR #284, opened 2016-06-06
Makefile for building binaryninja-api and example c++ plugin

bambu:
Here is a working makefile to build the `binaryninja-api`. The makefile only works for OSX and Linux and it builds an archive (.a) and library file (.dylib / .so) so that binaryninja-api may be statically or dynamically linked.  

Also included is the `breakpoint.py` sample plugin converted over to `c++` and a Makefile to build that. If left in that directory it should build just fine after `binaryninja-api` has been built. 
### Note
1. If the example doesn't build. Open the makefile and make sure that the first few variables match your setup. 
2. I do not know of a way to know where Binary Ninja is installed in Linux. It assumes that it is at `$(HOME)/binaryninja/` so this may need editing. 
3. This makefile links against `libbinaryninjacore`. As of this PR only the **Dev** branch will compile with the latest beta (1.0.295)

PR #573, opened 2016-12-06
CMake files for building binaryninja-api and example c++ plugin

jhurliman:
Based on https://github.com/Vector35/binaryninja-api/pull/284, this swaps out Makefiles for CMake files to make configuration of library directories and target folders a bit easier (along with IDE support). This should in theory work on OSX, Linux, and Windows but has only been tested on OSX with CMake 3.7.1.


lwerdna:
Please contact me (@andrewl) on slack, I sent you a message.

PR #589, opened 2016-12-27
Making exported svg pages have graphs closer to the original

ColdHeat:
This is the new look:

image


psifertex:
Thanks for the PR -- I fiddled with it when I first wrote it, but could never get the constants right.

PR #616, opened 2017-02-11
Function.get_low_level_il_at and Function.get_lifted_il_at return a LowLevelILInstruction

joshwatson:
Currently `Function.get_low_level_il_at` and `Function.get_lifted_il_at` return the instruction index of the corresponding IL instruction at a virtual address instead of a `LowLevelILInstruction` object. In order to get the `LowLevelILInstruction` object, one must do something like `current_function.low_level_il[current_function.get_low_level_il_at(here)]` which is kind of annoying and verbose.

With this PR, users would instead be able to use `current_function.get_low_level_il_at(here)` which is clear and concise. Further, if the user does want the instruction index, it is still available as `LowLevelILInstruction.instr_index`, so no information is lost.

### DIFF
--- a/python/function.py
+++ b/python/function.py
@@ -363,20 +363,20 @@ class Function(object):
 
 	def get_low_level_il_at(self, addr, arch=None):
 		"""
-		``get_low_level_il_at`` gets the LowLevelIL instruction address corresponding to the given virtual address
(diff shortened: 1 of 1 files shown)


joshwatson:
Looks like this change was made without using the PR, so I'm going to close this issue.

PR #636, opened 2017-03-07
Add flags parameter to LLIL store

alexforencich:
Add flags parameter to LLIL store operation


plafosse:
Thanks for this. This issue is fixed in latest dev. I don't think we ever got a CLA otherwise I would have just accepted the PR.

PR #668, opened 2017-04-26
Added __hash__ to Function.

joshwatson:
As it stands, `Function` objects use the default `__hash__` method, which is just a wrapper for `hash(x)`, where `x` is the object. Since this uses the `id` or something similar, two `Function` objects that have the same backing function are seen as unique when added to a `set` or used as a `dict` key.

My suggested modification (which could also be applied to other objects as well) is to implement the `__hash__` method, hashing the `self.start`, `self.arch.name`, and `self.platform.name`, which should be sufficient to dedupe functions.

However, if you'd prefer, I think that `hash(id(self.handle.contents))` may also be sufficient.

### DIFF
--- a/python/function.py
+++ b/python/function.py
@@ -254,6 +254,9 @@ class Function(object):
 			return True
 		return ctypes.addressof(self.handle.contents) != ctypes.addressof(value.handle.contents)
 
+	def __hash__(self):
+		return hash((self.start, self.arch.name, self.platform.name))
+
 	@classmethod
 	def _unregister(cls, func):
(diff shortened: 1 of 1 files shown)


psifertex:
Nice suggestion. Thinking of other things that could use a similar treatment, basic block is the obvious one, not sure what else.

joshwatson:
Had to recommit, I messed up the tabs vs. spaces somehow

joshwatson:
Adding a `__hash__` to `function.Variable` would be very useful as well. I think that one could just return `hash(self.identifier)` and work fine

PR #696, opened 2017-05-18
BinaryView doc string fixes

bambu:
Noticed `create_database` had the wrong function name in the docstring. Was able to grep a few more occurrences.


psifertex:
Thanks! Looks good.

PR #721, opened 2017-06-29
A LLIL C++ API Parser example with in depth build notes

robert-yates:
A small c++ example that would of helped me greatly if i would of had it when i first got into the api.
This is my first time using github etc, hope its all fine :)

### DIFF
--- a/examples/llil_parser/src/LowLevel_IL_Parser.cpp
+++ b/examples/llil_parser/src/LowLevel_IL_Parser.cpp
@@ -1,0 +1,440 @@
+/*
+LLIL Parser - Binary Ninja C++ API Sample
+ - Robert Yates - 22/JUN/17
+ */
+
+#include "LowLevel_IL_Parser.h"
+#include <iostream>
+
+int main(int argc, char* argv[])
+{
+
+	try
+	{
+		ShowBanner();
+
+
+		if (argc != 2)
+		{
+			printf("Usage: %s <input file>", argv[0]);
+			exit(-1);
+		}
+
+		std::string inputName = argv[1];
+
+		SetBundledPluginDirectory(get_plugins_directory());
+		InitCorePlugins();
+		InitUserPlugins();
+
+		auto bd = BinaryData(new FileMetadata(), inputName.c_str());
+		BinaryView *bv;
+
+		for (auto type : BinaryViewType::GetViewTypes())
+		{
+			if (type->IsTypeValidForData(&bd) && type->GetName() != "Raw")
+			{
+				bv = type->Create(&bd);
+				break;
+			}
+		}
+
+		printf("[i] Starting analysis\n");
+		bv->UpdateAnalysis();
+
+		while (bv->GetAnalysisProgress().state != IdleState);
+		printf("[i] Analysis done\n");
+
+		LlilParser myParser(bv);
+		myParser.decodeWholeFunction(bv->GetAnalysisFunctionList()[0]);
+
+		/*
+		// Show Single LLIL in function x at index x
+		myParser.decodeIndexInFunction(0x407930, 0);
+		
+		// Decode a whole function by address
+		myParser.decodeWholeFunction(0x407930);
+		
+		// Decode all functions
+		for (const auto& f : bv->GetAnalysisFunctionList())
+		{
+			// Decode a whole function by BinaryNinja::Function object
+			myParser.decodeWholeFunction(f);
+		}
+		*/
+
+	}
+	catch (const std::exception& e)
+	{
+		printf("An Exception Occured: %s\n", e.what());
+	}
+
+	printf("[i] Finished\n");
+}
+
+
+
+LlilParser::LlilParser(BinaryView *bv)
+	: m_bv(bv)
+{
+	m_currentFunction.clear();
+	m_tabs = 0;
+	m_currentInstructionId = 0;
+}
+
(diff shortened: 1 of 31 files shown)


examples/llil_parser/src/LowLevel_IL_Parser.cpp
```diff
@@ -0,0 +1,433 @@
...
+		{
+			if (type->IsTypeValidForData(&bd) && type->GetName() != "Raw")
+			{
+				bv = type->Create(&bd);
+				break;
+			}
+		}
+
+		printf("[i] Starting analysis\n");
+		bv->UpdateAnalysis();
```
plafosse: Please fix this, you copied some older code that hasn't been updated, a fact that we just noticed today. Should be bv->UpdateAnalysisAndWait(); and drop the spin loop.

examples/llil_parser/src/LowLevel_IL_Parser.cpp
```diff
@@ -0,0 +1,433 @@
...
+	printf ("|    |   |    |   |   |    |     \\______   \\_____ _______  ______ ___________ \n");
+	printf ("|    |   |    |   |   |    |      |     ___/\\__  \\\\_  __ \\/  ___// __ \\_  __ \\\n");
+	printf ("|    |___|    |___|   |    |___   |    |     / __ \\|  | \\/\\___ \\\\  ___/|  | \\/\n");
+	printf ("|_______ \\_______ \\___|_______ \\  |____|    (____  /__|  /____  >\\___  >__|   \n");
+	printf ("        \\/       \\/           \\/                 \\/           \\/     \\/    \n");
+	printf("====================================================================================\n\n");
+
+}
+
+#ifndef _WIN32
```
plafosse: Should use BinaryNinja::GetUserPluginDirectory() instead

psifertex:
Also, if you can please fill out a CLA and email it to us as described under https://github.com/Vector35/binaryninja-api#contributing that would be appreciated!

examples/llil_parser/src/LowLevel_IL_Parser.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+	BinaryNinja::Function *function = m_currentFunction[0];
+	BinaryNinja::Ref<BinaryNinja::LowLevelILFunction> llil = function->GetLowLevelIL();
+
+	
+
+	for (size_t i = 0; i < llil->GetInstructionCount(); i++)
+	{
+		m_currentInstructionId = i;
+		BNLowLevelILInstruction currentInstruction = llil->operator[](llil->GetIndexForInstruction(i));
```
plafosse: Perhaps a little cleaner than the above:
`BNLowLevelILInstruction currentInstruction = (*llil)[llil->GetIndexForInstruction(i)];`

examples/llil_parser/src/LowLevel_IL_Parser.cpp
robert-yates: seems like BinaryNinja::GetUserPluginDirectory() returns C:\Users\Development\AppData\Roaming\Binary Ninja\plugins which
is an empty directory instead of C:\Program Files\Vector35\BinaryNinja\plugins which has the required platform dlls, if i use that one i get 0 functions
inside

examples/llil_parser/src/LowLevel_IL_Parser.cpp
robert-yates: thanks but i dont see this in the c++ class, or did you mean i should add this because you will be adding it soon? cheers /R.

examples/llil_parser/src/LowLevel_IL_Parser.cpp
plafosse: Oh sorry thats for the user's plugin directory you want: GetBundledPluginDirectory()

examples/llil_parser/src/LowLevel_IL_Parser.cpp
plafosse: Oh sorry about that...I guess we never implemented that API :( The python api does do this and for some reason I though we had this API on the C++ side. It is possible to implement though through the `BNAddAnalysisCompletionEvent` C API, if you wanted to try to implement the UpdateAnalysisAndWait API yourself.

examples/llil_parser/src/LowLevel_IL_Parser.cpp
robert-yates: hmm this appears to add the word plugins onto the current directory that i run my compiled executable from so i dont think its designed for this kind of example

plafosse: Hmm, well I guess the way you're doing it is the correct way then :P We may need to change the way `GetBundledPluginDirectory` works. One one change though MacOS's directory is `/Applications/Binary\ Ninja.app/Contents/MacOS/plugins/` If you care to add.

examples/llil_parser/src/LowLevel_IL_Parser.cpp
robert-yates: added my own local function for syncing, im not sure its 100% correct so i didnt add it directly into the API

plafosse:
We just added UpdateAnalysisAndWait to the C++ API. Would you like to fix before I accept the PR?

plafosse:
Thanks for this example!

PR #746, opened 2017-07-22
Fix use after free in settings api

withzombies:
clang-tidy pointed out a use after free in the settings api.


plafosse:
Thanks, I fixed this directly with some other issues. So no need for a PR. Thanks you!

Issue #741, opened 2017-07-17
Missing operand handler for MLIL SSA form corresponding to x86 cdq instruction


ehennenfent:
x86:
```
080480fd   cdq   {0x0}
```
LLIL:
```
14 @ 080480fd   edx:eax = sx.q(eax)
```
MLIL (SSA Form):
```
6 @ 080480fd   edx#1:eax#2 = 0
```
Attempting to access the MLIL instruction in question results in the following traceback:
```
>>> current_function.medium_level_il.ssa_form[6]
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/home/ehennenfent/binaryninja/plugins/../python/binaryninja/mediumlevelil.py", line 631, in __getitem__
    return MediumLevelILInstruction(self, core.BNGetMediumLevelILIndexForInstruction(self.handle, i), i)
  File "/home/ehennenfent/binaryninja/plugins/../python/binaryninja/mediumlevelil.py", line 239, in __init__
    self.operands.append(value)
UnboundLocalError: local variable 'value' referenced before assignment
```

This issue does not occur when using the non-ssa form of the MLIL instruction. It likely occurs because `MediumLevelILInstruction.ILOperations[core.BNGetMediumLevelILByIndex(current_function.medium_level_il.ssa_form.handle, 6).operation]` returns an empty list, which results in the block of conditionals starting at [line 184 in mediumlevelil.py](https://github.com/Vector35/binaryninja-api/blob/dev/python/mediumlevelil.py#L184) getting skipped, so the `value` variable is never set.

ehennenfent:
Fixed in dev >850 per #749

PR #749, opened 2017-07-25
Fix MLIL operand list typo

joshwatson:
The `operand_type` for an `SSAVariable` should be `"var_ssa"` but the `MLIL_SET_VAR_SPLIT_SSA` operation accidentally had `"ssa_var"`, which causes an `UnboundLocalError` to be raised on line 242 because `value` is never assigned due to falling through the if/else block.


ehennenfent:
Have not tested, but this looks like it should fix #741

plafosse:
@D0ntPanic has a fix for this that he will be pushing soon so we won't be accepting this PR. I'll leave it open until he pushes his code.

plafosse:
Fixed in dev >= 850

PR #770, opened 2017-08-15
Updated BasicBlock.__eq__ to compare arch, start, and end.

fegge:
Also added a compatible `BasicBlock.__hash__`. The previous implementation of `__eq__` is based on `ctypes.addressof` which means that equality may fail to be reflexive.


fegge:
Different basic blocks may share the same `arch`, `start` and `end` (e.g. tail calls).

PR #775, opened 2017-08-17
Fixed possible IndexError in Function.get_*_at methods

joshwatson:
`Function.get_low_level_il_at` and `Function.get_lifted_il_at` methods could raise an IndexError because proper checking was not performed on the index returned by the core method that retrieves the IL instruction index of the requested instruction. The methods now check this value to see if it is equal to the length of the IL function, and if so, returns `None` instead. The onus of checking for `None` will be on the user, but at least they shouldn't have to wrap this in a try/except block anymore.

Given that it was my PR that originally introduced this functionality, I take full responsibility for fixing it :)

### DIFF
--- a/python/function.py
+++ b/python/function.py
@@ -648,7 +648,13 @@ class Function(object):
 		"""
 		if arch is None:
 			arch = self.arch
-		return self.low_level_il[core.BNGetLowLevelILForInstruction(self.handle, arch.handle, addr)]
+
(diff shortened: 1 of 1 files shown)


joshwatson:
This is a pretty simple patch, any reason why it can’t be accepted?

plafosse:
Wasn't accepted because it wasn't looked at basically. Sorry for the delay.

PR #793, opened 2017-09-08
Add shorthand variables to Python console for less typing in GUI

ctfhacker:
As an analyst, I want to be able to quickly access commonly used objects in the Python console in Binary Ninja GUI. Add the following shorthand variables to the Python console for less typing applicable for faster analysis:

```
f -> current_function
h -> here
llil -> current_function.low_level_il
llilssa -> current_function.low_level_il.ssa_form
mlil -> current_function.medium_level_il
mlilssa -> current_function.medium_level_il.ssa_form
```

### DIFF
--- a/python/scriptingprovider.py
+++ b/python/scriptingprovider.py
@@ -540,6 +540,12 @@ class PythonScriptingInstance(ScriptingInstance):
 						self.locals["current_address"] = self.active_addr
 						self.locals["here"] = self.active_addr
 						self.locals["current_selection"] = (self.active_selection_begin, self.active_selection_end)
+						self.locals["f"] = self.active_func
+						self.locals["h"] = self.active_addr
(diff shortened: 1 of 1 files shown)


psifertex:
I think what we will likely do instead of this is allow easier individualized extension of the console Python so that folks can define their own shorthand. My concern with this patch is the principle of least surprise, at least for `f` and `h` which might cause new users to get best surprises if, for example, they were to use those variables in existing code. 

We'll talk it over internally through first, appreciate the PR.

ctfhacker:
Fair enough! Just wanted to pass along some of the shorthand I've been using lately. Feel free to close.

psifertex:
We went ahead and implemented `current_llil` and `current_mlil` but didn't do the other ones per the reasons mentioned above. We're also still working on a system so you can implement your own aliases, so keep an eye out for that.

PR #805, opened 2017-09-19
Fixed AnalysisCompletionEvent callback

joshwatson:
The `AnalysisCompletionEvent` class takes a callback as a parameter in its `__init__` method, but this callback is not added to the object as a bound method automatically in python 2 as it is in python 3. This means that the callback has no parameters passed to it, which prevents it from doing anything meaningful. This PR fixes this by binding the `callback` parameter to an initialized `AnalysisCompletionEvent`.

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -105,7 +105,7 @@ class StringReference(object):
 class AnalysisCompletionEvent(object):
 	def __init__(self, view, callback):
 		self.view = view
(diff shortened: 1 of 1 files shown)


joshwatson:
This API method is completely broken without this PR, any reason why it can’t be accepted?

plafosse:
Fixed in a slightly different way. Thanks @joshwatson!

PR #824, opened 2017-10-04
Added __hash__ impl for BasicBlock instances.

mattrepl:
Simple fix so two BasicBlock instances that are equal (==) will also hash to the same value. Please verify.


psifertex:
Sorry for the delay -- unfortunately, this doesn't account for LowLevelILBasicBlocks or MediumLevelILBasicBlocks so I've re-implemented it a bit [differently](https://github.com/Vector35/binaryninja-api/commit/c396d2cfb3682cb4c152b9dba6f866f6783e4a80). 

Thanks for the suggestion.

PR #832, opened 2017-10-09
Add MLIL and LLIL properties to ReferenceSource

ctfhacker:
As a reverser, I would like to be able to do quickly retrieve the corresponding IL for a given cross reference. 

```
for xref in bv.get_code_refs(bv.symbols['malloc'].address):
 mlil = xref.function.get_low_level_il_at(xref.address).medium_level_il
```

This PR creates simple properties on the `ReferenceSource` in order to replace the above code with the following:

```
for xref in bv.get_code_refs(bv.symbols['malloc'].address):
 mlil = xref.medium_level_il
```

### DIFF
--- a/python/architecture.py
+++ b/python/architecture.py
@@ -1695,3 +1695,11 @@ class ReferenceSource(object):
 			return "<ref: %s@%#x>" % (self.arch.name, self.address)
 		else:
 			return "<ref: %#x>" % self.address
+    
+        @property
+        def low_level_il(self):
+            return self.function.get_low_level_il_at(self.address)
+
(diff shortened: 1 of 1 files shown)


plafosse:
Thank you for the PR. The theres a problem with this code though. An address frequently doesn't have MLIL associated with it due to dead code elimination, this will cause the function to return None, which I think would make the api more difficult to understand.  Additionally, this also has the problem that there can be multiple  mlil instructions for a llil instruction and this case would only return the first. I like the idea of the convenience functions but edge cases I think make them problematic.

PR #862, opened 2017-11-08
Cache arch and platform property values

mattrepl:
The `arch` and `platform` properties are used to calculate the hash value for `Function` instances. Presently, the values for these properties are constructed every time the property is accessed. This PR adds caching for the values of both properties which results in a speedup of the call to `__hash__`.

### DIFF
--- a/python/basicblock.py
+++ b/python/basicblock.py
@@ -64,6 +64,9 @@ class BasicBlock(object):
 			return True
 		return ctypes.addressof(self.handle.contents) != ctypes.addressof(value.handle.contents)
 
+        def __hash__(self):
+                return hash(ctypes.addressof(self.handle.contents))
+
 	def _create_instance(self, view, handle):
 		"""Internal method used to instantiante child instances"""
 		return BasicBlock(view, handle)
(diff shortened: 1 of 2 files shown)


plafosse:
Good idea, I'll accept this after I do some testing with it. Thank you.

plafosse:
Have you signed the contribution license agreement? https://binary.ninja/cla.pdf  It just says that you have the rights to the code you're submitting and you're contributing the code to Binary Ninja. Please email to info@vector35.com.

mattrepl:
Hey, thanks, sounds good. Yep, I have already signed and submitted the agreement. Was a few months ago, IIRC.

joshwatson:
This does have the side effect of violating the read-only restriction of the `arch` property

mattrepl:
Hmm, I’d have to check PEPs to determine if that’s a violation of idiom.

Similar caching is currently done in ‘BasicBlock’ for the ‘arch’ and ‘function’ read property methods.

plafosse:
Yeah, we don't do that since there are a lot of items that would need to be populated and it would slow down instance creation time. I like your fix.  Even though the property is marked as read only this is only to help people not shoot themselves in the foot. By adding a cached internal class variable yes we do slightly open the user up to shoot themselves in the foot, but I believe that risk is minimal and worth the speed gain from your code. Thanks again for the submission.

PR #868, opened 2017-11-14
Cache Platform object and not handle

amatus:


bpotchik:
Thanks for the PR! Fixed in build 1.1.1001-dev.

PR #877, opened 2017-11-20
Fix issues in linux-setup.sh

### LINKED ISSUE #865: scripts/linux-setup.sh fails
I've tested this on fresh installs of Ubuntu 16.04 LTS, Kubuntu 17.10, and Arch Linux and I've gotten the same results for all three. The following errors occur when running the script:

```
$ scripts/linux-setup.sh 
Creating .desktop file
Creating MIME settings
cp: cannot stat '/home/algebro/.local/share/applications/binaryninja.desktop': No such file or directory
lastrun already exists, remove to create a new one
Configuring python path
python: can't open file '/home/algebro/software/binaryninja/scripts/scripts/install_api.py': [Errno 2] No such file or directory
```
I am still troubleshooting but I've noticed a few bugs in the script such as the .desktop file not correctly adding newlines after each key:value entry. It also does not create the file in `$SHARE` if the script is ran as a normal user, but it later attempts to `cp` it from there to the Desktop...and I'm not quite sure what is going on with the Python API installation since it works if I manually run `python scripts/install_api.py`.

njfox:
Fixes #865 

Tested in Ubuntu 16.04, Fedora 26, Kali, and Arch Linux. The python API installation still fails on Arch and Fedora but I'm not sure why--I'm planning to address that in another PR.

### DIFF
--- a/scripts/linux-setup.sh
+++ b/scripts/linux-setup.sh
@@ -14,7 +14,7 @@ setvars()
 	APP="binaryninja"
 	FILECOMMENT="Binary Ninja Analysis Database"
 	APPCOMMENT="Binary Ninja: A Reverse Engineering Platform"
-	BNPATH="$(dirname $(readlink -f "$0"))/../"
+	BNPATH="$(dirname $(readlink -f "$0"))/.."
 	EXEC="${BNPATH}/binaryninja"
 	PNG="${BNPATH}/docs/images/logo.png"
 	EXT="bndb"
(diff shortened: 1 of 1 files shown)


plafosse:
Thanks for the PR. We'll look over this soon. Have you provided us with a CLA? https://binary.ninja/cla.pdf

njfox:
Yep, sent it a little bit ago.

njfox:
adfddcf fixes the python API installation error on platforms where `/usr/bin/python` is pointed at python3.

psifertex:
Looks good, thanks! Sorry for the delay, had some dead VMs it took me a while to rebuild for testing.

rmspeers:
Exactly, if I install python2, then it'll install into that -- but I need it accessible in Py3 for where I'll call it from. I'll submit a PR linked to this ticket for your review.

PR #885, opened 2017-12-07
Fix PluginManager Doc Example

toolCHAINZ:
Noticed that the doc string example for add_repository had a named argument that doesn't seem to exist any more.


plafosse:
Thanks for catching this.

PR #938, opened 2018-02-16
Fix get_symantic_flag_group_name typo

kukfa:
Exception `'Architecture' object has no attribute 'get_semantic_flag_group_name'` occurred when calling `get_lifted_il_at`. Fixed by correcting the API call since an index is being used here

Will get the Contribution License Agreement started tomorrow :)


psifertex:
Actually, the original code was correct, the function was just missing (along with a similar one for `class_name`). That code is supposed to return the name from the index, but using `get_semantic_flag_group_index` does the opposite. It takes a name and will return the index. 

Thanks for pointing out the bug though! Implemented that in 73b3f7fc2ff9dab40f29da2031a0c6c71220d93c and also manually added those two typo fixes.

PR #1061, opened 2018-06-11
Fix CMakeLists.txt for both mlil_parser and llil_parser examples

joshwatson:
The `file( GLOB_RECURSE ...` line of the CMakeLists.txt files causes a link error, because the recursive file search also finds a cmake cpp file that contains a `main` symbol.

Additionally, I fixed a compile issue with llil_parser, where `LLIL_CALL_PARAM_SSA` was used instead of `LLIL_CALL_PARAM`.


joshwatson:
anyone?

plafosse:
Thanks for the PR :)

PR #1068, opened 2018-06-22
fix #1057

### LINKED ISSUE #1057: BinaryView.get_strings TypeError
`BinaryView.get_strings` raises `exceptions.TypeError` if the first argument is supplied but not the second.

    >>> bv.get_strings()[0]
    <StringType.AsciiString: 0x899b9, len 0x4>
    >>> bv.get_strings(19385448)[0]
      File "<console>", line 1, in <module>
      File "/home/kevin/.local/share/binaryninja/plugins/../python/binaryninja/binaryview.py", line 2723, in get_strings
        strings = core.BNGetStringsInRange(self.handle, start, length, count)
      File "/home/kevin/.local/share/binaryninja/plugins/../python/binaryninja/_binaryninjacore.py", line 6133, in BNGetStringsInRange
        result = _BNGetStringsInRange(*args)
    ArgumentError: argument 3: <type 'exceptions.TypeError'>: wrong type
    >>> bv.get_strings(19385448, 1)[0]
    <StringType.AsciiString: 0x127cc68, len 0x1f>

I would think if the second argument is `None` (the default) then it should default to `bv.end - start`. I.e. the remaining length of the bv.

negasora:


lucasduffey:
https://github.com/Vector35/binaryninja-api/issues/1057

PR #1069, opened 2018-06-22
PythonX Support

### LINKED ISSUE #376: python3 support
See subject.

### LINKED ISSUE #454: wrap Architecture.assemble to return python style exceptions
Right now the `Architecture.assemble` api returns a tuple of None and an error message if assemble fails, we should instead throw an exception which is the more pythonic thing to do. (was #310)

KyleMiles:
Resolves #376 - Python3 Support
Resolves #454 - Assemble Should Throws Errors

Changelog:
* bv.write and bv.insert require a bytes object in python3
* Architecture.assemble outputs a bytes object in python3, a str in python2
* Architecture.assemble will throw a value error if it cannot assemble the given instruction
* API install script should be run in the version of python you want it installed in

Fundamental python changes to be aware of:
* Unicode-type strings are now just str, consequently anything that came out as a unicode string before (annotations) are now just str.
* Longs no longer exist.  They're just ints.


rmspeers:
I am trying to run a Python3 install which based on the comments in this ticket should work, but the import, although it finds the binaryninja/__init__.py script, can't find the other modules inside of it. This works fine on the system if I use Python2, but I need to call from Python3.

```
>>> import os
>>> import sys
>>> path = '/binaryninja/python'
>>> os.stat(path)
os.stat_result(st_mode=16877, st_ino=1661234, st_dev=1048719, st_nlink=3, st_uid=0, st_gid=0, st_size=4096, st_atime=1543404926, st_mtime=1543404585, st_ctime=1543404585)
>>> old_path = sys.path
>>> sys.path.append(path)
>>> sys.path
['', '/usr/local/lib/python3.5/dist-packages/mypackage-0.0.1-py3.5.egg', '/usr/local/lib/python3.5/dist-packages/mypackage1-0.0.1-py3.5.egg', '/usr/lib/python35.zip', '/usr/lib/python3.5', '/usr/lib/python3.5/plat-x86_64-linux-gnu', '/usr/lib/python3.5/lib-dynload', '/usr/local/lib/python3.5/dist-packages', '/usr/lib/python3/dist-packages', '/binaryninja/python']
>>> import binaryninja
  File "<stdin>", line 1, in <module>
  File "/binaryninja/python/binaryninja/__init__.py", line 27, in <module>
    import _binaryninjacore as core
ImportError: No module named '_binaryninjacore'
```

```
root@host:/binaryninja# python3 -V
Python 3.5.2
root@host:/binaryninja# python -V
Python 3.5.2
root@host:/binaryninja# uname -a
Linux host 4.9.125-linuxkit #1 SMP Fri Sep 7 05:20:00 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

PR #1095, opened 2018-07-28
Update windows plugin path

yrp604:
Unless my install is weird somehow, the Binary Ninja plugin folder on Windows lives under Roaming.


KevOrr:
`%APPDATA` on Vista and later refers to `C:\Users\<user name>\AppData\Roaming`.

https://www.microsoft.com/en-us/wdsi/help/folder-variables

You can easily browse to a path that includes variables by typing/pasting the path into the address bar in Windows Explorer.

yrp604:
Oops.

PR #1143, opened 2018-09-07
Fix an incorrect field name in FlowGraphLayoutRequest._complete

joshwatson:
`FlowGraphLayoutRequest.on_complete` was incorrectly referenced as `_on_complete` in `FlowGraphLayoutRequest._complete`. This simple PR fixes the issue.


bpotchik:
Thanks!

PR #1160, opened 2018-10-06
Fix link to issue tracker

jryans:
I noticed a broken link while scanning the docs, so here's a tiny PR to fix that up.


psifertex:
Thanks for the fix.

Issue #1093, opened 2018-07-27
Poor DataBuffer performance


0x1F9F1:
Binary Ninja Version: 1.1.1270-dev,  f0bc0ae
Platform: Windows 10 (10.0)

While working with DataBuffers I noticed they have very poor performance. After a bit of investigating it seems to be down to 2 main problems:
- No move constructor/assignment operator (in the C++ API) causing multiple copy calls in most situations (see https://github.com/DatBrick/binaryninja-api/blob/dev/databuffer.cpp)
- Implemented as a fixed size buffer internally, with any write requiring a call to realloc. This can end up making writes painfully slow. After a bit of profiling it appears this also causes a significant slowdown in parts of the core when saving a BinaryView.

joshwatson:
Eek, yeah I agree DataBuffer should definitely either use a technique like that of modern javascript engines' string objects, or it should reallocate similar to the vector class

plafosse:
All of our usage in the core doesn't have to go through the API so we don't see issues like this, thanks for the report.  FWIW lines 53 and 84 on that code you referenced have the move and move assignment operators. It could just be the speed of BNCreateDataBuffer as it does allocate a new structure each time.

joshwatson:
Yeah that's DatBrick's clone of the repo with those changes. The move operators aren't in the official API repo.

plafosse:
Oh ha :P

0x1F9F1:
After getting impatient with hour+ long database saves, I did some more testing.

Here is an idea of the time it takes to save a 250MB bndb (I gave up past that point):

| Percentage | Total Elapsed (Minutes) | Difference |
|------------|-------------------------|------------|
|         15 |                       3 |          3 |
|         25 |                       8 |          5 |
|         35 |                      15 |          7 |
|         45 |                      25 |         10 |
|         55 |                      37 |         12 |

After hooking realloc to make it act more like a vector (https://github.com/0x1F9F1/binja-realloc/blob/master/src/main.cpp#L82) the total save time dropped to under 3 minutes.

TLDR: DataBuffer shouldn't rely on realloc, and its usage is causing massive slowdowns in places.

bpotchik:
Fixed in 1.1.1485-dev.

psifertex:
Sorry, hit the wrong button :-) should still be closed. Brian mentioned in testing a 1.5hr save went to 10m.

PR #1172, opened 2018-10-19
Added move constructor/assignment operator to DataBuffer

DatBrick:
This should prevent a lot of unncessary copies when using DataBuffers (fixes part of #1093).


psifertex:
Thanks for the PR!

PR #1188, opened 2018-10-25
Return assemble error string as part of the ValueError exception

shane-runsafe:
CoreArchitecture.assemble now returns the error string from the assembler as text in the ValueError exception.


psifertex:
Looks good, thanks!

PR #1201, opened 2018-11-06
Add type check to `Variable.__eq__`

joshwatson:
`Variable.__eq__` fails to check if the other value is actually a Variable object before comparing member properties. This fixes the exception that gets raised when comparing a Variable against something else.

### DIFF
--- a/python/function.py
+++ b/python/function.py
@@ -293,6 +293,8 @@ class Variable(object):
 		return self.name
 
 	def __eq__(self, other):
+		if not isinstance(other, Variable):
+			return False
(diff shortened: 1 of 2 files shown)


joshwatson:
I added a second __eq__ to this, for `MediumLevelILOperationAndSize`; it will return True when compared against its corresponding `MediumLevelILOperation` as well as if `other` is the same size/operation

psifertex:
Hah! I was about to open a review about the missing return false but noticed it in the other commit before I did. ;-) 

Thanks, looks good.

PR #1202, opened 2018-11-10
Fix imports in functionrecognizer.py

joshwatson:
Some imports were left off for some reason. I added them.


psifertex:
Thanks for the PR!

PR #1204, opened 2018-11-12
Fix export_svg.py

joshwatson:
Fixed export_svg.py to work with FlowGraphs and fixed several issues due to the Python3 refactor


psifertex:
Thanks for the PR.

PR #1223, opened 2018-12-21
Added TransformParameter.__repr__

joshwatson:
`TransformParameter` didn't have a `__repr__` method so I added one.


bpotchik:
Fixed in 1.1.11460-dev.

PR #1224, opened 2018-12-22
Fix metadata

joshwatson:
Metadata doesn't work so well when you're pickling things that can potentially have a null-byte in the resulting pickled data. This PR does two things:

1. It allows raw Metadata objects to be passed into `BinaryView.store_metadata`, when previously they could not be.
2. It allows python3 to pass either a `str` or `bytes` object to `Metadata` as a value when the optional parameter `raw` is `True`.


bpotchik:
Fixed in 1.1.1459-dev.

PR #1226, opened 2018-12-26
Fix freeing of user allocated BNInstructionTextToken's

0x1F9F1:
BNInstructionTextToken typeName's were previously not freed.


plafosse:
Thanks for this, I already had a fix in my local branch and just pushed it.

PR #1235, opened 2019-01-08
Fixed install_api.py for virtualenv

joshwatson:
`virtualenv` has its own `site.py` that does not contain `getsitepackages` or `getusersitepackages`. This commit replaces those imports with equivalent ones that do work in `virtualenv`.


psifertex:
Looks good, thanks.

PR #1252, opened 2019-01-21
Fixed imports in undoaction.py

joshwatson:
`undoaction.py` was missing an import and so it was broken as soon as you tried to use it. This is just a quick fix for that.

### DIFF
--- a/python/undoaction.py
+++ b/python/undoaction.py
@@ -25,6 +25,7 @@ import ctypes
 # Binary Ninja components
 from binaryninja import _binaryninjacore as core
 from binaryninja.enums import ActionType
+import binaryninja
 
 
 class UndoAction(object):


joshwatson:
In this case, it is the specific import! Note that it's `import binaryninja` not `from binaryninja import *`

This is because [line 54](https://github.com/Vector35/binaryninja-api/blob/09bd22fb46e672a49991820facf47254cbe83949/python/undoaction.py#L54) calls `binaryninja._initialize_plugins()` which you is an underscored name, so the module name has to be imported. I based this PR on other API files that invoke that particular method.

psifertex:
Yup, it's totally fine, I missed that originally. Peter was giving me a hard time about it too. ;-) LGTM.

PR #1257, opened 2019-01-24
Initial commit of StructuredDataView and StructureDataValue classes

withzombies:
@joshwatson bugged me to try and get this merged into `binaryview.py`

The class represents a read-only interface into a structure mapped over memory. For example, in the API we can apply the `Elf64_Header` structure on an address (such as `0x400000`). Then access it's members directly:

```python
print "Analyzing {0}".format(inputfile)
bv = BinaryViewType.get_view_of_file(inputfile)
bv.update_analysis_and_wait()

structure = "Elf64_Header"
address = 0x400000
sdv = StructuredDataView(bv, structure, address)

print("ELF Machine: {:#x}".format(sdv.machine))
```

Will print 0x003e, the elf machine for my example binary.

The `__str__` method has also been overridden to pretty print a structure:
```python
print(sdv)
```

Outputs:
```
struct Elf64_Header 0x400000 {
	    +0 struct Elf64_Ident ident                 = 7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
	   +10 uint16_t type                            = 0002
	   +12 uint16_t machine                         = 003e
	   +14 uint32_t version                         = 00000001
	   +18 void (*)() entry                         = 0000000000400c90
	   +20 uint64_t program_header_offset           = 0000000000000040
	   +28 uint64_t section_header_offset           = 00000000000048b8
	   +30 uint32_t flags                           = 00000000
	   +34 uint16_t header_size                     = 0040
	   +36 uint16_t program_header_size             = 0038
	   +38 uint16_t program_header_count            = 0009
	   +3a uint16_t section_header_size             = 0040
	   +3c uint16_t section_header_count            = 0024
	   +3e uint16_t string_table                    = 0021
}
```

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -24,10 +24,12 @@ import ctypes
 import abc
 import numbers
 
+from collections import OrderedDict
+
 # Binary Ninja components
 from binaryninja import _binaryninjacore as core
 from binaryninja.enums import (AnalysisState, SymbolType, InstructionTextTokenType,
-	Endianness, ModificationStatus, StringType, SegmentFlag, SectionSemantics, FindFlag)
+	Endianness, ModificationStatus, StringType, SegmentFlag, SectionSemantics, FindFlag, TypeClass)
 import binaryninja
 from binaryninja import associateddatastore # required for _BinaryViewAssociatedDataStore
 from binaryninja import log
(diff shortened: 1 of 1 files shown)


plafosse:
This is very cool! We will review and accept soon.

psifertex:
Love the feature -- for reference, as we've talked about before on slack I think, we do want to integrate this into the C++ API as well or the core so that it's more broadly available but we don't want to hold up being able to use it now so we're going to accept the PR with the expectation that we'll either add similar functionality later to the C++ API or move it into the core and change this to use the core implementation.

Issue #1155, opened 2018-09-29
Metadata cannot round-trip


joshwatson:
```
>>> bv.store_metadata('buggy', {0: 12})
>>> bv.query_metadata('buggy')
{'0': 12L}
>>> bv.store_metadata('buggy2', _)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Applications/Binary Ninja.app/Contents/MacOS/plugins/../../Resources/python/binaryninja/binaryview.py", line 3574, in store_metadata
    core.BNBinaryViewStoreMetadata(self.handle, key, metadata.Metadata(md).handle)
  File "/Applications/Binary Ninja.app/Contents/MacOS/plugins/../../Resources/python/binaryninja/metadata.py", line 61, in __init__
    md = Metadata(value[elm], signed, raw)
  File "/Applications/Binary Ninja.app/Contents/MacOS/plugins/../../Resources/python/binaryninja/metadata.py", line 64, in __init__
    raise ValueError("List doesn't not contain type of: int, bool, str, float, list, dict")
ValueError: List doesn't not contain type of: int, bool, str, float, list, dict
```

plafosse:
Because I just talked about this in chat... This maps to a native API, and thus any language which has bindings to the native API.  e have to coerce python int to C++ uint64 then to python ... well we chose to do long because its the best option, if we changed it to int it would be broken in a different way. The only way to fix it would to be to include some other bit of metadata that says what each of the values should be coerced back to for python 2.7.

plafosse:
Because I just discussed this in chat... This is specifically a python2.7 issue that I don't think we will fix. We have to coerce python `int` to C++ `uint64` in the native API. To read it back out we have to decide how what python2 type to convert the native `uint64` into. We chose a python `long` because it can represent the full with `uint64 value.  The only way to fix it would to be to include some other bit of metadata that says what each of the values should be coerced back to for python 2.7. If anyone has a suggestion on a better way to do this I'm happy to entertain suggestions :)

plafosse:
Oh also another workaround is to use python3 instead!

joshwatson:
Fortunately for me I guess, I've moved completely to python3 at this point anyways!

PR #1270, opened 2019-02-05
Fix some python2 / python3 issues in Metadata

bambu:
I encountered same error @joshwatson did in issue #1155. In his case, it seems that the metadata was  not checking for `long` type. In my case it was a `tuple`. 

1. Long is not supported in python3 so using `numbers.Integral` instead.
2. added `tuple` to the list case
3. `strings` are restored as `bytes`. Changed to restore string as string and bytes as bytes. in python2 both functions restore as str. 
4. Added type to the ValueError exception to display the type that is not supported.

### DIFF
--- a/python/metadata.py
+++ b/python/metadata.py
@@ -21,6 +21,7 @@
 
 from __future__ import absolute_import
 import ctypes
+import numbers
 
 # Binary Ninja components
 from binaryninja import _binaryninjacore as core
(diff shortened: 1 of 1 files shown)


psifertex:
Sorry, I whitelisted your email as already signing it but it looks like that isn't working. It's the same form as before and super fast online acceptance.

PR #1277, opened 2019-02-13
Improved CMake Compilation

0x1F9F1:
The `binaryninjaapi` CMakeLists.txt will now:
* Require 64-bit compilation
* Find and link to `binaryninjacore`
* Set the target include directories
* Enable position independent code
* Optionally build examples with `-DBINJA_BUILD_EXAMPLES`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ CTestTestfile.cmake
 api-docs/source/python.rst
 api-docs/source/index.rst
 .coverage
+build
 # Make
 generator
 plugins/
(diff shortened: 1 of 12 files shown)


withzombies:
I checked this PR and it looks good. It was able to build on my mac, i was able to build the examples and run them. Also tested `make install` which installed the `.dylib` plugins to my user plugin folder.

I also reviewed the CMakeLists, and the "new" style `target_` directives are being used which is the current best practice. My only concern is the CMake minimum version number is set to 3.4 which might be too low.

PR #1295, opened 2019-03-11
Change get_next_data_var_after return type

### LINKED ISSUE #1294: get_next_data_var_after return type
Binary Ninja Version: 1.1.1540-dev, c51ae7c5
Platform: macOS 10.14

The example usage seems a bit roundabout:
```
>>> hex(bv.get_next_data_var_after(0x10000000))
'0x1000003cL'
>>> bv.get_data_var_at(0x1000003c)
<var 0x1000003c: int32_t>
```

I would expect:
```
>>> bv.get_next_data_var_after(0x10000000)
<var 0x1000003c: int32_t>
```

devtty1er:
(Doesn't quite fix #1294)

Not really sure how to handle this issue, but I thought I'd try my hand at a PR. I feel like `core.BNGetNextDataVariableAfterAddress` should probably change instead of the Python API? This  PR also doesn't address when there are no data variables after `addr`, ~~which will probably result in a KeyError.~~

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -3190,7 +3190,7 @@ class BinaryView(object):
 			<var 0x1000003c: int32_t>
 			>>>
 		"""
-		return self.data_vars.get(core.BNGetNextDataVariableAfterAddress(self.handle, addr), None)
+		return self.data_vars.get(core.BNGetNextDataVariableAfterAddress(self.handle, addr))
 
 	def get_previous_function_start_before(self, addr):
 		"""
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -3190,15 +3190,7 @@ class BinaryView(object):
 			<var 0x1000003c: int32_t>
 			>>>
 		"""
-		next_data_var_addr = core.BNGetNextDataVariableAfterAddress(self.handle, addr)
-		if next_data_var_addr == addr:
-			# core.BNGetNextDataVariableAfterAddress always returns an next addr,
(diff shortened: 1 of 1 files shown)


python/binaryview.py
```diff
@@ -3179,20 +3179,18 @@ def get_next_data_after(self, addr):
...
 
-			>>> hex(bv.get_next_data_var_after(0x10000000))
-			'0x1000003cL'
-			>>> bv.get_data_var_at(0x1000003c)
+			>>> bv.get_next_data_var_after(0x10000000)
 			<var 0x1000003c: int32_t>
 			>>>
 		"""
-		return core.BNGetNextDataVariableAfterAddress(self.handle, addr)
+		return self.data_vars.get(core.BNGetNextDataVariableAfterAddress(self.handle, addr), None)
```
joshwatson:
```suggestion
		return self.data_vars.get(core.BNGetNextDataVariableAfterAddress(self.handle, addr))
```
You don't need the `, None` because it's the default behavior of `dict.get`

devtty1er: Thanks!

bpotchik:
Fixed in 1.1.1613-dev. The API for data variables was cleaned up to have `_start` similar to the other relevant APIs. A convenience method was added to python for returning DataVariable objects.

This PR could not be accepted as is because the get_next_data_var method was changed to index into the data_vars property, which retrieves all data variables for the entire binary view, which would not be very performant.

PR #1297, opened 2019-03-12
Add __hash__ to IL Function objects

joshwatson:
They didn't have a __hash__ method so I added one so I can use them in sets. It's really just the hash of the Function object plus a string to indicate it's MLIL or LLIL

### DIFF
--- a/python/lowlevelil.py
+++ b/python/lowlevelil.py
@@ -752,6 +752,9 @@ class LowLevelILFunction(object):
 				func_handle = self.source_function.handle
 			self.handle = core.BNCreateLowLevelILFunction(arch.handle, func_handle)
 
+	def __hash__(self):
+		return hash('LLIL') + hash(self.source_function)
+
 	def __del__(self):
 		if self.handle is not None:
 			core.BNFreeLowLevelILFunction(self.handle)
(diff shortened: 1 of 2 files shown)


python/lowlevelil.py
```diff
@@ -752,6 +752,9 @@ def __init__(self, arch = None, handle = None, source_func = None):
 				func_handle = self.source_function.handle
 			self.handle = core.BNCreateLowLevelILFunction(arch.handle, func_handle)
 
+	def __hash__(self):
+		return hash('LLIL') + hash(self.source_function)
```
withzombies:
```suggestion
		return hash(('LLIL', self.source_function))
```

python/mediumlevelil.py
```diff
@@ -633,6 +633,9 @@ def __init__(self, arch = None, handle = None, source_func = None):
 			func_handle = self.source_function.handle
 			self.handle = core.BNCreateMediumLevelILFunction(arch.handle, func_handle)
 
+	def __hash__(self):
+		return hash('MLIL') + hash(self.source_function)
```
withzombies:
```suggestion
		return hash(('MLIL', self.source_function))
```

joshwatson:
Okay, accepted Ryan's suggestions, can I get this merged so I don't have to keep modifying my mediumlevelil.py file every time I update? Hahahaha

Issue #1298, opened 2019-03-13
get_strings improvements


devtty1er:
Binary Ninja Version: 1.1.1544-dev, e54a51f3
Platform: macOS 10.14

![https://imgur.com/OUwKzrh](https://imgur.com/download/OUwKzrh)
```
>>> bv.get_strings(0x15c8, 1)
[]
>>> bv.get_strings(0x15c7, 1)[0]
[<StringType.AsciiString: 0x15c7, len 0x8>]
>>> bv.get_strings(0x15c7, 1)[0].value
'@6Square'
```

[virtual.bndb](https://www.dropbox.com/s/07ud3u4qsrmmioa/virtual.bndb?dl=0) @ `0x15c8`

psifertex:
I can't reproduce the issue with a sample binary here. Is it just that binary that does it for you? Just that offset? Can you share the bndb?

devtty1er:
@psifertex [virtual.bndb](https://www.dropbox.com/s/07ud3u4qsrmmioa/virtual.bndb?dl=0) @ `0x15c8`

psifertex:
Ahh, perfect, thanks. So there's a few separate issues this points out actually that I'm going to track here:

- [ ] `get_strings` should better handle when a string exists at a starting location BEFORE the current address as long as it overlaps (in this case, our original strings sweep finds the string at 0x15c7 which is wrong, but just happens because there's a `@` character inside the previous pointer.
- [ ] When our naive strings sweep finds a string, if we later get better type information that supersedes it we should remove/update the original string type based on the better information (Bonus points: this should use our confidence system)
- [ ] A more moderate form of the previous step would be to adjust strings whenever we have information that creates other data references that overlap with the string to truncate the string to remove the overlapping characters. 

The simplest fix is likely the first one, but we'll track the others here too.

psifertex:
So we've been talking about it some more and I'm not convinced that the API /should/ include strings that start before the range specified. We might introduce a separate API that would do so, but this API is behaving as intended even if it's not quite as obvious. Of course, the second two improvements still remain.

devtty1er:
> I'm not convinced that the API /should/ include strings that start before the range specified.

[...]

PR #1299, opened 2019-03-14
Add get_ascii_string_at

devtty1er:
Vector35/binaryninja-api#1298

This handles only ASCII strings; presumably we would want something similar for UTF. This is one way to do things without changing the `bv.get_strings()` implementation.

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -3080,7 +3080,7 @@ class BinaryView(object):
 		:param int start: optional virtual address to start the string list from, defaults to start of the binary
 		:param int length: optional length range to return strings from, defaults to length of the binary
 		:return: a list of all strings or a list of strings defined between ``start`` and ``start+length``
-		:rtype: list(str())
+		:rtype: list(StringReference)
 		:Example:
 
 			>>> bv.get_strings(0x1000004d, 1)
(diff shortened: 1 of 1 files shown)


withzombies:
For the failure case, you return a StringReference with a length of 0.

Would it be better to return `None` or maybe throw an exception? I'm partial to returning `None`.

devtty1er:
Hmmm... I agree. @withzombies I'm also curious if you think I should even bother with StringReference or just return the string itself?

withzombies:
Honestly I kinda like returning the `StringReference` because i find it has more information that you can pass around. imo its best to keep data encapsulated as long as possible.

It also allows me the opportunity to do really horrible things like monkey patching `StringReference` and i can't reasonably do that with `str`.

withzombies:
the `strings` binary has a minimum string length option that defaults to 4 on macOS. This api should probably have a minimum length.

It also could look for CStrings (requiring a trailing null) and returning `None` otherwise.

The prototype could be something like:
```python
def get_ascii_string_at(self, addr, minimum_length=4, require_cstring=True):
```

I don't know how i feel about the argument names but an API to get a "string" to me should probably look for only cstrings. Right now the function will stop once it's hit a value that is `0` or `>= 0x80`.

withzombies:
I think this looks pretty good

My only other feedback would be: what happens when a BinaryReader hits the end of the file? Does it throw an exception? Should we handle the exception and return None?

I see a few approaches:
* If we hit the end of the page, catch the exception and return None
* If we hit the end of the page, catch the exception and return what we have already -- assuming it meets our requirements
* Don't handle the exception

withzombies:
i guess we have the same problem if you pass an invalid address in also

devtty1er:
I think? it will return None in all of those cases -- no exceptions.

withzombies:
I don't follow, if `read8` returns `None` rather than throwing an exception we have some other results:

* If we read 5 characters before the end of the file and `min_length=4` and `max_length=None` and `require_cstring=False` we return the `StringReference` with a length of 5.
* If we read 5 characters before the end of the file and `min_length=4` and `max_length=None` and `require_cstring=True` we return None because `ch` is `None`
* If we read 3 characters before the end of the file and `min_length=4` we return `None`

I guess this is probably how we would want it to behave, right?

plafosse:
I would also ask that you squash these commits down to a single commit, please.

devtty1er:
@withzombies Yes. Those assertions are correct, reflected by the current implementation to my knowledge and testing, and I think they are how we would want it to behave too.

@plafosse Done.

psifertex:
Thanks for this PR, per the previous discussion in #1298 we consider this a bit of a stop-gap until we've get better API support for broader classes of strings, but it's useful enough and a common-enough paradigm that it's helpful to add now.

One minor caveat that the python implementation of this is going to necessarily be relatively slow. If it's commonly used enough we might implement something more performant in the core and then just change the implementation out. Same goes for the broader string class support coming "Eventually".

PR #1339, opened 2019-05-04
made header file valid C

mpdairy:
I changed all your struct and enum definitions in binaryninjacore.h to typedefs to make them valid C, which is essential for automatic binding generators (particularly, c2hs for haskell).


psifertex:
There was some whitespace mis-matches (you've got a tabstop of 2?! ;-) ) that I resolved and re-implemented the patch separately so you didn't have to redo it. Thanks for the contribution!

PR #1346, opened 2019-05-10
Fix Architecture.assemble so that it works with ArchitectureHooks

joshwatson:


plafosse:
We're not going to accept this right away because this basically eats all errors, though it is better than what exists currently.

joshwatson:
It doesn't eat the exception though, it passes the error back and returns False, like assemble does: https://github.com/Vector35/binaryninja-api/blob/dev/python/architecture.py#L2395

PR #1367, opened 2019-06-12
make get_var_definition and get_var_uses return instructions instead of indexes

verylazyguy:


KyleMiles:
Should be good if all the tests pass.. Jordan would like to notify plugin maintainers that this breaks before merging it to dev.

verylazyguy:
it *shouldn't* break anything because the __getitem__ on a MediumLevelILFunction checks if the index is a MediumLevelILExpression and returns it as is if that's the case

joshwatson:
It shouldn't actually break much, because passing an index or instruction into the `__getitem__` method of an IL function object will still return the instruction

verylazyguy:
should I remake this pr?

KyleMiles:
No need to remake it; I did (you are credited): https://github.com/Vector35/binaryninja-api/commit/f942ccf4af2e12ed2a78ce1f55708c39ef461185

PR #1438, opened 2019-08-29
Make bv.get_functions_containing always return a list

verylazyguy:
`bv.get_functions_containing()` has to be assigned to a variable then tested for `None` currently.

This patch allows `for func in bv.get_functions_containing(addr):` to run without raising an exception. This should allow for cleaner code

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -2809,11 +2809,9 @@ class BinaryView(object):
 		``get_functions_containing`` returns a list of functions which contain the given address or None on failure.
 
 		:param int addr: virtual address to query.
-		:rtype: list of Function objects or None
+		:rtype: list of Function objects
 		"""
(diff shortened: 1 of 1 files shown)


devtty1er:
Fortunately it seems like a lot of people were cautious checking both `functions == None` and `len(functions) == 0` 😛

devtty1er:
Will help where I can...

psifertex:
Thanks. I just checked everything in the plugin manager and they all look similar enough to what you described above so yeah, it looks like they'll all work just fine with this change. To be fair, we thought that with the last one like this and a few things snuck through, but it's worth dealing with.

devtty1er:
Either that or just `bv.get_functions_containing(addr)[0]`, in which case we didn't break anything that wasn't already broken lol

PR #1439, opened 2019-08-30
Support all colors of the rainbow in export_svg.py

mechanicalnull:
Some scripts/plugins like bncov use block.set_user_highlight() with colors constructed by RGB value instead of default colors.  This addition allows export_svg.py to show these colors.


psifertex:
Looks good, thanks!

PR #1465, opened 2019-10-10
Add printf annotation for BNLog*

meme:
Adds nice inference for IDEs, and emits compiler warnings when incorrect format strings are being used.


psifertex:
Looks good, thanks!

PR #1476, opened 2019-10-19
Add visibility to MediumLevelILInstructionAccessException

joshwatson:
This gets rid of the massive number of warnings that look like this when you compile anything that includes `mediumlevelilinstruction.h`:
```
ld: warning: direct access in function 'BinaryNinja::MediumLevelILInstruction::GetOutputSSAVariables() const' from file '/Users/user/Documents/Repos/binaryninja-api/bin/libbinaryninjaapi.a(mediumlevelilinstruction.cpp.o)' to global weak symbol 'typeinfo for BinaryNinja::MediumLevelILInstructionAccessException' from file 'analysis.cpp.o' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```

### DIFF
--- a/mediumlevelilinstruction.h
+++ b/mediumlevelilinstruction.h
@@ -179,7 +179,7 @@ namespace BinaryNinjaCore
 namespace BinaryNinja
 #endif
 {
-	class MediumLevelILInstructionAccessException: public std::exception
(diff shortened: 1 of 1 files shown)


psifertex:
This will break our windows build currently. If you ifdef it out to just the platform you're working on, that would be fine.

PR #1490, opened 2019-11-26
Content for Types & Structures user guide page.

TACIXAT:
*insert passive aggressive snapchat of empty docs page here*

### DIFF
--- a/docs/guide/type.md
+++ b/docs/guide/type.md
@@ -1,2 +1,42 @@
 # Types and Structures
 
+The types view can be opened from the top menu `View > Types`. From here you can create structures, unions, and types using C-style syntax.
+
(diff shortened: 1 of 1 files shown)


TACIXAT:
Would be helpful to also add a note about creating structures. I struggled with random errors before I gave the struct a size.

psifertex:
Was this actually linked to from anywhere? We've got a few placeholders that aren't supposed to be visible.

That said, we agree this documentation would be helpful. We've made some improvements to the struct workflow that we're working into a long-form documentation (hint, use "s" in linear view to get nice automatic struct and member creation).

I'll add this for now though since I don't have a specific ETA for our improved docs.

Thanks!

PR #1492, opened 2019-12-02
Update architecture doctests to python3

devtty1er:
More PR's incoming for the other modules... I might end up combining into a single PR...

### DIFF
--- a/python/architecture.py
+++ b/python/architecture.py
@@ -96,16 +96,14 @@ class Architecture(with_metaclass(_ArchitectureMetaClass, object)):
 
 		>>> #List the architectures
 		>>> list(Architecture)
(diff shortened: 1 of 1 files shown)


python/architecture.py
```diff
@@ -2394,7 +2393,7 @@ def get_flag_write_low_level_il(self, op, size, write_type, flag, operands, il):
 				operand_list[i].constant = True
 				operand_list[i].value = operands[i]
 		return lowlevelil.LowLevelILExpr(core.BNGetArchitectureFlagWriteLowLevelIL(self.handle, op, size,
-		        self._flag_write_types[write_type], flag, operand_list, len(operand_list), il.handle))
+				self._flag_write_types[write_type], flag, operand_list, len(operand_list), il.handle))
```
psifertex: This looks like it might have been a change from an editor re-aligning? We typically use spaces for indentation for line-wraps intentionally.

devtty1er:
Fixed!

PR #1561, opened 2020-03-30
Enable use within a venv

3pidemix:
Checks for existence of **VIRTUAL_ENV** environment variable, and if it exists, install API at first path from _getsitepackages()_, which should always be the venv's site-packages folder. NOTE: Assumes user will _always_ want to do this if script is run from a virtual environment. Tested on macOS 10.15.3 with Homebrew Python3, Windows 10 with python.org 64-bit Python 3.8, and Ubuntu 18.04 with _python3-venv_ package installed.


psifertex:
I like this, though if you don't mind modifying it, can you add a warning about that behavior if it's triggered? That way folks can know they need to exit the VENV if they want it system-wide.

psifertex:
Great, thanks!

PR #1566, opened 2020-04-15
Improved installation process for venv

3pidemix:
Added the -v option to make the venv installation explicit. This should eliminate any confusion about normal installation and leaves the original functionality intact.


psifertex:
Looks good, thanks!

PR #1623, opened 2020-04-29
export_svg: Fix the fix for saving on Windows

meme:


psifertex:
It's only if you're /really sure/ of the platform is it a platfirm. 

Man, what a whoops. 👀

Thanks.

PR #1638, opened 2020-05-03
Fix rebase false

mechanicalnull:
The docstring for BinaryView.rebase() says it returns None on failure, but if you try a rebase with the GUI it returns False. This fixes that.


bpotchik:
It looked as if there were multiple commits above so I'm soft accepting this pull request. I fixed the return value as well as the documentation formatting and pushed to dev. Thanks for reporting!

PR #1794, opened 2020-07-01
Add setters for Segment attributes

nshp:


psifertex:
Looks good, thanks!

psifertex:
So, after some discussion internally it turns out those core APIs shouldn't have been exposed and are unlikely to work as expected. I'm currently reverting this commit and we'll be removing those APIs as well until we can sort them out properly. Sorry about the inconvenience! I've created #1795 to track the issue.

nshp:
Is that true for all of those properties, or just the offsets and lengths? The flag setters seemed to work for me, at least superficially

bpotchik:
Yes it's true for all of those setters. For now the correct way to update Segment/Section properties is to grab the object, update the property, and resubmit the object via add_user_[segment|section].

PR #1818, opened 2020-07-16
Implement comparison operators on HighLevelILInstruction class

### LINKED ISSUE #1659: Python HLIL instruction should implement lt, le, gt, ge
LLIL: https://github.com/Vector35/binaryninja-api/blob/dev/python/lowlevelil.py#L942-L960
MLIL: https://github.com/Vector35/binaryninja-api/blob/dev/python/mediumlevelil.py#L375-L394
HLIL: n/a

This breaks @withzombies's bnil graph here: https://github.com/withzombies/bnil-graph/blob/master/__init__.py#L202 because HLIL instructions are not sortable. I'd PR it, but as far as I can tell it's a literal copy/paste of your own code to fix it.

withzombies:
Closes #1659

### DIFF
--- a/python/highlevelil.py
+++ b/python/highlevelil.py
@@ -351,6 +351,26 @@ class HighLevelILInstruction(object):
 			return NotImplemented
 		return not (self == other)
 
+	def __lt__(self, other):
+		if not isinstance(other, self.__class__):
+			return NotImplemented
+		return self._function == other._function and self._expr_index < other._expr_index
+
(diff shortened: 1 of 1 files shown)


plafosse:
Hmm, I'm pretty sure Rusty didn't add those for a reason, so we may not be able to accept this one.

withzombies:
Why'd you wait 2 months to say that? #1659 was open on May 12th.

PR #1839, opened 2020-07-21
change submodule remote from https:// to git@

jeffli678:


rssor:
Better to do this via `.gitconfig` so that the API repo can clone for people without github accounts.

```
[url "ssh://git@github.com/"]
    insteadOf = https://github.com/
```

PR #1842, opened 2020-07-21
log_to_stderr when running in headless mode

verylazyguy:


rssor:
So this change on its own is fine, but accepting it before fixing something else related to Python binding init would cause problems for some use-cases, so we'll need to change some of the neighboring code slightly before accepting this.

bpotchik:
Fixed in 2.1.2396-dev.

PR #1857, opened 2020-07-30
Fix str/bytes conversions for DataBuffers

nshp:
This makes it safe to pass a `bytes` into several functions like `bv.find_next_data`.

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -2652,13 +2652,9 @@ class BinaryView(object):
 			>>> bv.read(0,4)
 			'AAAA'
 		"""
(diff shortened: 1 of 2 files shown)


KyleMiles:
`if isinstance(data, int) or isinstance(data, numbers.Integral):` isn't a comprehensive check; I think it would be better to explicitly check if it's `bytes`, `bytearray`, or sometimes a string and do the appropriate conversion.

I'm going to close this in favor of some other work (https://github.com/Vector35/binaryninja-api/commit/fb72d65faa0df025cc5a5aaaeca850b401ab1d93) and to also accommodate #1959.

PR #1911, opened 2020-08-17
Adds documentation for UIDF

chinmaydd:
This PR adds documentation for user-informed dataflow.

### DIFF
--- a/python/function.py
+++ b/python/function.py
@@ -591,6 +591,12 @@ class PossibleValueSet(object):
 
 		:param list(ValueRange) ranges: List of ValueRanges
 		:rtype: PossibleValueSet
+		:Example:
+			
+			>>> v_1 = ValueRange(-5, -1, 1)
+			>>> v_2 = ValueRange(7, 10, 1)
+			>>> val = PossibleValueSet.signed_range_value([v_1, v_2])
+			<signed ranges: [<range: -0x5 to -0x1>, <range: 0x7 to 0xa>]>
 		"""
 		result = PossibleValueSet()
 		result.value = 0
@@ -606,6 +612,12 @@ class PossibleValueSet(object):
 
 		:param list(ValueRange) ranges: List of ValueRanges
 		:rtype: PossibleValueSet
+		:Example:
+
+			>>> v_1 = ValueRange(0, 5, 1)
+			>>> v_2 = ValueRange(7, 10, 1)
+			>>> val = PossibleValueSet.unsigned_range_value([v_1, v_2])
+			<unsigned ranges: [<range: 0x0 to 0x5>, <range: 0x7 to 0xa>]>
 		"""
 		result = PossibleValueSet()
 		result.value = 0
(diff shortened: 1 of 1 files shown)


python/function.py
```diff
@@ -2603,6 +2684,13 @@ def set_user_var_value(self, var, def_addr, value):
 		core.BNSetUserVariableValue(self.handle, var_data, def_site, value._to_api_object())
 
 	def clear_user_var_value(self, var, def_addr):
+		"""
+		Clears a perviously informed user variable value.
```
psifertex, line 2688: Typo. Also, for consistence maybe "defined user variable value" instead of "informed"?

python/function.py
```diff
@@ -2582,6 +2642,27 @@ def is_call_instruction(self, addr, arch=None):
 		return core.BNIsCallInstruction(self.handle, arch.handle, addr)
 
 	def set_user_var_value(self, var, def_addr, value):
+		"""
+		`set_user_var_value` allows the user to specify a PossibleValueSet value for a (MLIL) variable at its
```
psifertex, line 2646: `value for an MLIL variable at its`...

instead of the parenthetical probably is better.

python/function.py
```diff
@@ -592,6 +644,14 @@ def not_in_set_of_values(self, values):
 
 	@classmethod
 	def lookup_table_value(self, lookup_table, mapping):
+		"""
+		Create a PossibleValueSet object for a value which is a member of a 
+		lookuptable.
+
+		:param list(LookupTableEntry) lookup_table: List of table entries
+		:param dict of (int, int) mapping: Mapping used for resolution
```
psifertex, line 652: Did you double check this is the pydoc formatting for the dict? I guess it doesn't matter too much since we'll move to proper type annotations soon.

python/function.py
```diff
@@ -532,32 +537,61 @@ def count(self, value):
...
 		result.type = RegisterValueType.StackFrameOffset
 		result.value = value
 		return result
 
 	@classmethod
 	def signed_range_value(self, ranges):
+		"""
+		Create a PossibleValueSet object for a signed range of values.
+
+		:param list(ValueRange) ranges: List of ValueRanges
```
psifertex, line 592: An example here would be nice since it's the only type not immediately obvious from the existing context.

python/function.py
```diff
@@ -567,6 +601,12 @@ def signed_range_value(self, ranges):
 
 	@classmethod
 	def unsigned_range_value(self, ranges):
+		"""
+		Create a PossibleValueSet object for a unsigned signed range of values.
+
+		:param list(ValueRange) ranges: List of ValueRanges
```
psifertex, line 607: See above about an example, same here, could use almost the same example but one showing the differences would be good.

python/function.py
chinmaydd: I tried to follow the format as was written [https://github.com/Vector35/binaryninja-api/blob/dev/python/binaryview.py#L3662](https://github.com/Vector35/binaryninja-api/blob/dev/python/binaryview.py#L3662). Validated - https://stackoverflow.com/questions/26712533/formatting-python-docstrings-for-dicts

python/function.py
chinmaydd: Right I can add an example. But we dont currently perform any processing of the `PossibleValueSet` at the API level and hence cant really showcase the difference of how `SignedRangeValue`s and `UnsignedRangeValue`s are processed.

PR #1962, opened 2020-09-09
Add tests for UIDF

chinmaydd:


python/function.py
```diff
@@ -377,10 +380,18 @@ def __repr__(self):
 		return "<undetermined>"
 
 	def __eq__(self, other):
-		if self.type in [RegisterValueType.ConstantValue, RegisterValueType.ConstantValue] and isinstance(other, numbers.Integral):
```
plafosse, line 380: You should probably add this as its the most idiomatic way of implementing a __eq__ override. This also gives "other.__eq__" a chance to handle the same comparison.
```
		if not isinstance(other, self.__class__):
			return NotImplemented
```

suite/testcommon.py
```diff
@@ -883,6 +883,56 @@ def get_functions(self, bv):
     def get_comments(self, bv):
         return bv.functions[0].comments
 
+    def test_possiblevalueset_parse(self):
+        """ Failed to parse PossibleValueSet from string"""
+        file_name = self.unpackage_file("helloworld")
+        try:
+            bv = binja.BinaryViewType.get_view_of_file(file_name)
```
plafosse, line 890: use 
```
with binja.BinaryViewType.get_view_of_file(file_name) as bv:
```
or better yet
```
with binja.open_view(file_name) as bv:
```
as these implicitly execute `bv.file.close()` when the context ends.

suite/testcommon.py
```diff
@@ -1205,3 +1255,168 @@ def test_univeral_loader(self):
         finally:
             binja.Settings().set_string_list("files.universal.architecturePreference", save_setting_value)
             self.delete_package("fat_macho_9arch")
+
+    def test_user_informed_dataflow(self):
+        """User-informed dataflow tests"""
+        file_name = self.unpackage_file("helloworld")
+        try:
+            with binja.BinaryViewType['ELF'].open(file_name) as bv:
```
plafosse, line 1263: slightly easier
```
with binja.open_view(file_name) as bv:
```

plafosse: This also means you don't need to do `bv.update_analysis_and_wait()`

suite/testcommon.py
```diff
@@ -1205,3 +1255,168 @@ def test_univeral_loader(self):
...
+                    def_ins_idx = func.mlil.get_instruction_start(def_site)
+                    def_ins = func.mlil[def_ins_idx]
+
+                    assert(def_ins.get_possible_reg_values_after('r3') == value)
+
+                    temp_name = next(tempfile._get_candidate_names()) + ".bndb"
+                    bv.create_database(temp_name)
+                    bv.file.close()
+
+                with binja.FileMetadata(temp_name).open_existing_database(temp_name).get_view_of_type('ELF') as bv:
```
plafosse, line 1397:
```
with binja.open_view(temp_name) as bv:
```
works here too. ( I realize the unit tests might not follow this convention but this is now preferred)
If you do this then no need for `bv.file.close()` and no need for `update_analysis_and_wait()`

plafosse (changes requested):
The code looks good. No correctness issues I can see, just minor API usage improvements.

PR #2010, opened 2020-10-02
Documentation fixes

withzombies:
* Fixed several incorrect docstrings
* Fixed a bug where `create_user_data_tag` would sometimes return None when it shouldn't


psifertex:
Want that t-shirt, huh? 😛 

Thanks.

PR #2013, opened 2020-10-03
Return tag in all cases

withzombies:
There's several copies of the create_*_tag function in `binaryview.py` and `function.py`:

```python
def create_auto_data_tag(self, addr, type, data, unique=False):
	"""
	``create_auto_data_tag`` creates and adds a Tag object at a data address.

	:param int addr: Address at which to add the tag
	:param TagType type: Tag Type for the Tag that is created
	:param str data: Additional data for the Tag
	:param bool unique: If a tag already exists at this location with this data, don't add another
	:return: The created Tag
	:rtype: Tag
	"""
	if unique:
		tags = self.get_data_tags_at(addr)
		for tag in tags:
			if tag.type == type and tag.data == data:
				return

	tag = self.create_tag(type, data, False)
	core.BNAddAutoDataTag(self.handle, addr, tag.handle)
	return tag
```

In the `unique=True` case, it only creates a Tag if it'd be a unique Tag. It returns `None` otherwise. The docstring says this should return a Tag in all cases, but the docstrings rarely mention if it's meant to be an optional return value or not. As the code is written, this returns an `Optional[binaryninja.Tag]`. 

This PR changes it to return a copy of the matching Tag if the tag is not unique and `unique=True` is specified. This seems to be what's indicated by the docstrings (aka return type is `binaryninja.Tag`.

However, from an ergonomics standpoint, maybe we do want it to return an `Optional[binaryninja.Tag]`. When calling the function we could do something like this:
```python
tag1 = bv.create_auto_tag(addr = 1234, type = some_tag_type, data = "zomg", unique=True)
tag2 = bv.create_auto_tag(addr = 1234, type = some_tag_type, data = "zomg", unique=True)
if tag2 is None:
   print("Tag already existed")
   # If i want that tag, I need to search for it in the tag list now
```

However, maybe we don't care if tags exist already or not? If we're setting `unique=True` maybe we just want a copy of the reference to the existing tag:
```python
tag1 = bv.create_auto_tag(addr = 1234, type = some_tag_type, data = "zomg", unique=True)
tag2 = bv.create_auto_tag(addr = 1234, type = some_tag_type, data = "zomg", unique=True)
assert(tag1 == tag2)            # true
assert(id(tag1) == id(tag2)) # true 
```

I think we should go the route specified in the docstrings and in this PR.

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -3791,7 +3791,7 @@ class BinaryView(object):
 			tags = self.get_data_tags_at(addr)
 			for tag in tags:
 				if tag.type == type and tag.data == data:
-					return
+					return tag
 
 		tag = self.create_tag(type, data, True)
 		core.BNAddUserDataTag(self.handle, addr, tag.handle)
--- a/python/function.py
+++ b/python/function.py
@@ -1352,7 +1352,7 @@ class Function(object):
 			tags = self.get_address_tags_at(addr, arch)
 			for tag in tags:
 				if tag.type == type and tag.data == data:
-					return
+					return tag
 
 		tag = self.create_tag(type, data, True)
 		core.BNAddUserAddressTag(self.handle, arch.handle, addr, tag.handle)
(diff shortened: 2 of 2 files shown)


CouleeApps (approved):
Yep, returning the tag is how these functions should behave.

PR #2017, opened 2020-10-07
Revert the tag api change introduced in 430bf12

withzombies:
Pull the tag change from 430bf12 into #2013 so it can be discussed there.

Currently the `dev` API is half-wrong or half-fixed (depending on your perspective)


CouleeApps:
Since the other PR was intended, this is unnecessary

PR #2029, opened 2020-10-14
Update lowlevelil.py

dfraze:
Add string handling to ILRegister comparison.

This is especially nice with the CallingConvention API. It returns a string for things like "int_return_reg". This allows comparison without sprinkling of str().

### DIFF
--- a/python/lowlevelil.py
+++ b/python/lowlevelil.py
@@ -61,9 +61,18 @@ class ILRegister(object):
 		return self._name
 
 	def __eq__(self, other):
-		if not isinstance(other, self.__class__):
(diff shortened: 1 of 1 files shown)


plafosse:
Similar code would be needed in the ``__ne__`` operator too.

plafosse:
perhaps in the ``__ne__`` it could be ``if not isinstance(other, self.__class__) and not isinstance(other, str):``

KyleMiles:
I just squashed the two commits and changed your whitespace... It's merged in at f6a8cb229b849fe2e946c9d92516dfbce293adf0.

PR #2215, opened 2021-01-04
Set x64 max_instr_len to 15

sabastiaan:
The maximum instruction length for x86-64 is defined to 15, not 16. 
While you can validly encode instructions longer than 15 bytes (infinitely actually), these are defined to be invalid by both Intel and AMD for all of their processors.

### DIFF
--- a/python/architecture.py
+++ b/python/architecture.py
@@ -110,7 +110,7 @@ class Architecture(with_metaclass(_ArchitectureMetaClass, object)):
 	address_size = 8
 	default_int_size = 4
 	instr_alignment = 1
-	max_instr_length = 16
+	max_instr_length = 15
 	opcode_display_length = 8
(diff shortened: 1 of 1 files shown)


psifertex:
This is not the right place for this change. First, this is just in python, and it will be overridden by the initializer. Second, there are many other architectures besides x86. If you really want to change this for x86, you'd want to do it in the architecture submodule, not here where it impacts all architectures. Third, there's actually a potential that this decreases performance because it isn't on a power of two.
Finally, this doesn't really impact anything. All this controls is how many bytes are handed to the architecture, it's still up to the architecture to decide how many bytes are part of the valid instruction so there's really no actual point to changing it. If you're seeing errors with the disassembly of 16-byte long instructions, then they are errors in the architecture submodule, not in this constant.

PR #2223, opened 2021-01-07
Update comment for has_data_variables

carsonharmon:
It looked like this comment was copied over from the `has_functions` function, so I just made a tweak.


psifertex:
Yup, good catch, thanks.

Issue #2236, opened 2021-01-19
Add ability to pass in a prefix to the logging subsystem


gkprow-runsafe:
Current Binja release:
Version 2.2.2603-dev (Build ID 8f9fea38)

Current functionality:
The logging API allows users to limit which LogLevels should be written to stdout/err, to enable writing logs to a file, etc. But, when a log message is written to stdout/err or a file, only the message itself is written. No metadata is written along with the message (e.g., no timestamps, no PID, etc...).

Desired functionality:
There are cases where it would be helpful to supply a custom prefix to the log messages written either by the user (e.g., `bn.log.log()`) or by Binja's core lib.

Here are a couple of approaches for your consideration:
1. Enhance the log API to allow the user to supply a callback for writing the logs. The user API would receive a string parameter that is the log message Binja wants to write. The user callback would simply pass that into print along with any other info the user wants to print.
2. Enhance the log API to receive a function that returns a string, or allows a user to specify a literal string, to prefix to the log written by the core lib.

nshp:
This is already how the core API works:

gkprow-runsafe:
Sure thing. I'll work to create a PR.

plafosse:
Changing title to reflect actual issue.

PR #2241, opened 2021-01-22
Add LogListener class that wraps BNLogListener and associated functions

gkprow-runsafe:
Implements solution for https://github.com/Vector35/binaryninja-api/issues/2236

### DIFF
--- a/python/log.py
+++ b/python/log.py
@@ -19,12 +19,30 @@
 # IN THE SOFTWARE.
 
 
+import ctypes
+
+
 # Binary Ninja components
 from binaryninja import _binaryninjacore as core
 from binaryninja.enums import LogLevel
 
 
 _output_to_log = False
+
+
+class LogListener(object):
+	def __init__(self, log, close, get_log_level):
+		self._listener = core.BNLogListener()
+		self._listener.context = 0
+		self._listener.log = self._listener.log.__class__(log)
+		self._listener.close = self._listener.close.__class__(close)
+		self._listener.getLogLevel = self._listener.getLogLevel.__class__(get_log_level)
+
+	def register(self):
+		core.BNRegisterLogListener(ctypes.pointer(self._listener))
+
+	def unregister(self):
+		core.BNUnregisterLogListener(ctypes.pointer(self._listener))
 
 
 def redirect_output_to_log():


python/log.py
```diff
@@ -27,6 +30,21 @@
 _output_to_log = False
 
 
+class LogListener(object):
+	def __init__(self, log, close, get_log_level):
+		self._listener = core.BNLogListener()
+		self._listener.context = 0
+		self._listener.log = self._listener.log.__class__(log)
```
nshp, line 37: This is in no way official code review, just my opinion. I think these should be required methods on the LogListener class, rather than parameters to __init__. It'd just be a bit more idiomatic to subclass LogListener and implement some methods rather than instantiating a LogListener with functions as parameters, IMO.

gkprow-runsafe: That makes sense. What do you mean by "required methods"? Do you mean stub the methods in the LogListener class and have them raise an exception or return NotImplemented or pass or something else?

gkprow-runsafe: Thanks for the suggestion. It results in a cleaner solution.

gkprow-runsafe: Should the `LogListener` base class include and reference private methods when registering a log listener in the core lib? For instance, should it do something similar to:
```python
class LogListener(object):
        """Base class for custom LogListener classes"""

        def __init__(self):
                ...
                self._listener.log = self._listener.log.__class__(self._log)

        ...
        def _log(self, ctxt, log_level, msg):
                try:
                        self.log(ctxt, log_level, msg)
                except:
                        log_error(traceback.format_exc())
```

My current implementation doesn't do that. Notice that the `except` block in `_log` calls `log_error(...)`. That is a common design pattern throughout the API, but is, I believe, problematic in the `LogListener` class proposed in this PR when the user's class handles `LogLevel.ErrorLog`.

Is there API to write a message to stderr that doesn't depend on the logging sub-system (i.e., it doesn't reference log levels, etc.)? In this situation, should the Python class avoid going through the core lib and instead just write to stderr?

psifertex:
So we had a chat about this during our weekly meeting and there's some concerns that this implementation is likely to have very serious performance implications. We do a lot of log_debug messages internally and putting a python callback into all of those would dramatically slowdown normal execution. We're debating between making sure we add a giant warning across the API documentation or even maybe making this not a part of the core API but rather a separate example that we can point people toward if they need the capability. Just a heads up that we're not ignoring it, just concerned about the implementation and debating how to make it available.

Have you been using the callback for moderate binaries to test the performance impact yet?

plafosse:
To be clear it's not your implementation we have a problem with. It's each analysis thread in the core having to wait on python's GIL which is the problem. We would have the ability to refactor things so they're still fast, but as of now it would essentially make multi-threading useless. This wouldn't be a problem if the solution were implemented in reentrant C/C++ or Rust.

PR #2242, opened 2021-01-23
Changed SymbolType to Type in define_auto_symbol_and_var_or_function

ZetaTwo:
The documentation and error message is currently wrong and claims the second argument "sym_type" should be a "SymbolType" while in reality it should be a "Type".

Should/can I add type hints as well?

PS. First binja PR 🎉


psifertex:
Thanks for the PR!

PR #2276, opened 2021-02-18
Fix panic when building with recent rust versions.

turbocool3r:
Currently a plugin that calls `register_architecture()` when loaded by BN because rustc doesn't allow memory allocated by `zeroed()` to be uninitialized. This change fixes that by using `MaybeUninit` instead of `zeroed()`.


KyleMiles:
Can you possibly share the plugin you're working on so I can test this?  Not having an arch to test is what's preventing me from approving this.

PR #2280, opened 2021-02-24
Add is_eligible_for_heuristics to builder

jrozner:
The ConventionBuilder doesn't expose a method to set is_eligible_for_heuristics.


KyleMiles (approved):
Thank you, I missed this.

PR #2299, opened 2021-03-06
Rust API - Implement BackgroundTask

joshwatson:
Implements the BackgroundTask API in Rust.


KyleMiles (changes requested):
Make sure to run `cargo fmt` if you haven't. Besides where I marked, looks good.

Thank you for helping flesh out the Rust bindings!

KyleMiles (approved):

PR #2322, opened 2021-03-15
Implement Platform::parse_types_from_source

joshwatson:


rust/src/platform.rs
```diff
@@ -230,6 +232,101 @@ impl Platform {
...
+
+        let src = source.as_bytes_with_nul();
+        let filename = filename.as_bytes_with_nul();
+        let auto_type_source = auto_type_source.as_bytes_with_nul();
+
+        let mut include_dirs = vec![];
+
+        for dir in include_directories.iter() {
+            let d = dir.as_ref().to_string_lossy().as_bytes_with_nul();
+            include_dirs.push(d.as_ptr() as *const i8);
```
KyleMiles, line 283: `as *const i8` -> `as _`

KyleMiles (changes requested):
And pls squash

joshwatson:
ok I think I have squashed and fixed issues 😃

KyleMiles:
What's there now seems like the original version

KyleMiles:
Ah, that would be because GitHub had a big red "Refresh" button at the top.

KyleMiles:
Okay, only the `.as_ref() as *const [u8] as *const i8` -> `.as_ref().as_ptr() as _` remains.

rust/src/platform.rs
```diff
@@ -230,6 +232,101 @@ impl Platform {
...
+        }
+
+        unsafe {
+            BNParseTypesFromSource(
+                self.handle,
+                src.as_ref() as *const [u8] as *const i8,
```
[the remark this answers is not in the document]
joshwatson: doh I completely missed those two

rust/src/platform.rs
```diff
@@ -230,6 +232,101 @@ impl Platform {
...
+            typeCount: 0,
+            variableCount: 0,
+            functions: ptr::null_mut(),
+            types: ptr::null_mut(),
+            variables: ptr::null_mut(),
+        };
+
+        let mut type_parser_result = TypeParserResult::default();
+
+        let mut error_string: *mut i8 = ptr::null_mut();
```
KyleMiles, line 273: This actually needs to be a `raw: *mut raw::c_char` per `string::BnString::raw`, and then you can take `BnString::from_raw(...)`; the error strings are left for us to deallocate, plus we need to account for the ARM char difference.

Alternatively, we could just not handle the errors and the fifth argument to `BNParseTypesFromSource` can be a nullptr.  I'm okay with either solution.

KyleMiles (changes requested):
Sorry, this should be the last one.  Beware the `i8`!

KyleMiles (approved):

PR #2355, opened 2021-04-06
Fix docstring confusing some syntax highlighters

KevOrr:
This is technically correct (and for illustration):

```py
>>> "asdf"""
'asdf'
>>> "asdf""qwer"
'asdfqwer'
>>> def foo():
...   "asdf"""
...   return 4
...
>>> foo()
4
>>> foo.__doc__
'asdf'
```

But as it currently exists, it messes up some (admittedly, incorrect) lexers/syntax highlighters (e.g. emacs's python-mode).

### DIFF
--- a/python/function.py
+++ b/python/function.py
@@ -1436,7 +1436,7 @@ class Function(object):
 
 	@property
 	def highest_address(self):
-		"The highest virtual address contained in a function."""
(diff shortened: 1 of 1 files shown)


psifertex:
Looks good, thanks.

PR #2379, opened 2021-04-19
Add View Type to BNIsBackedByDatabase

marpie:
The call to `BNIsBackedByDatabase` is missing the `binaryViewType` parameter.

This MR fixes that and enables the *dwarfdump* example to build.


KyleMiles (approved):
Thank you!

PR #2380, opened 2021-04-19
Fix wrong CorePluginABIVersion, plugin_abi_version, and plugin_abi_minimum_version

marpie:
The current Rust lib incorrectly defines the functions needed to register ABI versions.

As I'm not familiar with rustgen I cannot provide the right fix to generate `uitypes.h` and retrieve the variables `BN_MINIMUM_UI_ABI_VERSION` and `BN_CURRENT_UI_ABI_VERSION`.

With this fix the included plugins build and run with the latest BinaryNinja dev release.


KyleMiles (changes requested):
The flowgraph changes are good, the lib.rs changes aren't.  `plugin_abi_version` and `plugin_abi_minimum_version` are correct as they were, and only `UIPluginABIVersion` needs the band-aid patch to return 2.

marpie:
A yeah, you are right. If you want, you can close this PR and integrate the changes yourself; I'm not at my dev station right now and cannot make the changes till later this week.

KyleMiles (approved):

PR #2393, opened 2021-04-25
Rust bindings fix

yrp604:


KyleMiles (approved):
Thanks for sorting this out.

PR #2567, opened 2021-07-27
support for repository plugins loaded from a subdirectory

psifertex:
support for plugins in subdirectories


python/pluginmanager.py
```diff
@@ -49,6 +49,11 @@ def path(self):
 		"""Relative path from the base of the repository to the actual plugin"""
 		return core.BNPluginGetPath(self.handle)
 
+	@property
+	def subdir(self):
```
plafosse, line 53: Minor but you need to add a type hint `def subdir(self) -> str:`

PR #2568, opened 2021-07-27
vars_read and vars_written property for HighLevelILInstruction

corinnedixon:
I was hoping to add these MLIL properties into HLIL as well. Let me know if there are any other HLIL Operations that should be included.

### DIFF
--- a/python/highlevelil.py
+++ b/python/highlevelil.py
@@ -415,6 +415,40 @@ class HighLevelILInstruction(object):
 				result.append(operand)
 		result.append(HighLevelILOperationAndSize(self._operation, self._size))
 		return result
+	
+	@property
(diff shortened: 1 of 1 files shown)


plafosse:
Thank you for your contribution! Unfortunately we're in the middle of a big refactor to this code, so we're probably not going to accept this code as it will be much cleaner in the new version of the code. :)

cetfor:
Hey Peter, do you know if this refactor something you're planning for v2.5 or a later release?

psifertex:
The refactor is planned for 3.0 which is the very next release.

plafosse:
Again thanks for the PR. This feature now exists in the dev version

PR #2615, opened 2021-09-08
Fix order of arguments to `RegisterInfo` constructor

tetsuo-cpp:
In [this commit](https://github.com/Vector35/binaryninja-api/commit/920da93a153ff0e6cff405c76f6d5175a9cbcfa4), `RegisterInfo` got converted to a data class. This changed the order of arguments in `__init__` because the previous constructor had the fields in a different order to the way that they're listed in the class.


plafosse:
Rejecting this PR as the real issue is that I ordered the items in the data class wrong, and I even gave the wrong argument a default value.

plafosse:
Thanks for the PR though helped me find and fix a bug :)

PR #2616, opened 2021-09-08
Fix operand index usages in LLIL definitions

tetsuo-cpp:


python/lowlevelil.py
```diff
@@ -1474,7 +1474,7 @@ def dest(self) -> SSARegisterStack:
 
 	@property
 	def src(self) -> SSARegisterStack:
-		return self.get_reg_stack_ssa(0, 2)
+		return self.get_reg_stack_ssa(2, 3)
```
plafosse, line 1477: Hmmm I'm unsure about this one. The original code for this looked like
```
reg_stack = ILRegisterStack(func.arch, instr.operands[i])
i += 1
value = SSARegisterStack(reg_stack, instr.operands[i])
i += 1
self._operands.append(value)
self.dest = value
value = SSARegisterStack(reg_stack, instr.operands[i])
```
Which suggests the register stack is in operand `0` and `dest` is in `1` and  and the `src` is in operand `2` I'm struggling to actually find a test cases that covers this code. Do you have an example binary?

tetsuo-cpp: Yeah, my bad. I've reverted that commit out.

l didn't see this instruction causing problems. I was just eyeballing this area while making fixes and thought it looked wrong.

plafosse: I experimentally verify the original code to be correct thanks For all the other fixes￼!

plafosse:
Turns out a couple of the other ones were in error too. There were two that were very much bugs. Which have been fixed. Thanks.

PR #2617, opened 2021-09-08
Avoid using `isinstance` on objects returned by `NewType`

tetsuo-cpp:
I've been seeing some errors on these checks: `TypeError: isinstance() arg 2 must be a type or tuple of types`.

According to the docs [here](https://mypy.readthedocs.io/en/stable/more_types.html), this isn't supported:

> You cannot use isinstance() or issubclass() on the object returned by NewType(), because function objects don’t support these operations. You cannot create subclasses of these objects either.

I've edited these lines to use the underlying type. The other thing you could do is something like [this](https://github.com/python/mypy/issues/3325#issuecomment-299546004). Happy to do it like that if that's preferable.

### DIFF
--- a/python/architecture.py
+++ b/python/architecture.py
@@ -1440,7 +1440,7 @@ class Architecture(metaclass=_ArchitectureMetaClass):
 				log_error(traceback.format_exc())
 		elif isinstance(reg, lowlevelil.ILRegister):
 			return reg.index
-		elif isinstance(reg, RegisterIndex):
+		elif isinstance(reg, int):
 			return reg
 		raise Exception("Attempting to get register index of non-existant register")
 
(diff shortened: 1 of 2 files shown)


plafosse:
This PR is needed yet insufficient. We're using PyRight type checking internally. These checks were added to satisfy the type checker which is apparently not realizing that that syntax isn't acceptable.  To satisfy the type checker the code needs to be change from this
```
		if isinstance(flags, architecture.FlagWriteTypeName):
			_flags = self.arch.get_flag_write_type_by_name(flags)
```
```
		if isinstance(flags, str):
			_flags = self.arch.get_flag_write_type_by_name(architecture.FlagWriteTypeName(flags))
```

tetsuo-cpp:
@plafosse 
Understood! I've made those changes now. Can you let me know whether PyRight is happy with the patch?

plafosse:
Thanks for the PR, now fixed in dev

PR #2636, opened 2021-09-14
Return a `SemanticClassIndex` of zero for a `None` argument

tetsuo-cpp:
There are a few spots where we're leaving the index as `None` and then proceeding to pass it to a C function, resulting in a type error since you can't pass `None` to a function taking `ctype.u_int`. This error also seems to be causing Binja itself crash somehow.

Prior to [this change](https://github.com/Vector35/binaryninja-api/commit/61b4bb24e06aa955484293d35fa926c07887544b), the API was using 0 for when the the class type argument was `None` so I assume that's what we want.

### DIFF
--- a/python/architecture.py
+++ b/python/architecture.py
@@ -1465,7 +1465,9 @@ class Architecture(metaclass=_ArchitectureMetaClass):
 			return flag
 		raise Exception("flag is not convertable to index")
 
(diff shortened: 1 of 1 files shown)


plafosse:
Thanks for all these PRs! I will be looking at them and hopefully accepting them today.

plafosse:
Thanks for the PR I need to do some modifications to it to satisfy the type checker so I just committed it directly. Thanks.

PR #2644, opened 2021-09-16
Fix handling of `FlagIndex` in `LowLevelILFunction.expr`

ehennenfent:
Several functions, including `LowLevelILFunction.set_reg`, [default to `architecture.FlagIndex(0)` if no flags are explicitly provided](https://github.com/Vector35/binaryninja-api/blob/dev/python/lowlevelil.py#L3112-L3113). However, `LowLevelILFunction.expr` did not recognize `FlagIndex` as a valid type for flag arguments, and would thus [throw an assertion error](https://github.com/Vector35/binaryninja-api/blob/dev/python/lowlevelil.py#L3048-L3055)


plafosse:
Actually needs to use isinstance int. See: https://github.com/Vector35/binaryninja-api/pull/2617 I have a local fix for this that will be pushed shortly. Thank you for the PR

PR #2678, opened 2021-10-05
TriageView: path, size, md5, sha1, sha256

rose4096:


plafosse (approved):
Code looks good, I have no corrections.

PR #2689, opened 2021-10-11
Allow TriageView to be built standalone.

rose4096:
This allows TriageView to be built from dev or from BN_API_BUILD_EXAMPLES.


CouleeApps:
This was addressed in a previous commit: b8af211f9a3647687c7e45c0fd04b0768df87ffe

PR #2734, opened 2021-10-22
Fix TypedDataReader member accesses

joshwatson:
TypedDataReader failed to correctly create a TypedDataReader for structure members because it was just passing the offset rather than self.address+m.offset. This PR fixes that mistake.


plafosse:
Thanks for the fix I just made the change directly instead of accepting your PR

Issue #1910, opened 2020-08-15
Undefining a function should remove associated symbol


yrp604:
I have a bndb with some code I'm trying to understand. As part of this process, I'm moving a function boundary between two chunks of code. For example, if I have:

```
0x1000: func_a
...
0x2000: func_b
```

If I want to move `func_b` to start at 0x1800, my workflow is to create a new function at 0x1800, undefine `func_b`, then rename `sub_1800` to `func_b`. This mostly works, except it sometimes leaves the old symbol for `func_b`, (especially if I've modified the argument list/type?) in my `bv.symbols`. If I'm moving function boundaries around a lot, these can build up and be annoying.

As a workaround, they can be manually deleted with `bv.undefine_user_symbol(bv.symbols['sym_name'])`, but they should probably be deleted automatically with function undefine?

galenbwill:
I've implemented this as a change to both API and core, as well as UI (with a setting to toggle it, defaulting to off).

For now, I'm leaving it on a test branch [test_remove_user_symbols_with_functions](https://github.com/Vector35/binaryninja-api/tree/test_remove_user_symbols_with_functions), and I'll submit a PR for it.

Since it affects the C API, it needs further review.

fuzyll:
We decided to reject the PR on this for the API because we didn't want to complicate things. But, we still feel this is important to do through the UI, so we're keeping this ticket open to track that.

PR #2752, opened 2021-11-04
added remove symbol option to RemoveUserFunction API (skip-ci)

galenbwill:
Tied to https://github.com/Vector35/binaryninja/pull/269

Addresses https://github.com/Vector35/binaryninja-api/issues/1910


plafosse:
We're rejecting this PR. Instead we'd rather #1910 to do this implicitly in the UI rather than through an API change.

PR #2759, opened 2021-11-12
Fix mismatch in parameter names

cblichmann:
Function parameters should be named the same in both declaration and
definition.


cblichmann:
Friendly ping? :)

plafosse (approved):

plafosse:
Thanks for these fixes :)

PR #2831, opened 2022-01-11
rust: workflow bindings and example

yrp604:
So as far as I can tell the bindings are correct for the workflow types, _however_ there are no bindings for emitting new IL inside the workflow passes. After looking at the existing bindings, there is enough nuance and complication that I don't think I'd be able to write the code that's needed here (or if so it would require extensive v35 help). As a result this code is practically useless. If/when the bindings for rust workflows are written by someone, they'll likely need code that looks very similar if not identical to this (or this would be a good starting spot for that work), but I don't know when that might be.

I'm happy any way this goes, feel free to merge or close.

### DIFF
--- a/rust/examples/workflow/Cargo.toml
+++ b/rust/examples/workflow/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "workflow"
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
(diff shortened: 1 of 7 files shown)


yrp604:
As soon as I published this I realized something else -- Brian mentioned the workflow APIs are still in flux. Since their prototypes are likely to change then closing this probably makes the most sense so as to not break the rust bindings for no benefit. Can always revive it once workflows stabilize.

PR #2838, opened 2022-01-14
rust: add LLIL_TAILCALL bindings

yrp604:
image


yrp604:
This needs to be a separate type, not added onto call.

PR #2840, opened 2022-01-16
data was of type bytes when trying to retrieve input, proper string cast

v1X3Q0:
might have been a python3 change, but need the conversion to string type for data, which was of bytes. this has to be done with decode to maintain the "\n"


plafosse:
Thanks for the PR I've reimplemented this code and the fix will be in the next dev build

v1X3Q0:
not a problem man, its literally one line of code, glad to hear that its in!

PR #2863, opened 2022-01-22
fix: CustomViewContext using 'std::mem::zeroed'

EliseZeroTwo:
This is #2847 reopened as github automatically closed it when I renamed my branch, sorry...

Currently `CustomViewContext` uses `std::mem::zeroed` which can result in a runtime panic when compiling a custom binary view with a later version of Rust.

This MR fixes that by removing the usage of `std::mem::zeroed` in `CustomViewContext`. This is achieved by wrapping the view struct member in an `std::mem::MaybeUninit` and then replacing the call to `std::mem::zeroed` with a call to `std::mem::MaybeUninit::uninit`, and in `cb_init` when assigning the newly initialized view to the struct member it requires a call to `std::mem::MaybeUninit::new`. In places where the member is used it assumes the struct member is initialized, thus the same safety implications as before apply.

### DIFF
--- a/rust/src/custombinaryview.rs
+++ b/rust/src/custombinaryview.rs
@@ -371,7 +371,7 @@ impl<'a, T: CustomBinaryViewType> CustomViewBuilder<'a, T> {
         where
             V: CustomBinaryView,
         {
-            view: V,
+            view: mem::MaybeUninit<V>,
             raw_handle: *mut BNBinaryView,
             initialized: bool,
             args: V::Args,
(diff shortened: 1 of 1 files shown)


KyleMiles:
Could you please provide a code example I can compile to verify these changes?  We could use a minimal example to add to the examples folder/CI (if you're so inclined), otherwise just something quick to verify these changes.  If you don't want to share it publicly, feel free to message @Elyk on Slack.

EliseZeroTwo:
Sure I had already made an MRE to demo it before fixing it: https://gitlab.com/elise/mre-binja-crash

Please feel free to use it as you want, I will quickly add an MIT license now.

KyleMiles:
We can't merge this until we update our version of cargo.  We're currently on 2021-06-17, where `maybe_uninit_ref` is still unstable.

KyleMiles (approved):
Still need to make the internal changes mentioned above, just marking this as approved for book-keeping's sake.

KyleMiles:
This exposed another bug that I just fixed in 9c73b5f2b3f3963bb75ddb602ab5d6985df7ce93.

PR #2865, opened 2022-01-22
rust: feat: add demangle helpers for gnu3 and ms

EliseZeroTwo:
This MR adds helper functions for `demangle_gnu` and `demangle_ms`.

### DIFF
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -17,7 +17,7 @@
 // TODO : Test the get_enumeration and get_structure methods
 
 use binaryninjacore_sys::*;
-use std::{fmt, mem, ptr, result, slice};
+use std::{ffi::CStr, fmt, mem, ptr, result, slice};
 
 use crate::architecture::{Architecture, CoreArchitecture};
 use crate::callingconvention::CallingConvention;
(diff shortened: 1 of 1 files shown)


rust/src/types.rs
```diff
@@ -17,7 +17,7 @@
 // TODO : Test the get_enumeration and get_structure methods
 
 use binaryninjacore_sys::*;
-use std::{fmt, mem, ptr, result, slice};
+use std::{ffi::CStr, fmt, mem, ptr, result, slice};
```
KyleMiles, line 20: This adds the warning

```
warning: the item `CStr` is imported redundantly
    --> src/types.rs:1606:13
     |
20   | use std::{ffi::CStr, fmt, mem, ptr, result, slice};
     |           --------- the item `CStr` is already imported here
...
1606 |         use std::ffi::CStr;
     |             ^^^^^^^^^^^^^^
     |
     = note: `#[warn(unused_imports)]` on by default
```

Could you please remove that import to resolve that warning?

rust/src/types.rs
```diff
@@ -938,6 +938,128 @@ impl Type {
         let mut name = QualifiedName::from(name);
         unsafe { BnStr::from_raw(BNGenerateAutoDemangledTypeId(&mut name.0)) }
     }
+
+    pub fn demangle_gnu3<Mangled: BnStrCompatible>(
```
KyleMiles, line 942: I'm just going to ask that you move these to a new `demangle.rs` to match the structure of the Python/C++ APIs, please.

rust/src/types.rs
```diff
@@ -938,6 +938,128 @@ impl Type {
...
+        arch: &CoreArchitecture,
+        mangled_name: Mangled,
```
KyleMiles, line 944: I'm not sure it's a good convention, but we use `S` for all our `BnStrCompatible`'s.  Until/unless an effort is made to name those better, I'd prefer to stay consistent in the meantime.

rust/src/types.rs
```diff
@@ -938,6 +938,128 @@ impl Type {
...
+            let out_type = match out_type.is_null() {
+                true => {
+                    log::debug!("demangle_gnu3: out_type is NULL");
+                    None
+                }
+                false => Some(Type::from_raw(out_type)),
+            };
+
+            if out_name.is_null() {
+                log::error!("v: out_name is NULL");
```
KyleMiles, line 983: v?

rust/src/types.rs
```diff
@@ -938,6 +938,128 @@ impl Type {
...
     }
+
+    pub fn demangle_gnu3<Mangled: BnStrCompatible>(
+        arch: &CoreArchitecture,
+        mangled_name: Mangled,
+        simplify: bool,
+    ) -> Result<(Option<Type>, Vec<String>)> {
+        let mangled_name_bwn = mangled_name.as_bytes_with_nul();
+        let mangled_name_ptr = mangled_name_bwn.as_ref();
+        unsafe {
```
KyleMiles, line 949: Please make only the unsafe portions labeled as such.

rust/src/types.rs
```diff
@@ -938,6 +938,128 @@ impl Type {
         let mut name = QualifiedName::from(name);
         unsafe { BnStr::from_raw(BNGenerateAutoDemangledTypeId(&mut name.0)) }
     }
+
+    pub fn demangle_gnu3<Mangled: BnStrCompatible>(
+        arch: &CoreArchitecture,
+        mangled_name: Mangled,
+        simplify: bool,
+    ) -> Result<(Option<Type>, Vec<String>)> {
```
KyleMiles, line 946: The Python version of this API gets it wrong; the return value should instead match the C++ version: `Result<(Option<Type>, QualifiedName)>`

rust/src/types.rs
```diff
@@ -938,6 +938,128 @@ impl Type {
...
+                }
+            }
+
+            BNFreeDemangledName(&mut out_name, out_size);
+
+            Ok((out_type, names))
+        }
+    }
+
+    pub fn demangle_ms<Mangled: BnStrCompatible>(
```
KyleMiles, line 1003: Same comments as above.

KyleMiles (changes requested):

KyleMiles:
I'm not sure if I did something wrong when selecting where you put my comments, but all my comments refer to the last line of the preview it's showing above.

rust/src/types.rs
EliseZeroTwo: Ok I will change that here. Is it worth me trying to give all generics better names in another MR?

rust/src/types.rs
EliseZeroTwo: That was me trying to paste and not pressing the command key seemingly... Thank you, sorry for not noticing!

rust/src/types.rs
KyleMiles: I'm gonna say no, based on https://doc.rust-lang.org/book/ch10-01-syntax.html:

> You can use any identifier as a type parameter name. But we’ll use T because, by convention, parameter names in Rust are short, often just a letter, and Rust’s type-naming convention is CamelCase. Short for “type,” T is the default choice of most Rust programmers.

and https://rust-lang.github.io/api-guidelines/naming.html:

> Type parameters | concise UpperCamelCase, usually single uppercase letter: T

KyleMiles:
I've resolved the comments and merged the requested changes in f25acda2073f2d1250814f7a49323d71dff61f9c.

PR #2890, opened 2022-01-28
rust: Add LinearView APIs

fabianfreyer:
🦀

### DIFF
--- a/rust/src/disassembly.rs
+++ b/rust/src/disassembly.rs
@@ -16,8 +16,10 @@
 
 use binaryninjacore_sys::*;
 
-use crate::string::BnString;
+use crate::string::{BnStr, BnString};
 use crate::{BN_FULL_CONFIDENCE, BN_INVALID_EXPR};
+
+use crate::rc::*;
 
 use std::convert::From;
 use std::mem;
@@ -31,6 +33,9 @@ pub struct InstructionTextToken(pub(crate) BNInstructionTextToken);
 // TODO : Consider remodeling this after types::EnumerationMember
 impl InstructionTextToken {
     // TODO : New vs new_with_value ?
+    pub(crate) unsafe fn from_raw(raw: &BNInstructionTextToken) -> Self {
+        Self(raw.clone())
+    }
 
     pub fn new(type_: InstructionTextTokenType, text: &str, value: u64) -> Self {
         let raw_name = BnString::new(text);
@@ -59,6 +64,10 @@ impl InstructionTextToken {
     pub fn set_context(&mut self, context: InstructionTextTokenContext) {
         self.0.context = context;
     }
+
+    pub fn text(&self) -> &BnStr {
+        unsafe { BnStr::from_raw(self.0.text) }
+    }
 }
 
 impl Default for InstructionTextToken {
@@ -79,21 +88,73 @@ impl Default for InstructionTextToken {
     }
 }
 
+impl CoreArrayProvider for InstructionTextToken {
+    type Raw = BNInstructionTextToken;
+    type Context = ();
+}
+
+unsafe impl CoreOwnedArrayProvider for InstructionTextToken {
+    unsafe fn free(raw: *mut BNInstructionTextToken, count: usize, _context: &()) {
+        BNFreeInstructionText(raw, count);
+    }
+}
+
+unsafe impl<'a> CoreArrayWrapper<'a> for InstructionTextToken {
+    type Wrapped = Guard<'a, InstructionTextToken>;
+
+    unsafe fn wrap_raw(raw: &'a Self::Raw, _context: &'a Self::Context) -> Self::Wrapped {
+        Guard::new(InstructionTextToken::from_raw(raw), _context)
+    }
+}
+
 pub struct DisassemblyTextLine(pub(crate) BNDisassemblyTextLine);
 
 impl DisassemblyTextLine {
     // TODO : this should probably be removed, though it doesn't actually hurt anything
     pub fn debug_print(&self) {
+        println!("{}", self);
+    }
+
+    pub fn addr(&self) -> u64 {
+        self.0.addr
+    }
+
+    pub fn instr_idx(&self) -> usize {
+        self.0.instrIndex
+    }
+
+    pub fn count(&self) -> usize {
+        self.0.count
+    }
+
+    pub fn tag_count(&self) -> usize {
+        self.0.tagCount
+    }
+
+    pub fn tokens(&self) -> ArrayGuard<InstructionTextToken> {
+        unsafe { ArrayGuard::new(self.0.tokens, self.0.count, ()) }
+    }
+}
+
+impl std::fmt::Display for DisassemblyTextLine {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let tokens: Vec<InstructionTextToken> =
             unsafe { Vec::from_raw_parts(self.0.tokens as *mut _, self.0.count, self.0.count) };
 
         for token in &tokens {
             let token_string = unsafe { BnString::from_raw(token.0.text) };
-            print!("{}", token_string);
+            let result = write!(f, "{}", token_string);
             token_string.into_raw();
+
+            if result.is_err() {
+                mem::forget(tokens);
+                return result;
+            }
         }
 
         mem::forget(tokens);
+
+        Ok(())
     }
 }
 
@@ -208,5 +269,58 @@ impl Drop for DisassemblyTextLine {
         unsafe {
             Vec::from_raw_parts(self.0.tokens, self.0.count, self.0.count);
         }
+    }
+}
+
+pub type DisassemblyOption = BNDisassemblyOption;
+
+#[derive(PartialEq, Eq, Hash)]
+pub struct DisassemblySettings {
+    pub(crate) handle: *mut BNDisassemblySettings,
+}
+
+impl DisassemblySettings {
+    pub fn new() -> Ref<Self> {
+        unsafe {
+            let handle = BNCreateDisassemblySettings();
+
+            debug_assert!(!handle.is_null());
+
+            Ref::new(Self { handle })
+        }
+    }
+
+    pub fn set_option(&self, option: DisassemblyOption, state: bool) {
+        unsafe { BNSetDisassemblySettingsOption(self.handle, option, state) }
+    }
+
+    pub fn is_option_set(&self, option: DisassemblyOption) -> bool {
+        unsafe { BNIsDisassemblySettingsOptionSet(self.handle, option) }
+    }
+}
+
+impl AsRef<DisassemblySettings> for DisassemblySettings {
+    fn as_ref(&self) -> &Self {
+        self
+    }
+}
+
+impl ToOwned for DisassemblySettings {
+    type Owned = Ref<Self>;
+
+    fn to_owned(&self) -> Self::Owned {
+        unsafe { RefCountable::inc_ref(self) }
+    }
+}
+
+unsafe impl RefCountable for DisassemblySettings {
+    unsafe fn inc_ref(handle: &Self) -> Ref<Self> {
+        Ref::new(Self {
+            handle: BNNewDisassemblySettingsReference(handle.handle),
+        })
+    }
+
+    unsafe fn dec_ref(handle: &Self) {
+        BNFreeDisassemblySettings(handle.handle);
     }
 }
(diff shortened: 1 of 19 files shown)


rust/src/linearview.rs
```diff
@@ -0,0 +1,515 @@
...
+// I don't understand enough of the API to see where this should get freed otherwise.
+/*
+impl Drop for LinearViewLinesIterator {
+    fn drop(&mut self) {
+         unsafe {
+             BNFreeLinearDisassemblyLines(self.lines, self.count);
+        }
+    }
+}
+*/
```
KyleMiles, line 410: I think this should all be replaceable with a `CoreOwnedArrayWrapper`.  That does the fancy `Guard` stuff and should deallocate correctly.

rust/src/linearview.rs
```diff
@@ -0,0 +1,515 @@
...
+        }
+    }
+
+    // FIXME: can we implement clone without shadowing ToOwned?
+    pub fn duplicate(&self) -> Ref<Self> {
+        unsafe {
+            let handle = BNDuplicateLinearViewCursor(self.handle);
+            Self::from_raw(handle)
+        }
+    }
```
KyleMiles, line 258: I don't _think_ so, if I'm understanding you completely.  You get a completely new object in the core, so you have to own the object outside of it as well.  That said, this should probably be moved to a `impl Clone`

rust/src/linearview.rs
```diff
@@ -0,0 +1,515 @@
...
+        debug_assert!(!handle.is_null());
+        Self { handle }
+    }
+}
+
+impl Drop for LinearViewObjectIdentifier {
+    fn drop(&mut self) {
+        unsafe { BNFreeLinearViewObjectIdentifier(self.handle) }
+    }
+}
```
KyleMiles, line 231: I don't think you use this, so you could remove it.  If you wanna have it thought, since it's just a struct:
```
	struct BNLinearViewObjectIdentifier
	{
		char* name;
		BNLinearViewObjectIdentifierType type;
		uint64_t start, end;
	};
```
It should be broken up/deserialized from C.  In this case the Python version is a good reference: `_to_core_struct` (to_raw?) and `_from_api_object` (from_raw?).

rust/src/linearview.rs
```diff
@@ -0,0 +1,515 @@
...
+             BNFreeLinearDisassemblyLines(self.lines, self.count);
+        }
+    }
+}
+*/
+
+pub type LinearDisassemblyLineType = BNLinearDisassemblyLineType;
+
+pub struct LinearDisassemblyLine {
+    r#type: LinearDisassemblyLineType,
```
KyleMiles, line 415: Just use `t`.

rust/src/linearview.rs
KyleMiles: Coming back to say that you already do this for `BNLinearDisassemblyLine`...so this should be similar.

rust/src/linearview.rs
```diff
@@ -0,0 +1,515 @@
...
+
+    pub fn tag_count(&self) -> usize {
+        self.contents.tagCount
+    }
+
+    pub fn tokens(&self) -> LinearDisassemblyLineTokenIterator {
+        LinearDisassemblyLineTokenIterator::new(self.contents.count, self.contents.tokens)
+    }
+
+    pub fn function(&self) -> Ref<Function> {
```
KyleMiles, line 453: It might be better to return a `&Function` which they can `to_owned` if they need to.  I'll accept this as is though, since there are other parts of the API where we do exactly this.

rust/src/linearview.rs
```diff
@@ -0,0 +1,515 @@
...
+        if self.offset >= self.count {
+            return None;
+        }
+
+        let token = unsafe { *self.tokens.offset(self.offset as _) };
+
+        self.offset = self.offset + 1;
+        Some(InstructionTextToken(token))
+    }
+}
```
KyleMiles, line 515: Same note about replacing with a `CoreOwnedArrayWrapper` impl here... If you run in to problems with that, let me know.

rust/examples/decompile/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
+cmake_minimum_required(VERSION 3.9 FATAL_ERROR)
```
KyleMiles, line 1: Unless you're using this, please remove it; We only need it for examples we're building in CI.  We could definitely beef up our Rust API testing, but that should probably have a focused effort later on.

rust/src/disassembly.rs
```diff
@@ -210,3 +216,63 @@ impl Drop for DisassemblyTextLine {
         }
     }
 }
+
+pub use binaryninjacore_sys::BNDisassemblyOption as DisassemblyOption;
```
KyleMiles, line 220: I prefer the type syntax; `pub type DisassemblyOption = BNDisassemblyOption;`

rust/src/disassembly.rs
```diff
@@ -210,3 +216,63 @@ impl Drop for DisassemblyTextLine {
...
+    pub(crate) handle: *mut BNDisassemblySettings,
+}
+
+impl DisassemblySettings {
+    #[allow(unused)]
+    pub(crate) unsafe fn from_raw(handle: *mut BNDisassemblySettings) -> Ref<Self> {
+        debug_assert!(!handle.is_null());
+
+        Ref::new(Self { handle })
+    }
```
KyleMiles, line 233: Comment this out for now; when someone eventually needs it, I'd rather not have the tag lying around.... Unless Cargo proactively warns you that you're using an "allow(unused)"?  Either way, I'd like for it to be loud when we start using this.

KyleMiles (changes requested):

rust/src/linearview.rs
fabianfreyer: This is what I originally intended, but that conflicted with some blanket impl from the refcounting.

rust/examples/decompile/CMakeLists.txt
fabianfreyer: Ah, I just copied this over from the basic example

rust/src/linearview.rs
fabianfreyer: I think the problem is in the when `LinearDisassemblyLine` gets dropped. I'll look at how `CoreOwnedArrayWrapper` works!

rust/src/disassembly.rs
fabianfreyer: Adressed in 8b5ab443ae0d7e29c2080aca3fe9f897433a7f45

rust/src/linearview.rs
fabianfreyer: Addressed in 155886564e007683342ad263b1d939fb804a954c

rust/src/linearview.rs
fabianfreyer: Removed in 23eb57d7874a9b7ffad4fb5d0c4f742eba845b50.

rust/src/disassembly.rs
fabianfreyer: Adressed in d06905d844a74aff66667b7afbc262f6da343ba2

rust/examples/decompile/CMakeLists.txt
fabianfreyer: addressed in 12afe637b60855ccba51d5dfe81b6bfece859534 -- is this enough?

KyleMiles: I meant the whole file

rust/src/linearview.rs
fabianfreyer: FWIW, here's what happens when I try to `impl Clone`:
```
error[E0119]: conflicting implementations of trait `std::borrow::ToOwned` for type `linearview::LinearViewCursor`
   --> /Users/user/Code/binaryninja-api/rust/src/linearview.rs:335:1
    |
335 | impl ToOwned for LinearViewCursor {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: conflicting implementation in crate `alloc`:
            - impl<T> ToOwned for T
              where T: Clone;

For more information about this error, try `rustc --explain E0119`.
error: could not compile `binaryninja` due to previous error
```

rust/src/linearview.rs
fabianfreyer: The allocation issue persists also with `CoreOwnedArrayWrapper`, see 0d66f7b0910792107b60e3c2de550ebfca6f38eb

fabianfreyer:
Note: I'd like to rebase & squash some of the changes before they are merged, keeping them as individual commits for easier review for now.

rust/src/linearview.rs
fabianfreyer: Okay, after doing some digging, I believe the issue is this:
* `BNFreeLinearDisassemblyLines` will iterate over all lines, and call `BNFreeInstructionText` on each line's text.
* We have an `impl CoreOwnedArrayProvider for LinearDisassemblyLine` which (or in the case of the iterator, an `impl Drop`), in its `free` (or `drop`, which calls `free`) method calls `BNFreeLinearDisassemblyLines`, the lines will be freed.
* We have an `impl CoreOwnedArrayProvider for InstructionTextToken` calls `BNFreeInstructionText(raw, count);`.
* When we create an Array over the tokens in a Line, and that array gets dropped, the tokens are already freed.
* When we create an Array over the lines, and that array gets dropped, the lines will be attempted to be freed again, causing a double free.

I can think of the following solutions to this off the top of my head:
* Don't free the tokens in `<InstructionTextToken as CoreOwnedArrayProvider>::free`. This is not a nice option, since there are other scenarios where we would want an `Array<InstructionTextToken>`.
* Extend `CoreOwnedArrayProvider` to take a generic to implement context-sensitive resource deallocation. 
* Use an Iterator over the `InstructionTextToken` but an `Array<LinearDisassemblyLine>`.

rssor: This sounds like an error in the bindings surrounding the InstructionTextToken array then; if data is being handed out of a Array<LinearDisassemblyLine> we shouldn't also be constructing an Array<InstructionTextToken> out of its contents, as Array<> objects have ownership semantics. If a LinearDisassemblyLine itself owns its array of InstructionTextTokens we can't be placing that interior array into an object that will free them out from under the actual owner when it drops.

rssor: This might necessitate the quick creation of a `ArrayGuard<T>` path which is mostly equivalent to `Array` in containing ptr/count/ctxt but without the freeing behavior, similar to `Guard<T>` vs `Ref<T>`.

I'll double check to be sure, it's been several years since I wrote it originally.

fabianfreyer: 8be0b1a4f1e774b397df51b0bac265104f88972e adds an `ArrayGuard` for non-owned arrays by splitting `CoreOwnedArrayProvider` up.

rust/src/lib.rs
```diff
@@ -52,6 +52,7 @@ pub mod filemetadata;
 pub mod flowgraph;
 pub mod function;
 pub mod headless;
+pub mod linearview;
```
debugmen-vergil, line 55: Correct me if I'm wrong, or this was intentional but it appears this file needs the following line:

```pub mod interaction;```

in order to include interaction.rs as a module that can be used when including binaryninja as a cargo dependency?

Comment (fabianfreyer) [reply rust/src/lib.rs:55]: Yes, you are absolutely correct. Seems I dropped that at some point when rebasing things.

**Comment by fabianfreyer** [comment]
I split out the interaction-related commit, which was pretty self-contained into #2948 and squashed this into fewer commits.

**Comment by fabianfreyer** [comment]
Rebased and fixed conflicts

**Comment by fabianfreyer** [comment]
Split out the bindgen bump to #2950

File: rust/src/disassembly.rs [hunk 11]
```diff
@@ -210,3 +271,56 @@ impl Drop for DisassemblyTextLine {
...
+        unsafe {
+            let handle = BNCreateDisassemblySettings();
+
+            debug_assert!(!handle.is_null());
+
+            Ref::new(Self { handle })
+        }
+    }
+
+    pub fn set_option(&mut self, option: DisassemblyOption, state: bool) {
```
KyleMiles, line 293: Make these non-mut please.

fabianfreyer: done.

KyleMiles:
I manually merged this in 6cd49edb317112acc73f7caeb285f4134a472d72.  Made some minor tweaks in 2845ba6208ce3c29998a48df5073ed15a11ead77.  Thank you you for improving the Rust bindings!

PR #2904, opened 2022-01-31
Remove user data tag docstrings for auto tags

fabianfreyer:


KyleMiles (approved):
Thank you for catching this typo

PR #2950, opened 2022-02-08
rust: bump bindgen version to ^0.59

fabianfreyer:
Pinning it directly to 0.58.1 caused issues with dependencies that used
conflicting `proc-macro2` versions. Being more lax in the version allows
consumers of the rust API more chances to resolve dependencies.

### DIFF
--- a/rust/binaryninjacore-sys/Cargo.toml
+++ b/rust/binaryninjacore-sys/Cargo.toml
@@ -5,4 +5,4 @@ authors = ["Ryan Snyder <ryan@vector35.com>", "Kyle Martin <kyle@vector35.com>"]
 build = "build.rs"
 
 [build-dependencies]
-bindgen = "0.58.1"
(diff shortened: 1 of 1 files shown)


psifertex:
We've got some internal dependencies that require this particular version but we're looking into what it would take to increase it. Being lax unfortunately isn't particularly easy either as we've had forward versions break things randomly which is why we pinned it in the first place.

fabianfreyer:
Ah damn. Unfortunately this currently breaks when compiling with anything that depends on `proc-macro2`. :/

psifertex:
Scratch that -- I was conflating two different internal things we were discussing. Ignore me and just listen to @KyleMiles who will almost certainly have a better answer. 😛

KyleMiles (approved):
Yeah there are no problems with this.

fabianfreyer:
thanks!

PR #2961, opened 2022-02-11
rewrote the open_view function to be both a bit more rusty and nicer to read

Lukas-Dresel:
When debugging why the `basic_script` example does not build for me, I ended up checking out the open_view function. 

After scratching my head for a while, I decided to rewrite parts of it in what I think is both a bit more "rust-y" and also more readable. Is this something you guys want to do? I have an easier time overlooking what happens in the new implementation, if you guys disagree feel free to ignore me :D.


KyleMiles (changes requested):
These changes are _fantastic_, thank you!  Please run `cargo fmt`, force push, and I'll merge.

KyleMiles:
I went ahead did that myself. This is merged in `ae7ae4f63f931342e155d77bb4cf6d75289682bf`

PR #2976, opened 2022-02-18
[WIP] Rust: Better errors

Lukas-Dresel:
I made an initial attempt at better error messages as discussed in the slack channel. I've implemented a custom BNError type we can use to designate various different forms of errors while keeping track of dependent Errors (causes) as well as the backtrace of each error. 

This does unfortunately currently rely on the unstable [backtrace](https://github.com/rust-lang/rust/issues/53487) feature: 
`#![feature(backtrace)]`

For now, I only modified the code in `binaryview.rs` to use them, I'm happy to modify the rest after some review if this is something that will be incorporated :).

### DIFF
--- a/rust/src/backgroundtask.rs
+++ b/rust/src/backgroundtask.rs
@@ -43,7 +43,7 @@ impl BackgroundTask {
             return Err(
                 bn_api_error!(
                     BNBeginBackgroundTask,
-                    &format!("initial_text={:?}, can_cancel={:?}", bytes_error_repr(text.as_ref()), can_cancel)
+                    &format!("initial_text={:?}, can_cancel={:?}", Utf8Display(&text), can_cancel)
                 )
             );
         }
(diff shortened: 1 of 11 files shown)


Lukas-Dresel:
I did in the meantime push the code for all uses of `Err` in the API if i'm not mistaken, so it should be fully testable with all API logic. Unfortunately I only have a personal license, so I can't test any headless functionality (other than that I actually get an error now when trying to `open_view` in headless :P

Lukas-Dresel:
On Slack, people noted that they disliked the naming of `BNError` and `BNResult`, I am more than happy to change this, it simply persisted while I was trying to `grep` for old uses of `Result` that I had missed.

KyleMiles:
This is still something we want, and need to work towards resolving.  No one is currently tasked on working on this, however, so we're closing this due to inactivity and will continue to track the need for this in an issue (#3245).  If there's more work you want to do on this PR, feel free to reopen it.

PR #3054, opened 2022-04-02
Finish a BackgroundTaskThread on exception

### LINKED ISSUE #1431: Progress text of BackgroundTaskThreads that raise exception persists
If a BackgroundTaskThread raises an exception, the progress text in the status bar sticks around. Additionally, it does not look like any other BackgroundTaskThread can change the progress text afterwards.

fabianfreyer:
Fixes #1431


plafosse:
I think this would be better 🤔 

```python
	try:
		self.task.run()
	finally:
		self.task.finish()
		self.task = None
```

fabianfreyer:
@plafosse Sure! Fixed :)

PR #3099, opened 2022-04-29
Missing closing bracket in the first page of the documentation.

Stolas:
It's small, but it has been driving me crazy.


psifertex:
Thanks!

PR #3135, opened 2022-05-19
Open Source document fixups

alecnunn:
Hello, I was going through the docs, namely the Open Source listing and couldn't help but notice that there were a few items that appeared to have broken Markdown rendering. I went to check it out and it seems to have been caused by missing anchor links, which I have fixed up.


psifertex:
Thanks, much appreciated!

PR #3212, opened 2022-06-20
Implement .is_thunk property for HLILFunction.

WhatTheFuzz:
## Problem

There is not currently a way for a developer to query whether or not a HighLevelILFunction is a thunk from within the object itself. This is especially useful when writing snippets that operate on all functions within a BinaryView, but where thunks are not needed.

## Solution

Implement the `is_thunk` property that looks at the first two instructions of a HighLevelILFunction. If one of those instructions is a Tailcall, we can confidently say the function is a thunk. I have not observed a case where this does not hold for HighLevelILFunctions but additional testing may be necessary.

### DIFF
--- a/python/highlevelil.py
+++ b/python/highlevelil.py
@@ -2525,6 +2525,18 @@ class HighLevelILFunction:
 		core.BNFreeILInstructionList(uses)
 		return result
 
+	@property
+	def is_thunk(self) -> bool:
+		"""Return True if the function is a thunk, else False. (read-only)"""
+		instructions = self.instructions
+		try:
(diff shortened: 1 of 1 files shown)


fuzyll:
Thanks for the pull request!

This is a reasonable solution to the problem, however, we believe it should be done at a more fundamental level so that it's available to more than just HLIL. If this were done on the `Function` object _itself_, using LLIL, then I think we would accept it.

WhatTheFuzz:
I agree, that's probably more useful. I'll get on it and see what I observe with thunks in LLIL. Thanks!

PR #3243, opened 2022-07-06
Removed -v; now autodetects an active venv

kr1tzy:
Removed the virtualenv flag (-v) and added a prefix/base_prefix comparison to autodetect an active virtualenv.

### DIFF
--- a/scripts/install_api.py
+++ b/scripts/install_api.py
@@ -25,8 +25,9 @@ INTERACTIVE = True
 if '-s' in sys.argv[1:]:
 	INTERACTIVE = False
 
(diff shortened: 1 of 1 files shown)


psifertex (approved):
Thanks for the idea! Should also probably remove https://github.com/Vector35/binaryninja-api/pull/3243/files#diff-d722ba972d0554b4308d27fa0e7b46af7339cebce59efc78b245dee466d1570cR33-R35 but that will trigger the contributor agreement form request as too big of a change to make without it so I'll simply do it after accepting this. 😉

PR #3278, opened 2022-07-14
typo fix: venv detection requires prefix != base_prefix

kr1tzy:
my bad! 🙈 @psifertex


plafosse:
This one is actually a duplicate of the previous PR. Thanks any way :)

PR #3282, opened 2022-07-16
Fix bninspect on documented properties

### LINKED ISSUE #3246: `?` at the end of a python command doesn't work for properties
- Binary Ninja Version: 3.1-dev

**Bug Description:**
type `bv.functions?` in the python console. This will spit out an error instead of the docstring.

cxnder:
This fixes #3246 . 

How: 
* check whether the value we're trying to get docstrings from is a problematic type (non-class (not non-class-instance, but non-class-object) or non-method)
* Call `type(blah)` on everything before the property, then re-append the `.property` and re-eval. 
* This will work even on method types, but we try to be safe and fall back to the original behavior where we know it worked.

image

Here:
`bv.length` is an undocumented `@property` and returns a primitive. It will print `int.__doc__` as it did before
`bv.functions` is a documented `@property` that returns a Class. It will print the docstring for the property function
`bv.types` is an undocumented `@property` that returns a documented Class. It will print the class docs
`bv.mlil_functions` is a documented function that already worked; It falls back to the original behavior as intended.
`current_function.start` is a documented `@property` that returns a primitive. It will print the property function docs.

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -6150,7 +6150,7 @@ class BinaryView:
 		``get_linear_disassembly`` gets an iterator for all lines in the linear disassembly of the view for the given
 		disassembly settings.
 
-		.. note:: linear_disassembly doesn't just return disassembly it will return a single line from the linear view,\
+		.. note:: linear_disassembly doesn't just return disassembly; it will return a single line from the linear view,\
 		 and thus will contain both data views, and disassembly.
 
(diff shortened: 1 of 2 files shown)


psifertex:
https://docs.python.org/3/library/inspect.html#inspect.ismethod is a better way to check if something is a method

psifertex:
lgtm

PR #3343, opened 2022-08-02
Implement .is_thunk in LLILFunction.

WhatTheFuzz:
## Problem

There is not currently a way to check whether or not a function is a thunk.

## Solution

I think, ideally, this would be exposed from the core. However, I think think it is valid to check whether or not a Tailcall occurs within X amount of LLIL expressions in the function. From testing with different architectures, I found this number to be five LLIL expressions. I implemented this in both the LLILFunction and Function classes. 

## Related

See PR #3212

### DIFF
--- a/python/function.py
+++ b/python/function.py
@@ -567,6 +567,14 @@ class Function:
 		return BasicBlockList(self)
 
 	@property
+	def is_thunk(self) -> bool:
+		"""Returns True if the function starts with a Tailcall (read-only)"""
+		if self.llil_if_available is not None:
+			return self.llil_if_available.is_thunk
+		else:
+			return False
+
+	@property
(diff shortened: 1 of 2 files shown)


psifertex:
Thanks very much for re-working this. We actually came up with a slightly different implementation that doesn't depend on the size of the block. The other concern is that this could false-positive on a function that happened to have a basic block that tail called but other logic elsewhere that did a lot more (and thus wasn't a thunk). 

So in the end used the same comments and wrapper, thanks for that, but [changed the detection](https://github.com/Vector35/binaryninja-api/commit/f28fb80b7b49226006bbb1696c3c2bec07df891e).

Will be out on the next dev build.

WhatTheFuzz:
Looks great! I will definitely use this in my workflow. Thanks!

PR #3439, opened 2022-08-27
Remove `Segment::SetStart` which is not backed by C API

galenbwill:


plafosse (approved):
This should be safe to remove

PR #3489, opened 2022-09-30
Avoid referring to `std::vector<T>` members when `T` is incomplete.

cblichmann:
This is [not legal](https://timsong-cpp.github.io/cppwp/n4868/vector#overview-4) according to the C++ standard, and causes build errors in particular in C++20 mode. Fix it by defining the vector's type before using the vector.

This is a particular problem for C++20 because the compiler now generates code for `std::vector` members earlier, due to changes to which functions are `constexpr` in C++20. So code that was already illegal is now detected as such by the compiler, causing errors in the vector header about the use of incomplete types.

This was found during an internal #Cleanup at Google. No functional changes.

### DIFF
--- a/binaryninjaapi.h
+++ b/binaryninjaapi.h
@@ -1237,7 +1237,44 @@ namespace BinaryNinja {
 		std::string GetId();
 	};
 
-	struct InstructionTextToken;
+	struct InstructionTextToken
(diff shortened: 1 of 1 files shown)


plafosse:
Thanks

PR #3570, opened 2022-10-20
Include C++ standard headers

cblichmann:
When compiling in C++ mode, the C++ standard library headers should be included instead of their C counterparts.

Drive-by:
- Removed include of `json/json.h` from `settings.cpp`, where it was not used


plafosse:
Thanks for the PR 👍

plafosse:
This was merged I'm not sure why it didn't show up here though 🤔

PR #3571, opened 2022-10-20
Introduce a macro to implement `printf` attribute

cblichmann:
This lets us remove a lot of `#ifdef __GNUC__` in function declarations, which makes them more readable.


plafosse:
Thanks for the PR this was merged manually as there were some *fun* internal changes that I had to make to be able to pull in this code.

Issue #3456, opened 2022-09-02
Triage View: Import/export addresses should be green


CouleeApps:
**What is the feature you'd like to have?**
Make the import/export/entrypoint/base addresses all green, like everywhere else an address is shown in binja.

**Is your feature request related to a problem?**
UI inconsistency is always a problem :)

**Are any alternative solutions acceptable?**
Not really, it's just a standard thing

xusheng6:
Added in 6afd790abeccc98e4319e392ed1f8d8dd78baa8f

PR #3619, opened 2022-10-31
set address color to AddressColor for import and export tree in triage view

op2786:
Tried to solve #3456


op2786:
Am I need to do this painting trough delegate? If so, I can change the code.

xusheng6:
Merged into 6afd790abeccc98e4319e392ed1f8d8dd78baa8f

Issue #1682, opened 2020-05-18
triage view to show strings


psifertex:
It would be nice if the triage view also showed strings.

xusheng6:
Added in b4b6d3f5ca9751df27287684fa1fc21170e62d59

PR #3622, opened 2022-10-31
added string list for triage view

op2786:
Added strings into triage view, which requested in #1682


xusheng6:
Added in b4b6d3f5ca9751df27287684fa1fc21170e62d59

PR #3701, opened 2022-11-28
Bring BinaryReader.seek more in line with File

fabianfreyer:
Adds a `whence` parameter to BinaryReader.seek which behaves similarly to FileObjects seek method, but defaults to 0, keeping backwards compatibility.

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -8232,11 +8232,12 @@ class BinaryReader:
 			return None
 		return struct.unpack(">Q", result)[0]
 
-	def seek(self, offset: int) -> None:
+	def seek(self, offset: int, whence: int = 0) -> None:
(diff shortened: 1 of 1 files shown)


plafosse (approved):

fabianfreyer:
To allow whence to seek_end, is there any way to get the length?

KyleMiles:
The binary view has length information, but that doesn't seem to be passed through to or saved by the BinaryReader. Maybe it'd be better to expose that, but for now just saving the length from the binary view might be easier.

This PR looks good to me. Just waiting for @fabianfreyer to sign the CLA. I seem to remember you having signed that before, so I'm not sure if our CLA changed, CLAassistant is wrong, or what.

PR #3724, opened 2022-12-08
Change type from array->string.

WhatTheFuzz:
`enum` seems to be the list of strings that are possible values for the `string` type. When the type is set to `array` and `enum` is present, Settings fails to load. Additionally, I could not find any examples of Settings where the type is set `array` and `enum` is added.

I confirmed that the list of strings defined in `enum` populate a drop down list when the `type` is set to `string`. For reference, check out the example at [openai-binaryninja](https://github.com/WhatTheFuzz/binaryninja-openai/blob/main/src/settings.py#L45-L55).

### DIFF
--- a/python/settings.py
+++ b/python/settings.py
@@ -65,8 +65,8 @@ class Settings:
 		"title"              string                                   None                 No         Concise Setting Title
 		"type"               string                                   None                 No         {"array", "boolean", "number", "string"}
 		"elementType"        string                                   "type" is "array"    No         {"string"}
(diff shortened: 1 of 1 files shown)


bpotchik (commented):
The "type" property being set to `array` is a prerequisite for the "enum" and "enumDescriptions" properties to be valid and used by the settings system. The current documentation is correct.

WhatTheFuzz:
No problem, it really confused me when I tried to implement the drop-down menu. Hopefully this saves someone some time! 🙂

PR #3729, opened 2022-12-12
Bump Rust edition to 2021

jrozner:
There's a bunch of quality of life changes and cleanup that can happen after this goes in. I can get those in with this if desired but they aren't necessary and it will touch a LOT of stuff.


KyleMiles:
I'm going to update cargo at the same time as merging this... Which will require me updating the build servers and CMakeLists first. Not sure if that's required to merge this PR or not.

jrozner:
Updating Cargo shouldn't be required unless you're running a version so old it doesn't support edition 2021 but that would be REALLY old.

PR #3770, opened 2023-01-06
General cleanup of rust bindings

mkrasnitski:
This PR performs a sweep across the rust bindings to clean up cruft and make things more idiomatic. Broken up into commits as follows:
1. https://github.com/Vector35/binaryninja-api/pull/3770/commits/2b39acba0ca47eda39490079485f4cec2051c16d - Clean up `open_view` and `open_view_with_options` to reduce code duplication.
2. https://github.com/Vector35/binaryninja-api/pull/3770/commits/1ad8236629de917be5fe60232b6fd21465e290b7 - Move the `logger` module out of `lib.rs` and into its own file.
3. https://github.com/Vector35/binaryninja-api/pull/3770/commits/c28925981920ea15dc9713993fd59905f11900dd - Clean up various builders. Introduces the following breaking changes:
    - Rename `{Section,Segment,Symbol}::new` to `builder`.
    - Similarly, remove `{Enumeration,Structure}::new` and add a `builder` method. No such method was added to `Type`, as it doesn't seem possible to initialize a `TypeBuilder` without an existing handle (which is what `Type::to_builder` does).
    - Rename `{Section,Segment,Symbol}Builder::create` to `build`.
    - Remove the implementation of `From<*Builder> for Ref<*>` for `Type`, `Enumeration`, and `Structure` that wrapped `Builder::finalize`. Users should call the method directly. (Should it also be renamed to `build`?)
    - Replace `impl Into<Expression> for ExpressionBuilder` with `ExpressionBuilder::build`.
    - Mark `Enumeration::ref_from_raw` as an unsafe fn.
 4. https://github.com/Vector35/binaryninja-api/pull/3770/commits/df8cede5fce9c9e9eda19282a26e63f7b4f096ed - Add `Expression::new` to reduce code duplication.
 5. https://github.com/Vector35/binaryninja-api/pull/3770/commits/b9141e618594beaf87c8740b7e52582ef1254115 - Remove any unused `impl AsRef<T> for T`. This is a breaking change, but not likely to cause problems.


rust/src/segment.rs
```diff
@@ -88,7 +88,7 @@ impl SegmentBuilder {
         self
     }
 
-    pub(crate) fn create(self, view: &BinaryView) {
+    pub(crate) fn build(self, view: &BinaryView) {
```
KyleMiles, line 91: What's the reasoning behind this change?

mkrasnitski: It's simply the canonical method name for builders. Seems like right now the method name matches the underlying FFI call. This is also true for `{TypeBuilder, EnumerationBuilder, StructureBuilder}::finalize()`, but that method call is public so I was unsure whether or not break it.

KyleMiles: Are you saying that the Rust convention for builders to have terminals named `build`?  I can't seem to find that documented with the rest of the [builder conventions](https://rust-lang.github.io/api-guidelines/type-safety.html?highlight=build#builders-enable-construction-of-complex-values-c-builder).

If there's not a solid Rustism behind the change, I'd rather leave it as it is for people porting from Python / our existing devs.

mkrasnitski: I think my notion of that came from the [Rust Design Patterns](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) book, which I will grant, is unofficial. The example in the official guidelines you linked uses a different verb that better suits the example use-case, so I guess it's not a super concrete -ism. I'll revert the change, squash, and force push.

KyleMiles (approved):

KyleMiles:
Thank you for the PR! I really appreciate all this!

PR #4265, opened 2023-04-26
Added un/define_user_data_var functions for Rust

mmaekr:


KyleMiles (changes requested):
This looks good to me, but would mind adding a doc string commenting that the user of `define_user_data_var` also likely wants to be calling `define_user_symbol` to bind a symbol to that variable please?

KyleMiles:
Actually, this is so minor I'll do it myself. Thanks again for the PR!

PR #4279, opened 2023-05-04
Improved open_view for bndb with invalid view

tanakalian:
calling open_view on a bndb with an invalid view would panic as unwrap was called. Fixed by matching to handle the error case.


KyleMiles (approved):
Looks good to me. I'll merge this after I check it locally.

PR #4284, opened 2023-05-05
[Rust] Stack Layout, Code Reference, Data Reference APIs

toolCHAINZ:


psifertex:
This was manually verified for CLA. Not sure why it's still not showing up as approved. Probably because submodules and I have to add to each one?

KyleMiles (approved):
Thank you for the huge PR! This is great, it all looks good to me. I verified that your handling of `QualifiedName`s is fine...they're refcounted internally so at the worst the core will handle any internal references it takes.

PR #4378, opened 2023-06-01
Add Var, SSAVar, and AliasedVar to commonil.py

KyleMiles:


meithecatte:
Note that this conflicts with `mediumlevelil.SSAVariable` and `variable.Variable`.

PR #4423, opened 2023-06-19
Ensure that binaryninjacore.h is valid C.

### LINKED ISSUE #4422: `binaryninjacore.h` is not valid C, despite seemingly trying to be
The header has some `__cplusplus` checks, so it seems like it's trying to be valid C. But it actually isn't:

```text
... snip ...
../../api\binaryninjacore.h(1026,3): error : must use 'enum' tag to refer to type 'BNRegisterValueType'
../../api\binaryninjacore.h(1064,3): error : must use 'struct' tag to refer to type 'BNType'
... snip ...
```

So what's the deal there? Is it actually meant to be consumable from a C project or only C++?

(For context, it's a lot easier to cross-compile C code than C++ code, so for my plugin, I would much prefer to compile it as C.)

alexrp:
Fixes #4422.

### DIFF
--- a/binaryninjacore.h
+++ b/binaryninjacore.h
@@ -27,6 +27,7 @@
 #include <cstddef>
 #include <cstdlib>
 #else
+#include <stdbool.h>
 #include <stdint.h>
 #include <stddef.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


binaryninjacore.h
```diff
@@ -27,6 +27,7 @@
 #include <cstddef>
 #include <cstdlib>
 #else
+#include <stdbool.h>
```
alexrp, line 30: This being C11, I'm hoping it's common enough by now...

binaryninjacore.h
```diff
@@ -137,11 +138,16 @@
 #ifdef DEMO_VERSION
 	#define BN_DECLARE_CORE_ABI_VERSION
 #else
-	#define BN_DECLARE_CORE_ABI_VERSION \
-		extern "C" \
-		{ \
-			BINARYNINJAPLUGIN uint32_t CorePluginABIVersion() { return BN_CURRENT_CORE_ABI_VERSION; } \
-		}
+	#ifdef __cplusplus
```
alexrp, line 141: This is necessary because C has no idea what `extern "C"` means.

binaryninjacore.h
```diff
@@ -163,127 +169,127 @@
...
+	typedef enum BNPluginLoadOrder
 	{
 		EarlyPluginLoadOrder,
 		NormalPluginLoadOrder,
 		LatePluginLoadOrder
-	};
+	} BNPluginLoadOrder;
 
-	enum PluginLoadStatus
+	typedef enum PluginLoadStatus
```
alexrp, line 179: This one stood out to me for lacking the `BN` prefix. Is that something worth fixing?

CouleeApps:
Alright, I've poked this into the CI. Assuming it passes I'll merge this.

binaryninjacore.h
CouleeApps: Quick check of the source shows this is ... not actually used in any APIs and only exists here to have parity between the UI plugin loader and the core plugin loader. I recommend Just Not Touching It And Things Will Be Fine.

CouleeApps:
Looks like it passed. Thanks for the contribution!

PR #4476, opened 2023-07-11
[Rust] Nightly fix, QoL for clangd, add BinaryView function

unknowntrojan:
This PR is relatively small but I'm looking to spend some time in the future getting the Rust API more on par with the C++ API, so expect more to come. It changes 3 things that are relatively unrelated, but opening 3 seperate PRs seemed counterproductive.

This PR does the following:
- add the clangd compilation cache to .gitignore
- bump the bindgen version for binaryninjaapi-sys from ^0.62 to ^0.66, which uses a new version of syn, which uses a new version of proc-macro2, which used a feature flag that is now merged with another. (proc_macro_span & proc_macro_span_shrink). This fixes the build for the current nightly builds of rustc. There is no change for stable builds.
- add BinaryView::functions_containing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ python/generator.*.log
 python/generator.exe
 python/generator.log
 python/generator.tlog
+.cache/
 
 # Unit Tests
 suite/unit.py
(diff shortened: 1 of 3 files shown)


KyleMiles (approved):
Looks good to me. I just want to pull this and compile before merging, just to make sure there are no other artifacts of updating that dependency.

For your future PRs, I prefer wrapping the smallest possible amount of code in unsafe blocks... eg: `let thing = unsafe {BNCoreFunctionCall};`, but that's really just a style note more than anything.

Thank you for the PR!

KyleMiles:
We have some lock files in the core I needed to commit per this update was all.

PR #4496, opened 2023-07-18
rust: Add support for defining auto data vars and auto types

mkrasnitski:
In particular, added the following methods to `BinaryViewExt`:
  * `define_auto_data_var`
  * `undefine_auto_data_var`
  * `define_auto_type`
  * `undefine_auto_type`
  * `undefine_user_type`

I'd like feedback on the names - should they include `auto` in order to maintain consistency with `define_auto_symbol`, or should they instead reflect the underlying FFI calls and not include them? I like the dichotomy of user/auto more than user/nothing - it seems clearer

Additionally, I decided to have `define_auto_type` take a `source` parameter that it uses to auto-generate the id, rather than taking a pre-generated `id`, to improve usability. Users can get the id back by calling `BinaryViewExt::get_type_id` if they need to undefine the type. Is this ok to do?

### DIFF
--- a/rust/src/binaryview.rs
+++ b/rust/src/binaryview.rs
@@ -476,10 +476,22 @@ pub trait BinaryViewExt: BinaryViewBase {
         }
     }
 
+    fn define_auto_data_var(&self, dv: DataVariable) {
+        unsafe {
+            BNDefineDataVariable(self.as_ref().handle, dv.address, &mut dv.t.into());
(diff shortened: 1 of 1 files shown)


KyleMiles:
Yeah, hiding the ID is a good call. I also prefer the explicit auto/user.

KyleMiles (approved):

PR #4512, opened 2023-07-24
rust: Change `StructureBuilder::insert` to take and return `&self` instead of `&mut self`

mkrasnitski:
Since all other methods on `StructureBuilder` take and return `&self`, it is impossible to chain a call to `insert` or `insert_member` in the middle of a sequence of method calls when using the builder pattern on the type, since they require mutable access which cannot be granted by the other methods.

### DIFF
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -1744,11 +1744,7 @@ impl StructureBuilder {
         self
     }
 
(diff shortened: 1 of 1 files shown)


KyleMiles (approved):
Yup, thank you. Must've missed that in my cleanup way back when.

PR #4513, opened 2023-07-24
rust: Add `StructureBuilder::with_members` method

mkrasnitski:
To allow for better inline creation of structures with many fields, this method takes a sequence of tuples `(type, name)` and appends them to the structure all at once.

Also, added an impl of `From<&'a Ref<T>>` for `Conf<&'a T>` so that converting between the two can be done implicitly without having to call `Ref::as_ref` or to reborrow using `&*`. Unfortunately, an owned impl that converts directly from `Ref<T>` doesn't seem to be possible to write (I tried for quite a long time and the borrow checker kept screaming at me).

### DIFF
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -1786,6 +1786,16 @@ impl StructureBuilder {
         self
     }
 
+    pub fn with_members<'a, S: BnStrCompatible, T: Into<Conf<&'a Type>>>(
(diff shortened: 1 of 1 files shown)


KyleMiles (approved):
I was running into something similar with regards to the impl here recently, thanks for taking a crack at it!

PR #4548, opened 2023-08-05
Add API to get all SSA versions of a variable

emesare:
Added a missing helper for both `MediumLevelILFunction` and `HighLevelILFunction` classes to get a list of all SSA versions of a variable.

### DIFF
--- a/binaryninjaapi.h
+++ b/binaryninjaapi.h
@@ -11850,6 +11850,7 @@ namespace BinaryNinja {
 		std::set<size_t> GetSSAMemoryUses(size_t version) const;
 		bool IsSSAVarLive(const SSAVariable& var) const;
 
+        std::set<size_t> GetVariableSSAVersions(const Variable& var) const;
 		std::set<size_t> GetVariableDefinitions(const Variable& var) const;
 		std::set<size_t> GetVariableUses(const Variable& var) const;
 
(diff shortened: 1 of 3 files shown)


xusheng6 (changes requested):
Nice catch and thx for the contribution!

There is one little change needed before I can merge this: The versions returned by the core API are allocated in the core (https://github.com/Vector35/binaryninja-api/pull/4548/files#diff-8ac95093418e4830504276ed52f586164037fd3bd25b24634199cda30dd7f18fR347), and the current implementation is causing memory leak because the API side does not free the array. You need to add a call to free it as we do here: https://github.com/Vector35/binaryninja-api/blob/7598688466960427890036590239565364310171/function.cpp#L1412. (You can ignore the name `BNFreeILInstructionList`, it essentially just frees an size_t array via its pointer)

If you are not sure about what I mean, you can let me know and I can take over the job and commit it directly.

emesare:
> . (You can ignore the name `BNFreeILInstructionList`, it essentially just frees an size_t array via its pointer)

I was unsure about if that was only for instruction lists, thank you for the clarification!

xusheng6:
Merged into b6ce48c4f5be4cd43482e1ff762e088fd11b7227

PR #4631, opened 2023-09-17
Create ExceptionWithStackTrace when compiling with Core.

reciate:
Had an issue compiling BinExport due to it compiling with BINARYNINJACORE_LIBRARY defined. This creates a second struct in exceptions.h since it seems like the plan was to have a separate core implementation.

Related Pull Request: https://github.com/google/binexport/pull/115

### DIFF
--- a/exceptions.h
+++ b/exceptions.h
@@ -39,10 +39,26 @@ namespace BinaryNinja
 		}
 	};
 }
+#else
+namespace BinaryNinja
+{
+	struct ExceptionWithStackTrace : std::exception
+	{
+		std::string m_originalMessage;
+		std::string m_message;
+		std::string m_stackTrace;
+		ExceptionWithStackTrace(const std::string& message);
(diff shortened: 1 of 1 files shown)


CouleeApps:
Why are you compiling with `BINARYNINJACORE_LIBRARY`? That is the define specifically set by the binaryninjacore (private) library, and is not expected to be defined by any api/consumers of the api.

reciate:
I personally am not sure as to why. BinExport's stub regeneration script defines it [here](https://github.com/google/binexport/blob/main/third_party/binaryninja_api/regenerate-api-stubs.sh#L70) to produce this [stub file](https://github.com/google/binexport/blob/main/third_party/binaryninja_api/binaryninjacore_stable.cc#L21).

CouleeApps:
With binexport commit https://github.com/google/binexport/commit/f908bebbf616dab9cf5d88b07ce062288a66e147 this patch may no longer be necessary?. Either way, this is not a change set I would be willing to merge, so I'm closing this PR.

PR #4676, opened 2023-10-10
Add visit_simple to LLIL, MLIL, and HLIL

KyleMiles:
This has been coming up in the last couple training classes....

The normal visitor API is hard to use and has some information in it most people don't want/need. This attempts to provide a easier interface by only requiring a single parameter, and allows the visitor to be more versatile by allowing user defined parameters and return values.

This enables extracting more complex information without needing to use a global variable.

I think this would be a good replacement for the existing visitors, but properly deprecating them would be a bit messier given all the type signature changes.

### DIFF
--- a/python/highlevelil.py
+++ b/python/highlevelil.py
@@ -20,7 +20,7 @@
 
 import ctypes
 import struct
-from typing import Optional, Generator, List, Union, NewType, Tuple, ClassVar, Mapping, Set, Callable
+from typing import Optional, Generator, List, Union, NewType, Tuple, ClassVar, Mapping, Set, Callable, Any
 from dataclasses import dataclass
 from enum import Enum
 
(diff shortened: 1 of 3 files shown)


KyleMiles:
In interest of getting this merged, I rebased this branch and hopefully quelled the discussion about what to name the new function by properly deprecating the old one and auto-dispatching to the new one.

Limitations of this implementation include:
 - You can't have a new visitor with exactly four parameters (any introspection that's more in-depth I believe relies on type annotations, which we can't assume are there)
 - You can't use the names `parent` nor `name` as a parameter in new visitors.

I can probably fix that second limitation if you like this new approach, please let me know!

plafosse (changes requested):
Instead of doing it like this I'd prefer a distinct api called `traverse`

KyleMiles:
Changes were made. Should be good to merge now. Thank you!

PR #4678, opened 2023-10-11
Resolves #4144; Add ergonomic way of retrieving string from ConstPtr …

### LINKED ISSUE #4144: More ergonomic way of retrieving an IL string operand
I'd like to have an ergonomic API for retrieving the string referenced by the operand of an IL operation. This _may_ be as easy as just fixing the output of `str()` on an operand, but I'm not entirely sure as I'm not familiar with the implementation of these APIs.

Given line 13 in MLIL in this screenshot:

image

...I want to get the string `'MS Word document scan in progress...\n'` out. The _only_ ways that I have been successful have been by doing:

```
bv.get_ascii_string_at(current_il_instruction.params[1].value.value).value
# or
bv.get_ascii_string_at(current_mlil[13].params[1].value.value).value
```
...with it selected (so that `current_il_instruction` refers to that one). If there's a better way right now, please feel free to close a pointer to the right API. But, other possible solutions I came up with like this one:

```
current_il_instruction.operands[2][1].__str__()
```

...don't actually get me the full string. They're truncated and return `'"MS Word document scan in progres…"'` instead.

As mentioned above, if `str()` just worked on the operand, I think that would be fine? Otherwise, modifying `get_string_at()` and friends to be able to unpack the value I want would be my suggestion.

I would also accept just having explicit documentation on this as a way to close the issue. I feel like it's a common enough operation that we should at least try to make it more discoverable.

**Additional Information:**
This came up while trying to help another user with this problem tonight. Their issue was different and involved something that had been outlined, but the solution was similar.

KyleMiles:
…IL instructions

### DIFF
--- a/python/highlevelil.py
+++ b/python/highlevelil.py
@@ -26,7 +26,7 @@ from enum import Enum
 
 # Binary Ninja components
 from . import _binaryninjacore as core
-from .enums import HighLevelILOperation, DataFlowQueryOption, FunctionGraphType, ILInstructionAttribute
+from .enums import HighLevelILOperation, DataFlowQueryOption, FunctionGraphType, ILInstructionAttribute, StringType
 from . import function
 from . import binaryview
 from . import architecture
@@ -1690,6 +1690,10 @@ class HighLevelILConstPtr(HighLevelILInstruction, Constant):
 		return [
 			("constant", self.constant, "int"),
 		]
+
+	@property
+	def string(self) -> Optional[Tuple[str, StringType]]:
+		return self.function.view.check_for_string_annotation_type(self.constant, True, True, 0)
 
 
 @dataclass(frozen=True, repr=False, eq=False)
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -43,8 +43,7 @@ from . import _binaryninjacore as core
 from . import decorators
 from .enums import (
     AnalysisState, SymbolType, Endianness, ModificationStatus, StringType, SegmentFlag, SectionSemantics, FindFlag,
-    TypeClass, BinaryViewEventType, FunctionGraphType, TagReferenceType, TagTypeType, RegisterValueType, LogLevel,
-	DisassemblyOption
+    TypeClass, BinaryViewEventType, FunctionGraphType, TagReferenceType, TagTypeType, RegisterValueType, DisassemblyOption
 )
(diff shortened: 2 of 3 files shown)


python/highlevelil.py
```diff
@@ -1646,6 +1646,10 @@ def detailed_operands(self) -> List[Tuple[str, HighLevelILOperandType, str]]:
 			("constant", self.constant, "int"),
 		]
 
+	@property
+	def string(self) -> Optional[binaryview.StringReference]:
```
fuzyll, line 1650: I rebased this branch on `dev` and I have no Python Console.

After some debugging (and setting up a completely new, fresh Python just for a debug build), doing an `import binaryninja` with the built API from this branch results in a big traceback because:

```
    class HighLevelILConstPtr(HighLevelILInstruction, Constant):
  File "/Users/fuzyll/Builds/binja-dbg/out/binaryninja.app/Contents/Resources/python/binaryninja/highlevelil.py", line 1650, in HighLevelILConstPtr
    def string(self) -> Optional[binaryview.StringReference]:
AttributeError: partially initialized module 'binaryninja.binaryview' has no attribute 'StringReference' (most likely due to a circular import)
```

Does this actually work? Or, am I Doing It Wrong..?

fuzyll (changes requested):
If I am understanding this correctly, after this PR we would go from:

```
bv.get_ascii_string_at(current_il_instruction.params[1].value.value).value
```

...to:

```
current_il_instruction.params[1].string.value
```

Is that accurate? If so, I think that's much more reasonable and I, at least, would be willing to accept this. But, as of right now, I can't actually get it to build and work. So, I'm going to need some help first.

KyleMiles:
Sorry about that @fuzyll! Should be fixed now: circular imports and Python being bad...needed to add quotes around the `'binaryview.StringReference'`s.

KyleMiles:
And yes, how this API should be used is correct in your comment above.

plafosse (changes requested):
This should be getting strings similar to the way the core does it. Which boils down to using `BNCheckForStringAnnotationType`  which also would need to have a python wrapper written for it.

KyleMiles:
Made the requested changes. On a quick test on `/bin/bash`, I'm getting over 2500 additional strings returned when using `BNCheckForStringAnnotationType` instead of doing it the old way. A good improvement!

PR #4698, opened 2023-10-21
add rust mlil api

rbran:
This PR try to implement a initial api for MLIL, it's based on the LLIL implementation.


rbran:
One open question that I have is the function [get_list](https://github.com/Vector35/binaryninja-api/blob/dc0e5fc53137c4e039794516cd27efd7149b4f2d/rust/src/mlil/operation.rs#L62).

The other APIs in [python](https://github.com/Vector35/binaryninja-api/blob/52eeacdafc4d5eff861951db2099f831fb76ff10/python/mediumlevelil.py#L926) and [cpp](https://github.com/Vector35/binaryninja-api/blob/52eeacdafc4d5eff861951db2099f831fb76ff10/mediumlevelil.cpp#L155) use the sys function `BNMediumLevelILGetOperandList`, that allocate memory and return the array with all the elements.

In my implementation I return a iterator that get values from the tree one-by-one.

Is this adequate or the function `BNMediumLevelILGetOperandList` is required?

KyleMiles:
First of all, thank you tremendously for all the work you put in to this PR. We would love to have MLIL and HLIL bindings in Rust, and it seems like you put a lot of effort in to this.

But before you put more work in to this, we should chat: https://slack.binary.ninja (my handle is Elyk on there). There's a dedicated rust-api channel.

This PR seems to be modeled off of the LLIL bindings, which were originally designed primarily for writing architecture plugins, rather than consumption by plugins and scripts. It needs to be refactored. This also diverges from another draft we proposed internally (https://github.com/Vector35/binaryninja-api/tree/test_mlil_bindings/rust/src/mlil), so I need to discuss with the team about which direction we want to go.

rust/src/mlil/operation.rs
```diff
@@ -0,0 +1,1151 @@
...
+    pub fn size(&self) -> usize {
+        self.op.size
+    }
+
+    fn get_expr(&self, operand_index: usize) -> Expression<'func, A, M, F, ValueExpr> {
+        Expression::new(self.function, self.op.operands[operand_index] as usize)
+    }
+
+    fn get_int(&self, operand_index: usize) -> u64 {
+        // TODO check mask...
```
rbran, line 49: The python API allow all values to be positive/negative, its unclear witch ones are signed/unsigned.

rust/src/mlil/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+mod operation;
+
+pub use self::block::*;
+pub use self::expression::*;
+pub use self::function::*;
+pub use self::info::*;
+pub use self::lift::*;
+pub use self::operation::*;
+
+pub use self::block::Block as MediumLevelBlock;
```
KyleMiles, line 20: Instead of aliasing them, just rename them where they're defined. It'll make maintaining this a little easier not having to remember the duality of the naming.

rust/src/mlil/expression.rs
KyleMiles, line 1: I feel like this intermediate layer shouldn't be needed....see later comments.

rust/src/mlil/info.rs
KyleMiles, line 1: What is this used for? Can it be removed?

rust/src/mlil/operation.rs
```diff
@@ -0,0 +1,1234 @@
+use binaryninjacore_sys::BNMediumLevelILInstruction;
+use binaryninjacore_sys::BNMediumLevelILOperation;
+
+use core::marker::PhantomData;
+
+use crate::{rc::Ref, types::SSAVariable};
+
+use super::*;
+
+pub struct Operation<O: OperationArguments> {
```
KyleMiles, line 10: If I'm reading this right:
 - Operations can return Expressions (Operation.left, operation.right, etc)
 - Expressions can return ExprLifted and ExprInfo
 - ExprLifed is the thing we like, it's what we can match on
 - ExprInfo....I can't tell what that's being used for

Can we remove `Expression`s and `ExperLifted` and just have operations return ExprInfo?
Also, if `Operation` objects are the "main thing you interact with," they should probably be renamed to `MediumLevelILInstruction`s

KyleMiles (changes requested):
This is a great improvement, thank you!

I hid my main feedback in operation.rs, I'm copy-pasting it here so it's not lost:

> If I'm reading this right:
>
>  - Operations can return Expressions (Operation.left, operation.right, etc)
>  - Expressions can return ExprLifted and ExprInfo
>  - ExprLifed is the thing we like, it's what we can match on
>  - ExprInfo....I can't tell what that's being used for
>
> Can we remove Expressions and ExperLifted and just have operations return ExprInfo?
> Also, if Operation objects are the "main thing you interact with," they should probably be renamed to MediumLevelILInstructions

rust/src/mlil/operation.rs
rbran: > ExprLifed is the thing we like, it's what we can match on
> Can we remove Expressions and ExperLifted and just have operations return ExprInfo?

Do you want to remove or keep ExprLifed?

Can be removed, but you mentioned on slack:

> I think what it boils down to is being able to write match cases like `If(condition: LessThan(right: Add(lhs, rhs)))`

The main reason that I implemented both is:

`ExprLifted` is used for matching and is easy to use. Like in https://github.com/Vector35/binaryninja-api/tree/test_mlil_bindings/rust/src/mlil.

`ExprInfo` is useful if you need a more elegant code, is the equivalent of LLIL ExprInfo, only calling the API if the data is required, not preemptively allocate Vectors (like in ExprLifted) and is useful for implementing a visitor pattern.

rbran: BTW ExprLifted is build using ExprInfo, It doesn't really make sense removing ExprInfo and keeping ExprLifted.

KyleMiles: Sorry, I misspoke there... I meant to say:
> Can we remove Expressions and ExprInfo and just have operations return ExprLifted?

In which case, can ExprLifted build itself? I don't actually see where ExperInfo is being used to make ExprLifted.

rbran:
Maybe I should better explain the order with the data is processed:

```
MLIL -> Block -> Operation<T> ---(group into categories)---> ExprInfo ---(collect all data into owned form)---> ExprLifted
```
Note that the API only exposes `ExprLifted` and `ExprInfo`. And deleting anything in the middle, just hides it from the user.

KyleMiles:
Okay, what I was missing was [this call](https://github.com/Vector35/binaryninja-api/pull/4698/files#diff-27582bafd795ad1fd005041308b988dd77b5d0ce882fb16f64077ae2be36ddedR29) to `self.info()`. It still seems like an unnecessary step, and I still don't see what `ExprInfo` is useful for.

If you could add examples to this PR showcasing how to use the MLIL bindings, including ExprInfo and ExprLifted, that might clarify things.

KyleMiles:
Copying the answer to my question from Slack:

> The [ExprInfo](https://github.com/rbran/binaryninja-api/blob/c99359a707cf3390f807156f9fdcd1913726cf9f/rust/src/mlil/info.rs#L7) was keep, it's similar to llil ExpInfo, except it's a just a single enum with regular/ssa variants. Can be used for implementing the visiting pattern.
> The [ExprLifted](https://github.com/rbran/binaryninja-api/blob/c99359a707cf3390f807156f9fdcd1913726cf9f/rust/src/mlil/lift.rs#L6) was created, it's just a owned version of ExprInfo, very useful for matching.
> The ExprLifted implementation allow to match similar operations (all binary/unary ops) but also specific operations.
> The equivalent of match If(condition: LessThan(right: Add(lhs, rhs))) is:
> 
> ```match &lifted {
>     ExprLifted::If { condition, .. } => match condition.as_ref() {
>         ExprLifted::BinaryOp {
>             op_type: BinaryOpType::CMP_ULT,
>             left,
>             right,
>         } => match (left.as_ref(), right.as_ref()) {
>             (
>                 ExprLifted::Var { src: left },
>                 ExprLifted::BinaryOp {
>                     op_type: BinaryOpType::ADD,
>                     left: add_left,
>                     right: add_right,
>                 },
>             ) => todo!("if ({left:?} < {add_left:?} + {add_right:?})`"),
>             _ => {}
>         },
>         _ => {}
>     },
>     _ => {}
> }```

KyleMiles:
Feel free to drop that commit if it conflicts with changes you're working on, figured I'd help out a little with the changes I request.

KyleMiles:
After pouring over this for some time, I think what I want to see before this can be merged is:

- [ ] Merge `Expression` and `Operation` into `MediumLevelILInstruction`
- [ ] Delete `ExprInfo`
- [ ] Expand `ExprLifted` to have variants for each BNMediumLevelILOperation, rename `ExperLifted` to be `MediumLevelILOperation`
- [ ] Delete all the `Operation<T>` variants.

Right now it feels like there are several ways to interact with MLIL, through the ExprInfo, ExprLifted, Operation, and Expression objects, which is too much. I'd like the main way to interact with MLIL to be through just one object. The preferred implementation for us is going to be through an enum-style object (`ExprLifted` but expanded for all the `BNMediumLevelILOperation`s), rather than all the implementations of `Operation` (and rather than all the Implementations of `Operation` wrapped in `ExprInfo` variants).

I'm still open to having all the member functions of `Operation` on the new MediumLevelILInstruction object, as functions that return options, if you want that, so you don't need to say `inst.operation.value` and instead `inst.value.unwrap()`.

We can still provide a visitor helper function like we do in the Python API so we don't lose visitability you mentioned that `ExprInfo` provides.

I'm thinking we might want to provide a macro to help with matching. The example you provided above is exactly what I want to do, but it ain't pretty (thanks, Rust). You don't need to write this, but I'm thinking we might provide one macro that does the internal match statement expansion automatically, and then perhaps another one that does the same but for "IL Groups"...so you can still match over "Arithmetic" instructions, or "Binary" operators. The second one might be a bit trickier, but should still be doable. I'm thinking it might look something like:

```rust
match inst.match_friendly_thing() {
  ilgroup!(Call {dest: dest}) => (),
  mlilinstruction!(If{condition: LessThan{right: Add{left: lhs, right: rhs}} }), // This would be great, but might need to be more like:
  mlilinstruction!(If{condition: mlilinstruction!(LessThan{right: mlilinstruction!(Add{left: lhs, right: rhs})})}), 
}
```

rbran:
Unfortunately I'm unable to continue this PR.

> After pouring over this for some time, I think what I want to see before this can be merged is:
>     * [ ]  Delete `ExprInfo`

The project that I'm working need an way to access mlil-nodes using a visitor-like pattern with minimum overhead, I can only do that using `ExprInfo`, and I'm not getting paid to develop something that I can't use.

Fell free to copy my code, unfortunately I'm unable to continue with those changes.

Thanks for you time.

KyleMiles:
Yeah, I totally understand that. Thank you for the time you put in to this PR! It's a great starting point for when we decide to get time to build out the API ourselves. I don't think that the changes I've requested would introduce significant computational overhead for the visitor pattern, but I was originally worried when I saw this PR that the changes needed to it would be nearly a complete rewrite.

Sorry we couldn't get this merged, but thank you so much for your contribution!

PR #4765, opened 2023-11-22
Show repr of BasicBlock in hex

### LINKED ISSUE #4708: repr of BasicBlock does not show start and end address in hex
```
>>> current_basic_block
<BasicBlock: x86_64@5368754260-5368754267>
```

SmoothHacker:
Fixes #4708. The hex conversion only applies to non-IL basic blocks. All IL basic blocks now have their own repr method that displays the start and end indices in decimal.

### DIFF
--- a/python/basicblock.py
+++ b/python/basicblock.py
@@ -68,9 +68,9 @@ class BasicBlock:
 	def __repr__(self):
 		arch = self.arch
 		if arch:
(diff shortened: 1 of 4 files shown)


xusheng6:
Hi @SmoothHacker, thx for the PR! I will work with you to get this PR merged in the following days!

xusheng6:
@SmoothHacker LGTM and I have merged the PR!

The only change I made is to use `{:#x}` to display a value as hex, rather then using `hex()`

PR #4769, opened 2023-11-22
add hlil to rust

rbran:
This is a draft for the HLIL rust implementation based on MLIL implementation.

TODO list:
- [x] implement ConstantData data type
- [x] implement Intrinsic data type
- [x] allow the user to set (or not) the ast-full option
- [x] implement the Label type, used by Goto and Label operations
- [x] handle cross reference pointers, eg two jmp instructions one point to the other
- [x] create examples for the hlil usage
- [x] implement function Hash that is not be based on pointers.
- [x] merge hlil Instruction and Operation
- [x] merge mlil Instruction and Operation
- [x] code cleaning


rbran:
@KyleMiles I didn't understand this line: https://github.com/Vector35/binaryninja-api/blob/66fe34a49322d0a63d8f6efe00ca20d9a2c4ce05/python/highlevelil.py#L2070

The operand 2 is used twice?
```
intrinsic: 0
params: 1, 2
dest_memory: 2
src_memory: 3
```

The correct operands should be `0, 1, 2, 3,4`?:
```
intrinsic: 0
params: 1, 2
dest_memory: 3
src_memory: 4
```

KyleMiles:
Yes, that's correct. That's a typo in the HLIL bindings. Thanks for noticing that, I'll fix it.

fuzyll:
I believe the corresponding C++ API implementation is here: https://github.com/Vector35/binaryninja-api/blob/dev/highlevelilinstruction.h#L1147.

I would definitely match what the C++ API is doing in this case. We'll look into getting the Python API fixed. Thanks for bringing it up.

rust/src/hlil/function.rs
```diff
@@ -0,0 +1,100 @@
...
+impl Eq for HighLevelILFunction {}
+impl PartialEq for HighLevelILFunction {
+    fn eq(&self, rhs: &Self) -> bool {
+        self.handle == rhs.handle
+    }
+}
+
+impl Hash for HighLevelILFunction {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.handle.hash(state);
```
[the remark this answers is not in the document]
KyleMiles: While technically pointer-based should be unique, it could change if the object gets moved.

rust/src/hlil/function.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+impl HighLevelILFunction {
+    pub(crate) unsafe fn from_raw(handle: *mut BNHighLevelILFunction) -> Self {
```
KyleMiles, line 38: It might be better to change this to `ref_from_raw` so we don't accidentally forget to wrap this in a `Ref<>` other places in the API.

KyleMiles (changes requested):
This mostly looks good to me. Please make sure to run `cargo fmt`, fix anything `cargo clippy` complains about (in your code, on latest Rust please...not the one we're currently running), and please confirm you're ready for this to be merged...you have some outstanding todo items in the top comment of this PR, so I just want to double check whether you're ready for this to be merged.

rust/src/hlil/function.rs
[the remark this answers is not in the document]
rbran: I'll add that to the todo list. That got copied from the mlil API, that on it's turn, got copied from the llil API.

rust/src/hlil/function.rs
rbran: Like the hash, this is also legacy from the llil. I'll remove it.

rbran: Should I also change that in LLIL and MLIL? Maybe create a new PR?

KyleMiles: If you fix it in HLIL, I don't mind copying it over to MLIL/LLIL. If you want to do it, you're welcome to. In this PR or in another doesn't make a difference to me.

KyleMiles: It's mostly just a timebomb/footgun. Doesn't need to block merging this PR if you don't change it.

rbran: I solved that by making all `from_raw` functions return `Ref<Self>`.

While working on it I realized that binja `Ref` is kind of an unnecessary abstraction, because all types that use `Ref` are always owned and only need to exist and a `Ref<Self>`. In this case you might just make the type own it's handle by implementing `Clone` and `Drop` directly, instead of implement `RefCountable` and make it a `Ref<Self>`.

But I diverge, and that's beyond the scope for this PR.

rust/src/hlil/function.rs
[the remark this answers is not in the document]
rbran: I implemented the hash, like in python, directly on the main Function type.

rbran:
Currently, the implementation of Instructions is a little awkward, because getting fields from the operation requires a pointer to the function:
```
let instr: HighLevelIlInstruction = block.next().unwrap();
match instr.operation() {
    HighLevelILOperation::Goto(goto) => {
        let name = goto.target(instr.function()).name();
        println!("goto {name}")
    }
    _ => println!("other"),
}
```

This could be solved, although making the struct a little less efficient, by adding a `Ref<HighLevelILFunction>` to the operator if necessary.

rust/src/hlil/function.rs
```diff
@@ -0,0 +1,104 @@
...
+impl Hash for HighLevelILFunction {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.get_function().hash(state)
+    }
+}
+
+impl HighLevelILFunction {
+    pub(crate) unsafe fn from_raw(handle: *mut BNHighLevelILFunction, full_ast: bool) -> Ref<Self> {
+        debug_assert!(!handle.is_null());
+        Self { handle, full_ast }.to_owned()
```
KyleMiles, line 41: `from_raw` functions that return `Ref<_>`s should be called `ref_from_raw` to match the rest of the API.

KyleMiles:
> adding a Ref<HighLevelILFunction> to the operator if necessary.

Sounds like a fair simple/safe fix.

> Alternatively, add all fields from HighLevelIlInstruction into each HighLevelILOperation, then make the operation an instruction directly, instead of using it as a field of instruction struct.

I'm okay with the `inst.lift()`/`inst.operands()` split right now, but are you proposing merging `.operands()` into the instruction? I'd love anything that can reduce/simply the APIs like that.

rbran:
> I'm okay with the `inst.lift()`/`inst.operands()` split right now, but are you proposing merging `.operands()` into the instruction? I'd love anything that can reduce/simply the APIs like that.

I mean, instead of having, a struct for instruction (`HighLevelILInstruction`) and a enum for all operations (`HighLevelILOperation`), like the PR is doing right now:
```rust
pub struct HighLevelILInstruction {
    pub(crate) function: Ref<HighLevelILFunction>, // and fields common to all instructions
    pub(crate) operation: HighLevelILOperation,
}
pub enum HighLevelILOperation {
    Adc(BinaryOpCarry),
    // ... other operations
}
pub struct BinaryOpCarry {
    left: usize,
    right: usize,
    carry: usize,
}
```

We change the operation enum into the instruction struct, adding all fields common for all instruction into each operation struct:
```rust
pub enum HighLevelILInstruction {
    Adc(BinaryOpCarry),
    // ... other operations
}
pub struct BinaryOpCarry {
    pub(crate) function: Ref<HighLevelILFunction>, // and fields common to all instructions
    left: usize,
    right: usize,
    carry: usize,
}
```

This way, there will be no need to pass the function reference to `lift` and related functions and avoiding the unnecessary duplication of the function Ref.

rbran:
@KyleMiles there still lots of implements/simplification to do on the API, but I spend too much time on this PR already and it's time for me to move on.

Do you need any changes before doing a merge?

KyleMiles (approved):

KyleMiles:
Sorry for the delay on this, was only able to get back to this PR this week. Thank you for your tremendous effort getting this PR through! I'm extremely glad to finally have HLIL and MLIL bindings in our API. Thank you!

I just merged this manually in f682ca69b356755fe6c06bcdd8f4fab73f7d2c78.

PR #4783, opened 2023-11-29
Fix formatters for BinaryNinja::Confidence

saagarjha:
(Disclaimer: I don't actually know how to use fmtlib, I just know that this doesn't compile on my computer)


CouleeApps:
Rebased and included the confidence in the printing too. LGTM

PR #4843, opened 2023-12-21
add sources and link to rust online docs

greaka:
After a quick exchange on Slack, I'm back with a PR.

This PR aims to enhance the DX of the docs.

When working with the api, you regularly need to look up implementations because you write your own wrapping functions for things not yet part of the official api and it quickly becomes tedious to go to the github code search for every single struct or method.

Additionally, if you skip deps when building docs for a plugin, setting `html_root_url` will cause rustdoc to instead generate a link to the struct hosted on that page.

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -16,7 +16,7 @@
 #![allow(clippy::missing_safety_doc)]
 #![allow(clippy::result_unit_err)]
 #![allow(clippy::type_complexity)]
-#![doc(html_no_source)]
+#![doc(html_root_url = "https://dev-rust.binary.ninja/")]
 #![doc(html_favicon_url = "/favicon.ico")]
 #![doc(html_logo_url = "/logo.png")]
 #![doc(issue_tracker_base_url = "https://github.com/Vector35/binaryninja-api/issues/")]


psifertex:
The increase in total file size isn't an issue but it doesn't appear to be linking to any online docs at all when it generates the documentation. Not necessarily an issue but if you were expecting it, it's not doing that. It just references local source instead.

greaka:
Your comment made me double check it locally and I found that it works just fine when passing `--no-deps` to cargo doc - just as expected. However, I noticed that I made a mistake when copy pasting the url and fixed that.

greaka:
@psifertex I just saw your commit 24504c4cf6bdc9acd90e4f95fd7185ae61b1030d and noticed that it's using the wrong link

greaka:
I'm sure this PR isn't needed anymore for this change

psifertex:
yeah, I'm doing a few passes on all the docs and I'll double check the rust stuff but I don't think that parameter was working anyway at least the way we're building the rust docs.

greaka:
I think there was a misconception about that parameter.
This is only for when a downstream app uses `--no-deps` in their doc run. Instead of simply not defining types etc. at all, it will generate links to the types hosted on https://dev-rust.binary.ninja/

psifertex:
Ahh, I gotcha. Either way, the current URL should be good.

greaka:
The current url generates this link in my docs: https://dev-rust.binary.ninja/binaryninja/binaryninja/types/struct.Conf.html

Notice the `binaryninja/binaryninja`. This turns into a redirect to the root docs instead of the type.

PR #4866, opened 2024-01-06
Fix relocation vector assignemnt

patacca:
The current implementation of the new `BinaryView::GetRelocationsAt` API suffers from a problem: the vector `result` is defined to be of size `count` but then its elements are pushed back, hence resulting in doubling its size.


patacca:
I noticed that the recent commit [69b6b8a0853a8d54151a1888e29817f67b0ee645](https://github.com/Vector35/binaryninja-api/commit/69b6b8a0853a8d54151a1888e29817f67b0ee645) already fixed the issue. There's no need for this one anymore then.

PR #4935, opened 2024-01-23
Fix RISC-V disassembly argument order for CSR instructions

pdietl:
Fix RISC-V disassembly argument order for CSR instructions

and

Fix RISC-V CSRRW, CSRRS, and CSRRC instructions disassembly to show the CSR 12-bit immediate operand as an unsigned number


pdietl:
@D0ntPanic

PR #4997, opened 2024-02-01
RISC-V: Fix disassembly of compressed JAL instruction

pdietl:
Change integer register involved via compressed JAL instruction from 0 (x0 (zero)) to 1 (x1 (ra))


pdietl:
@D0ntPanic

PR #5026, opened 2024-02-06
add function set_can_return_auto and set_can_return_user

0xcaff:
related convo https://binaryninja.slack.com/archives/C01L5RB47S4/p1707248241735869

### DIFF
--- a/rust/src/function.rs
+++ b/rust/src/function.rs
@@ -179,14 +179,14 @@ impl Function {
         }
     }
 
-    pub fn set_can_return_auto(&self, can_return: Conf<bool>) {
-        let mut raw = BNBoolWithConfidence::from(can_return);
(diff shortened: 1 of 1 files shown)


KyleMiles (changes requested):
Thank you for the PR!

We try not to require users of the Rust API to create Conf types as much as possible, I'd prefer if these functions took `T: Into<Conf<bool>>` instead. You can use [this](https://github.com/Vector35/binaryninja-api/blob/dev/rust/src/types.rs#L333) as an example.

KyleMiles (approved):

KyleMiles:
Everything looks good! Thank you for this, I appreciate it!

I'll merge this later today hopefully - just waiting on some unrelated fixes to get merged ahead of it.

PR #5034, opened 2024-02-07
add register_event_handler

0xcaff:
Expose a safe wrapper around BNRegisterBinaryViewEvent

Looking for feedback on:

* Naming. I don't have a strong grasp on the names of the things in binary ninja in general. Specifically, maybe this should be named register_binary_view_event.

* Semantics. Do we care about exposing the `BNBinaryViewEventType`? Should it be wrapped maybe? Why are traits like CustomBinaryView unsafe? I followed the unsafe here but I don't think it is necessary. The unsafe thing is that the handler will live for the lifetime of the program on the heap.  

* The approach. I tried to const specialize the EventHandler trait against the BNBinaryViewEventType so one context could be used for multiple event types but seems like const specialization in traits in rust is unstable (and along with many other const features, progress seems like it has kinda stalled)

### DIFF
--- a/rust/src/binaryview.rs
+++ b/rust/src/binaryview.rs
@@ -1452,3 +1452,64 @@ impl std::fmt::Debug for BinaryView {
         )
     }
 }
+
+pub trait BinaryViewEventHandler: 'static + Sync {
+    fn on_event(&self, binary_view: &BinaryView);
+}
+
+pub type BinaryViewEventType = BNBinaryViewEventType;
+
+/// Registers an event listener for binary view events.
+///
+/// # Example
+///
+/// ```rust
+/// use binaryninja::binaryview::{BinaryView, BinaryViewEventHandler, BinaryViewEventType, register_binary_view_event};
+///
+/// struct EventHandlerContext {
+///     // Context holding state available to event handler
+/// }
+///
+/// impl BinaryViewEventHandler for EventHandlerContext {
+///     fn on_event(&mut self, binary_view: &BinaryView) {
+///         // handle event
+///     }
+/// }
+///
+/// #[no_mangle]
+/// pub extern "C" fn CorePluginInit() {
+///     let context = EventHandlerContext { };
+///
+///     register_binary_view_event(
+///         BinaryViewEventType::BinaryViewInitialAnalysisCompletionEvent,
+///         context,
+///     );
+/// }
+/// ```
+pub fn register_binary_view_event<Handler>(event_type: BinaryViewEventType, handler: Handler)
+where
+    Handler: BinaryViewEventHandler,
+{
+    unsafe extern "C" fn on_event<Handler: BinaryViewEventHandler>(
+        ctx: *mut ::std::os::raw::c_void,
+        view: *mut BNBinaryView,
+    ) {
+        ffi_wrap!("EventHandler::on_event", unsafe {
+            let mut context = &mut *(ctx as *mut Handler);
+
+            let handle = BinaryView::from_raw(BNNewViewReference(view));
+            Handler::on_event(&mut context, handle.as_ref());
+        })
+    }
+
+    let boxed = Box::new(handler);
+    let raw = Box::into_raw(boxed);
+
+    unsafe {
+        BNRegisterBinaryViewEvent(
+            event_type,
+            Some(on_event::<Handler>),
+            raw as *mut ::std::os::raw::c_void,
+        );
+    }
+}


KyleMiles (changes requested):
Thank you for another PR! We really appreciate the effort to expand our Rust bindings!

To answer your questions...  
Naming:
 - To figure out what something should be named, I search the codebase for the core API function, find where it's wrapped in Python, and copy/paste the name the Python API uses, adjusting it for any relevant Rust conventions. In this case, looks like Python chose `add_binaryview_finalized_event`, so we should use something closer to that.

Semantics:
 - I'd prefer not to expose any core stuff, if we can avoid it. There are a handful of examples in the API already where we just do `pub type TypeName = BNTypeName;` or where we fully reimplement the underlying type and provide private conversions for the API to use when throwing things over the FFI boundary.

Approach:
 - Generally, you're on the right track. Look at [src/command.rs](https://github.com/Vector35/binaryninja-api/blob/dev/rust/src/command.rs) and take just a little more inspiration from there, and I think you'll be set. I don't think the trait needs to be unsafe, and it might answer some of your other questions too.

0xcaff:
Ok, I think this is ready for another look. Some things to note:

1. Is a closure which is `'static` and `Sync` more useful than a function pointer? When I first saw the impl https://github.com/Vector35/binaryninja-api/blob/ddce31c8d62c7ce0a385e5af65b0134c56e3b6c8/rust/src/command.rs#L49-L60 I got excited because I thought I could use it with a closure which captured context but after reading the bounds I realized it doesn't seem to be able to do much past a function pointer.

rust/src/lib.rs
```diff
@@ -128,6 +128,7 @@ mod ffi;
 pub mod architecture;
 pub mod backgroundtask;
 pub mod basicblock;
+pub mod binary_view_event_handler;
```
KyleMiles, line 131: Whatever this was named before was fine. If you're gonna name it like this, though, it would be better in `src/binaryview.rs` instead of its own file.

rust/src/binary_view_event_handler.rs
```diff
@@ -0,0 +1,66 @@
+use crate::binaryview::BinaryView;
+use binaryninjacore_sys::{
+    BNBinaryView, BNBinaryViewEventType, BNNewViewReference, BNRegisterBinaryViewEvent,
+};
+
+pub trait BinaryViewEventHandler: 'static + Sync {
+    fn on_event(&mut self, binary_view: &BinaryView);
```
KyleMiles, line 7: I don't think this should be mut? If you need mut, there should probably be a non-mut version? But I think everything should be able to be done in a non-mut way....

rust/src/binary_view_event_handler.rs
```diff
@@ -0,0 +1,66 @@
...
+///         BinaryViewEventType::BinaryViewInitialAnalysisCompletionEvent,
+///         context,
+///     );
+/// }
+/// ```
+pub fn register_binary_view_event<Handler>(event_type: BinaryViewEventType, handler: Handler)
+where
+    Handler: BinaryViewEventHandler,
+{
+    unsafe extern "C" fn on_event<Handler: BinaryViewEventHandler>(
```
KyleMiles, line 44: I don't think this needs to be unsafe. You seem to wrap the rest of the unsafe blocks down below, which is the style used in most the API.

KyleMiles (changes requested):
We require `'static` and `Sync` as a safety precaution more than anything. We could drop those requirements, but since Binary Ninja's core makes no promises about when it's going to trigger that callback (on the same thread, on another, when your context might not be valid, etc), we're going with the strictest version of those bounds for now.

rust/src/lib.rs
0xcaff: ok, moving to src/binaryview.rs

rust/src/binary_view_event_handler.rs
0xcaff: Hmm, upon further consideration I agree this instance should be &self to agree with `Command` and friends. The `&mut` makes more sense if we know for certain that binary ninja isn't going to call the fn at the same time from multiple threads and given `Sync we are not making that assumption.

0xcaff:
Thanks for the feedback, I appreciate the upfront time investment reviewing these PRs while sharing information. I plan to continue to make contributions to this project. I appreciate you taking the long view on this.

As for the PR, let me know if there's anything else you'd like changed.

0xcaff:
done! fwiw, there is a button in github which allows owners to squash the commits how they see fit. https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

KyleMiles (approved):
Thank you! We have that option disabled, but I appreciate the thought!

I'll merge this (hopefully) later this week after we exit code freeze.

PR #5047, opened 2024-02-12
Rust: refactor MLIL and HLIL instructions

mkrasnitski:
This took like a week but the end result I think is a lot more idiomatic and concise.

Here's a summary of changes:
1. Change `MediumLevelILInstruction` into a struct with `function`, `address`, and `kind` fields. This factors the function and address fields out of the enum variants. The same thing applies to `HighLevelILInstruction`.
2. Implement lifting inside the `{Medium,High}LevelILInstruction::lift` method rather than deferring to `lift` methods on each variant.
3. Remove field getters on each variant and make their fields public. The motivation here is that these getters would perform lifting of their corresponding field, which means using them invokes a kind of "partial" lifting which feels weird. Better to just go through the `{Medium,High}LevelILLiftedInstruction` structs.
    - Some of the fields were tuples that I expanded out into named fields. For example, list metadata fields included the index of the first element as well as the length. I did my best to give some good names, sticking to the `num_*` and `first_*` template, but probably some could be named something better.
4. Move the `operands` method onto the lifted instructions, as it called the now-removed field getters. Also, it returns an owned `Vec` of operands, each of which also own their data. This introduces some extra clones but seems a bit easier to handle for users, and avoids exposing the custom `OperandIter` type publicly.
5. Instantiate instructions directly from the list of raw operands rather than using `new` methods. Then, delete the constructors. In the end, this lets `operation.rs` just be a big collection of struct definitions.
6. The `OperandList` type was duplicated between MLIL and HLIL, so I made a generic version which can hold either type of function. Also, the implementation of `ExactSizeIterator` for the type was wrong because the value of `self.remaining` was being decreased 4 at a time and would be out of sync with the actual number of values given out.

Some miscellaneous changes:
- Rename the `ConstantData` instruction variants to just `ConstData` to avoid confusion with the `types::ConstantData` struct.
- Add `LiftedJump` and `LiftedLabel` types for HLIL. These were removed halfway through the original HLIL PR but I don't know why.


KyleMiles (changes requested):
Generally, I like these changes. A net -4k code diff with the same behavior sure is something!

Nitpick: I'm of two minds on the changes such as `function: &HighLevelILFunction,` → `function: Ref<HighLevelILFunction>,`. On one hand, it's more clear to the user that you're making a copy of the shared pointer containing a `HighLevlILFunction`, on the other, the user shouldn't care and we should just do it for them under the hood like we already were...in general that should cut down on littering users' code with `.to_owned` that the sacrifice of writing that ourselves instead.

This PR reintroduces something I rejected in the original PR, which is that we one again have a bunch of different types to remember that all represent a HLIL Instruction:

```
HighLevelILInstruction
HighLevelILInstructionKind

HighLevelILLiftedInstruction
HighLevelILLiftedInstructionKind

HighLevelILOperand
HighLevelILLiftedOperand
```

Which is a lot to remember. We start with the HLILInstruction object, from which we can get either a `HighLevelILInstructionKind` (from `.kind`) or `HighLevelILLiftedInstruction` (from `.lift()`). But now that `HighLevelILInstructionKind` is fully structured/provided when you have a `HighLevelILInstruction`, it's making me question the utility of having the lifted instructions at all. They kinda seem redundant now, no? Having worked on this, what are your thoughts?

mkrasnitski:
One solution is to remove the non-lifted versions and just have a single type of instruction. Having a single `Kind` enum then won't be much of an issue. It seemed there was a want for a easy and fast version of the IL instructions (hence the non-lifted versions, which also implement `Copy`), so it's worth considering how much of a performance tradeoff working with just lifted instructions really is. Some of the variants are recursive and store whole instructions within them, and I don't know how large a single instruction can get, or how many sub-nodes it can contain.

As for your first concern, the `HighLevelILInstruction::new` constructor is `pub(crate)` and so the change to take `Ref` directly is purely internal. Is this what you're referring to, or are there places where calling `.to_owned()` is now necessary?

KyleMiles:
I missed that `HighLevelILInstruction::new` is crate-internal. You can disregard that comment. As for removing `Lifted` or non-lifted, perhaps we should reconsider that later. This PR should be good as-is regardless if we decide to cull one of those later.

KyleMiles (approved):

KyleMiles:
I'll leave it to @plafosse to decide if we should merge this before code freeze.

KyleMiles (approved):

rust/src/mlil/function.rs
```diff
@@ -89,6 +96,21 @@ impl MediumLevelILFunction {
...
+    type BNInstruction = BNMediumLevelILInstruction;
+    type Instruction = MediumLevelILInstruction;
+
+    fn il_instruction_from_idx(&self, expr_idx: usize) -> Self::Instruction {
+        self.instruction_from_idx(expr_idx)
+    }
+
+    fn operands_from_idx(&self, expr_idx: usize) -> [u64; 5] {
+        let node = unsafe { BNGetMediumLevelILByIndex(self.handle, expr_idx) };
+        assert_eq!(node.operation, BNMediumLevelILOperation::MLIL_UNDEF);
```
mkrasnitski, line 109: Was just testing the `mlil_visitor` example and seems like this assertion ends up failing for some reason. I'll have to diagnose this and figure out what the behavioral difference here is.

mkrasnitski: Figured it out - my iterator implementation was only accounting for lists of non-zero length.

mkrasnitski:
Let me know if I should squash the extra bugfix commits.

KyleMiles:
Thank you again for this awesome PR! Sorry for the delay on getting it merged.

PR #5072, opened 2024-02-18
Fix RISC-V CSRRW, CSRRS, and CSRRC instr disass

pdietl:
Make the CSR instructions show the CSR 12-bit immediate operand as an unsigned integer.
This is a special case of the I-instruction format, where it is same except that the 12-bit immediate is treated as unsigned rather than signed.


pdietl:
@D0ntPanic here I only change the interpretation of the CSR number as unsigned rather than signed. 
My other PR where I changed the order in which the operands are displayed was incorrect, at least compared to how the GNU assembler expects the instruction operands.

PR #5107, opened 2024-02-26
Require Hash & Eq implementations for all Register and Flag implementations

emesare:
Already present in all the core implementations ([1], [2], [3]), however when being generic over `Architecture` you won't get them. I.e.

```rs
// A: Architecture
let registers: HashMap<A::Register, u128> = HashMap::new();
```

Requiring both Hash & Eq for `Register` and `Flag` implementations means we can use them as `HashMap` keys, which I have ran into on multiple occasions.

This is a very small PR and probably should be expanded on further later to more traits.

[1]: https://github.com/emesare/binaryninja-api/blob/91d762fb5c8666fbe830a48ce6354819cc802e08/rust/src/architecture.rs#L684
[2]: https://github.com/emesare/binaryninja-api/blob/91d762fb5c8666fbe830a48ce6354819cc802e08/rust/src/architecture.rs#L581
[3]: https://github.com/emesare/binaryninja-api/blob/91d762fb5c8666fbe830a48ce6354819cc802e08/rust/src/architecture.rs#L781

### DIFF
--- a/arch/riscv/src/lib.rs
+++ b/arch/riscv/src/lib.rs
@@ -6,6 +6,7 @@
 
 use std::borrow::Cow;
 use std::fmt;
+use std::hash::Hash;
 use std::marker::PhantomData;
 
 use binaryninja::relocation::{Relocation, RelocationHandlerExt};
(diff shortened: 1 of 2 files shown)


KyleMiles (changes requested):
I'm perfectly fine adding these trait bounds, but merging this as-is would break the build for [riscv](https://github.com/Vector35/binaryninja-api/tree/dev/arch/riscv). You, @D0ntPanic, or I should fix that before this merges.

emesare:
I went ahead and implemented `Hash`, `PartialEq` & `Eq` for the riscv register type.

KyleMiles (approved):

KyleMiles:
Looks great, thank you!

PR #5145, opened 2024-03-01
ensures python site package directory exists

deadc0de6:
This PR ensures the first entry in the python site packages is created. 
This avoids issues when, for some reason, the first directory of `site.getsitepackages()` does not exist.

### DIFF
--- a/scripts/install_api.py
+++ b/scripts/install_api.py
@@ -40,6 +40,21 @@ def check_virtual_environment() -> bool:
     return False
 
 
+def getsitepackage() -> str:
+    """
+    gets the site package and creates it if needed
+    returns empty if fails
+    """
+    install_path = getsitepackages()[0]
+    if not os.path.exists(install_path):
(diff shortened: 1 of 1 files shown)


xusheng6:
I have approved this PR. Looks good to me, though Jordan brought up some concerns that the sitepackages may not be writable anyways. However, it seem to me that in the case of a pipenv, etc, the `site.getsitepackages` will return a unique path for each environment, which is also often times writable. If we use `getusersitepackages`, then it will actually always return the same path, which does not consider the presence a pipenv. Here is some output that could help explain this:

(Running inside pipenv)

```
>>> site.getsitepackages()
['/Users/xusheng/.local/share/virtualenvs/xusheng-2ZPXY7r0/lib/python3.11/site-packages']
>>> site.getusersitepackages()
'/Users/xusheng/Library/Python/3.11/lib/python/site-packages'
```

PR #5163, opened 2024-03-07
Expose get_expr_count() on python llil and mlil

samrussell:
This is already exposed in the CPP interface, and is useful for workflows for enumerating all expressions in an ILFunction


CouleeApps:
This function also exists on high level il functions, but I'll merge this and add that myself.

PR #5166, opened 2024-03-10
Update cookbook.md

jrmo14:
Document how to call plugins headless


psifertex:
Great suggestion! My only hesitation is whether this should be mentioned in the batch processing section? Maybe we can add it here and I'll just add a reference to it from the batch section. I'll decide tomorrow when back in the office. 

Thanks for the PR.

PR #5176, opened 2024-03-12
added all the ps and other ppc extensions, need to decide on getopt for windows stuff and resolve default architecture

v1X3Q0:
Adding the other powerpc extensions, unfortunately unix to windows differences require some external libraries.

- Need to decide if its ok to either submodule https://github.com/v1X3Q0/getopt-for-windows.git or if we should not include windows support.
- Default to ppc_ps for the architecture right now, need to fix it so it defaults to ppc and correctly allows selection of each architecture arbitrarily.
- To have each architecture properly selected, may require adding platforms for each.

### DIFF
--- a/arch/powerpc/disassembler.cpp
+++ b/arch/powerpc/disassembler.cpp
@@ -19,7 +19,7 @@ thread_local csh handle_lil = 0;
 thread_local csh handle_big = 0;
 
 extern "C" int
-powerpc_init(void)
+powerpc_init(int cs_mode_arg)
 {
 	int rc = -1;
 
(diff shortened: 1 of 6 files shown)


galenbwill:
Just to put it here, I've asked the author (on slack) to remove the getopt dependency, so test_asm/test_disasm won't support Windows for now.

v1X3Q0:
I'm mariomain, windows stuff has been removed. will still look into the other 2 bullets in a bit.

arch/powerpc/disassembler.cpp
```diff
@@ -12,14 +12,92 @@ architecture plugin picture.
...
+
+	res->insn.id = PPC_INS_BN_XXPERMR;
+	res->detail.ppc.op_count = 3;
+	strncpy(res->insn.mnemonic, "xxpermr", sizeof(res->insn.mnemonic));
+}
+
+bool PerformLocalDisassembly(const uint8_t *data, uint64_t addr, size_t &len, decomp_result* res, bool bigendian)
+{
+	uint32_t local_op = 0;
+	uint32_t insword = 0;
```
galenbwill, line 79: `insword` is always 0 when passed into the `ppc_fcmpo`/`ppc_xxpermr` functions

galenbwill (changes requested):

galenbwill:
Squashed and merged in builds >= 4.3.6580

PR #5180, opened 2024-03-13
Fixed segment / section misalignment in TE view

zznop:
This commit fixes an issue in the TE view while creating segments and sections. Segments and sections are calculated from the base of the original PE file, prior to the PE headers being stripped and replaced with the TE header


plafosse (approved):
LGTM

PR #5181, opened 2024-03-13
Fix mismatching operand size when lifting `movd/movq/vmovd/vmovq` instructions

nbailluet:
Moved from https://github.com/Vector35/arch-x86/pull/38.

### DIFF
--- a/arch/x86/il.cpp
+++ b/arch/x86/il.cpp
@@ -2110,7 +2110,7 @@ bool GetLowLevelILForInstruction(Architecture* arch, const uint64_t addr, LowLev
 	case XED_ICLASS_MOVDIR64B:
 		il.AddInstruction(
 			WriteILOperand(il, xedd, addr, 0, 0,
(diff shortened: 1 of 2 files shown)


galenbwill:
Copied this over from the original PR:

> Prior to this fix `movd eax, xmm0` was lifted to `eax = xmm0` instead of `eax = xmm0.d`.
> In SSA form, this resulted in a semantically invalid zero-extend instruction that made no sense: `rax#1 = zx.q(zmm0.xmm0)`, which is wrong since `xmm0` is bigger than a quad-word.
> 
> Fixed by specifying the explicit size of the source operand:
> https://github.com/Vector35/arch-x86/blob/a87d5df489eed969dec7f36235ebd81c3ff87acb/il.cpp#L2104-L2115

galenbwill:
I am closing this PR, as there were a number of problems with it. I will link to the corrected commit here once it's pushed (I'm still testing it).

PR #5182, opened 2024-03-13
Improve TE Header type definitions

mkrasnitski:
Brings the TE header definitions more in-line with the PE view:
1. Rename header fields to camelCase
2. Use arrays when applicable
3. Define some named data symbols

What remains to be done is to define the three enum types used for the "machine", "subsystem", and "characteristics" fields. Since these enums are unmodified from the PE standard, I'm not sure I want to redefine them on their own in the TE view, or if they should somehow be imported/re-used between the two views.

### DIFF
--- a/view/pe/teview.cpp
+++ b/view/pe/teview.cpp
@@ -160,7 +160,8 @@ void TEView::AssignHeaderTypes()
 	headerBuilder.AddMember(Type::IntegerType(4, false), "addressOfEntryPoint");
 	headerBuilder.AddMember(Type::IntegerType(4, false), "baseOfCode");
 	headerBuilder.AddMember(Type::IntegerType(8, false), "imageBase");
-	headerBuilder.AddMember(Type::ArrayType(Type::NamedType(this, dataDirectoryTypeName), 2), "dataDirectory");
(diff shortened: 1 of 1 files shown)


view/pe/teview.cpp
```diff
@@ -143,52 +143,56 @@ void TEView::CreateSections()
...
-			m_imageBase + m_headersOffset + EFI_TE_IMAGE_HEADER_SIZE + (EFI_TE_SECTION_HEADER_SIZE * i),
-			Type::NamedType(this, sectionTypeName)
-		);
-	}
+	QualifiedName sectionName = string("TE_Section_Header");
+	auto sectionTypeId = Type::GenerateAutoTypeId("te", sectionName);
+	QualifiedName sectionTypeName = DefineType(sectionTypeId, sectionName, sectionType);
+	DefineDataVariable(
+		m_imageBase + m_headersOffset + EFI_TE_IMAGE_HEADER_SIZE,
+		Type::ArrayType(Type::NamedType(this, sectionTypeName), header.numberOfSections)
```
zznop, line 193: `header` is not defined in this function. Use `m_sections.size()`.

mkrasnitski: Done.

zznop:
Thank you for the contribution!

PR #5191, opened 2024-03-18
Add lifting for some ppc64 instructions

galenbwill:
added lifting for EXTSW LD LDU LDUX LDX SLD SRAD SRADI SRD STD STDU STDUX STDX


galenbwill:
Merged to dev, should be in builds > 4.1.5003 (no build number allocated yet)

PR #5223, opened 2024-03-29
Rust: Add `name` method and `index` field to MLIL and HLIL instructions

mkrasnitski:


KyleMiles:
Thank you for the PR!

PR #5238, opened 2024-04-05
reduce complexity by removing rc::Ref

rbran:
Types that use `rc::Ref` only exists as References, so it makes sense to replace all occurrences of `Ref<T>` with `T` and move the `Ref::inc_ref` and `Ref::dec_ref` implementations to `Clone` and `Drop`.


KyleMiles:
This is very interesting. I'll have to talk about this internally with the team to see if we want this. Thank you for the initial effort!

PR #5239, opened 2024-04-05
Remove unecessary architecture impl requirements

rbran:
Requiring the user to implement `Copy` in all types of `Architecture` is unnecessary and very limiting for the user. Not allowing non-copy types, like String, Vec, HashMap, etc.

### DIFF
--- a/rust/src/architecture.rs
+++ b/rust/src/architecture.rs
@@ -190,7 +190,7 @@ pub trait RegisterInfo: Sized {
     fn implicit_extend(&self) -> ImplicitRegisterExtend;
 }
 
-pub trait Register: Sized + Clone + Copy + Hash + Eq {
+pub trait Register: Sized + Clone + Hash + Eq {
     type InfoType: RegisterInfo<RegType = Self>;
 
     fn name(&self) -> Cow<str>;
@@ -212,7 +212,7 @@ pub trait RegisterStackInfo: Sized {
     fn stack_top_reg(&self) -> Self::RegType;
 }
 
-pub trait RegisterStack: Sized + Clone + Copy {
+pub trait RegisterStack: Sized + Clone {
     type InfoType: RegisterStackInfo<
         RegType = Self::RegType,
         RegInfoType = Self::RegInfoType,
@@ -230,7 +230,7 @@ pub trait RegisterStack: Sized + Clone + Copy {
     fn id(&self) -> u32;
 }
 
-pub trait Flag: Sized + Clone + Copy + Hash + Eq {
+pub trait Flag: Sized + Clone + Hash + Eq {
     type FlagClass: FlagClass;
 
     fn name(&self) -> Cow<str>;
@@ -242,7 +242,7 @@ pub trait Flag: Sized + Clone + Copy + Hash + Eq {
     fn id(&self) -> u32;
 }
 
-pub trait FlagWrite: Sized + Clone + Copy {
+pub trait FlagWrite: Sized + Clone {
     type FlagType: Flag;
     type FlagClass: FlagClass;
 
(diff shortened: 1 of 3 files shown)


rssor:
So those types correspond to what are simply `uint32_t`s at the API boundary, they are intended to be _extremely_ cheap to throw around/copy. They should rarely be more than a newtype around an integer or enum.

Similarly, they are likely to be used in ways where requiring .clone() to be invoked frequently would be burdensome.

rssor:
The MaybeUninit and other minor changes all look good though

rbran:
I have a project that parses at runtime the architecture information from a file into a `Arc<Arch>`, so registers needs to have the usize (reg index) and the Arc for the architecture, making impl copy impossible.

I believe that copy making things faster is a misconception. Impl copy makes it almost impossible to make a clone slow, but it doesn't necessarily make it faster.

The `MaybeUninit` is required by my code, initializing non-primitives with `zeroed` can be UB, in my code it caused a crash.

rssor:
It's less about speed and more about the burden of having to use .clone() everywhere, although since for most API consumers working with the ILs they'll be working with CoreRegister/CoreFlag etc. which _are_ Copy perhaps the burden is less significant for those cases where only these bounds apply.

I've also not been a fan of the fact that there's no way to forcibly tie the Register/Flag etc. types to the 'right' architecture object they came from, which is exactly what you're trying to do, which is another point in favor of your change.

Some other thoughts that might be possible to exploit: all architectures are effectively `'static` as we can't ever free them, so once registered it will never be possible to free or move them.

rbran:
> Some other thoughts that might be possible to exploit: all architectures are effectively 'static as we can't ever free them, so once registered it will never be possible to free or move them.

That's a good point, if it's static, I could replace all the `Arc` with `&'static `, allowing the `Copy` to be implemented.

Although, I assume that `headless::init()` and `headless::shutdown()` could be called multiple times, not making the lifetime truly static.

Something like this:
```rust
// test all the architectures impl
for arch in archs {
    let parsed_arch = Arc::new(custom_arch::parse(arch).unwrap());
    binaryninja::headless::init();
    binaryninja::architecture::register_architecture(parsed_arch.name(), parsed_arch);
    binaryninja::headless::shutdown();
    drop(parsed_arch);
}
```

rssor:
Correct, `BNArchitecture` and `BNPlatform` will never/can never be freed; `BNCustomArchitecture` doesn't even have a callback for it. That said, requiring that cast to `'static` in an API consumer's code is not ideal.

I do suspect that I placed the traits for registers/flags/etc. in the wrong location initially. It would have been far better I think to have a fixed object that contains the Architecture reference _and_ the interior implementor's implementation (essentially, make a representation similar to `CoreRegister` used for _all_ arch objects, and only have the interior `id` field representation be switched out), and then traits for CustomRegister/CustomFlag etc. could be more like:

```
pub trait RegisterImplementation: Sized + Clone + Copy + Hash + Eq {
    type InfoType: RegisterInfo<RegType = Self>;
    type Arch: Architecture<>...;

    fn name(&self, arch: &Self::Arch) -> Cow<str>;
    fn info(&self, arch: &Self::Arch) -> Self::InfoType;

    /// Unique identifier for this `Register`.
    ///
    /// *MUST* be in the range [0, 0x7fff_ffff]
    fn id(&self) -> u32;
}
```

That way the bindings could always handle making the arch handle available to the arch-related implementations, and that wrapped object would simply have `name(&self)`/`info(&self)` methods that all API consumers would interact with.

rssor:
I'm not sure a lot of my efforts to share a lot of the code between a custom architecture and interacting with the API were worth it, I wonder if simply cleanly separating Architecture/CustomArchitecture entirely to cut down on all the use of associated types is worth considering.

The primary justification was to make it possible to implement calling conventions on both architecture objects from the API and implemented objects, but I wonder if there might have been a better way to go about that.

rssor:
To go back up to the trait reorganization, it would also free up some useful names, `CoreRegister`/`CoreFlag` etc. could just become `Register`/`Flag`, and then the trait used to implement them could be more clearly named as the implementation used by the struct that everything shares.

Would clean up some of the names for types used in the API, I was never really happy with the most visible names being for the trait rather than the actual _objects_ whose name shows up in the API docs far more often.

KyleMiles (changes requested):
Marking this as needing changes so we don't accidentally merge!

rbran:
I think those restrictions could be removed, but I'm closing this PR for now, because having the custom architectures being fully static makes this unnecessary.

PR #5244, opened 2024-04-06
replace BStr with str

rbran:
The objective of this PR is to make the API more natural to use.

This PR:
* removes the BnStr completely, in favor of `core::ffi::CStr` internally and `str` publicly.
* remove the generic from `FunctionParameter`, now always implemented using String.
* changing the `name` field in both `EnumerationMember` and `StructureMember` to String.

Is also possible to have the BnString removed from the API, by removing it's `pub` flag, and using it only internally, this way, only `String` and `str` are exposed. But that should be implemented in a separated PR.

### DIFF
--- a/rust/examples/basic_script/src/main.rs
+++ b/rust/examples/basic_script/src/main.rs
@@ -26,9 +26,7 @@ fn main() {
                         .get_data(),
                     addr,
                 ) {
-                    tokens
-                        .iter()
-                        .for_each(|token| print!("{}", token.text().as_str()));
+                    tokens.iter().for_each(|token| print!("{}", token.text()));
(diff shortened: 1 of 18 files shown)


KyleMiles (approved):
This is tremendous, thank you! This really cleans up a lot of APIs. We're gettin' Rusty!

I think we should remove BnString from pub if we can, absolutely agree.

We should consider `EnumerationMember` and `StructureMember` taking a `&str` instead, but I'm not sure what the impact of that would be.

This has lots of internal merge conflicts, so I'm going to fix those and make sure everything passes through CI before merging this. Please don't push anything in the meantime.

PR #5248, opened 2024-04-08
Simplify Edge and EdgeContext to own its references

### LINKED ISSUE #5212: [Rust API] Impossible to use block `Edge` due to lifetime error
I may be missing something, but it looks like it is impossible to use the edges of a basic_block (`outgoing_edges` or `incoming_edges`)  due to some lifetimes shenanigans errors.

Simple example:

```rust
let m_func = func.medium_level_il().unwrap();
let basic_blocks = m_func.basic_blocks();
let basic_block = basic_blocks.get(0);
let edges = basic_block.outgoing_edges();
for edge in edges.iter() { // ERROR: `edges` does not live long enough
}
```

Here is the full compilation error:
```txt
error[E0597]: `edges` does not live long enough
  --> src/lib.rs:72:21
   |
71 |         let edges = basic_block.outgoing_edges();
   |             ----- binding `edges` declared here
72 |         for edge in edges.iter() {
   |                     ^^^^^ borrowed value does not live long enough
...
99 |     }
   |     -
   |     |
   |     `edges` dropped here while still borrowed
   |     borrow might be used here, when `edges` is dropped and runs the `Drop` code for type `Array`
```

rbran:
Simplify the `Edge` and `EdgeContext` structs to fix #5212

### DIFF
--- a/rust/src/basicblock.rs
+++ b/rust/src/basicblock.rs
@@ -25,14 +25,14 @@ enum EdgeDirection {
     Outgoing,
 }
 
-pub struct Edge<'a, C: 'a + BlockContext> {
+pub struct Edge<C: BlockContext> {
     branch: super::BranchType,
     back_edge: bool,
-    source: Guard<'a, BasicBlock<C>>,
-    target: Guard<'a, BasicBlock<C>>,
+    source: Ref<BasicBlock<C>>,
+    target: Ref<BasicBlock<C>>,
 }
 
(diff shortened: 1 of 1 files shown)


KyleMiles:
This looks good to me, but I'm going to assign @rssor as a reviewer in case he has any perf or other concerns.

rbran:
Alternatively the PR #5252 completelly rewrite the `*ArrayProvider` implementation, fixing this bug and others.

PR #5252, opened 2024-04-09
reimplement the Array using GAT

### LINKED ISSUE #5212: [Rust API] Impossible to use block `Edge` due to lifetime error

rbran:
Fix multiple problems with the Array implementation, including but not limited to:

* multiple transmutes without the flag `#[repr(transparent)]`.
* remove unnecessary use of `Guard`, replacing them with simple references.
* fix #5212  by rewriting the `*ArrayProvider` trait using [GAT](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html).

TODO: 
- [ ] `DataVariable` and `NameAndType` where previously incorrectly transmuted, and was not fixed yet.


mkrasnitski:
Looking on from the sidelines, it's a little difficult to understand the changes to the `Array` API you made - could you summarize them?

rbran:
There is no changes to the API itself, it's just implement differently.

mkrasnitski:
Well, you've removed the `ArrayGuard` type as well as reworked all the array-related traits. How come the `ArrayProvider` trait is necessary? Why created dedicated `BasicBlocks`, `BasicBlockEdges`, and `CallingConventions` structs instead of just using `Array` like before?

rbran:
I didn't found any difference between `ArrayGuard` and `Array`, so I just deleted `ArrayGuard`.

It's not really necessary, it's just a way to re-utilize functions from the body of the trait and the `Iterator` implementation.

Anything that is just a simple array (basic slice, pointer and len) is defined as an `Array<T>`, anything else that have any other complexity should make it's own implementation.
I don't like the idea of having a basic "class" that preoccupy itself into allowing all the complexities of it's children, so my philosophy was to have simple implementation, and anything complex should by implemented according to with it's own complexities.

rbran:
Oh, now I see `ArrayGuard` lacks the `Drop` implementation. I'll fix the 2 places that uses it in `demangle.rs`

rssor:
`ArrayGuard` was meant to be used where there was an array _inside_ some other object so you could work with it without the owning semantics of `Array`, though I think the original motivation must have been removed at some point. The use in demangle looks like it possibly should have been with the normal Array wrapper.

I'm wary of some of the transmutes instead of the explicit creation of wrapper objects and use of `Guard<>` though.

rbran:
ArrayGuard is only used locally on `demangle.rs` so I just used `ManuallyDrop` instead.

mkrasnitski:
> Anything that is just a simple array (basic slice, pointer and len) is defined as an `Array<T>`, anything else that have any other complexity should make it's own implementation. I don't like the idea of having a basic "class" that preoccupy itself into allowing all the complexities of it's children, so my philosophy was to have simple implementation, and anything complex should by implemented according to with it's own complexities.

The thing is, traits aren't classes and IMO using the type system (via trait bounds on `Array`) to turn special cases (like the three special types I mentioned) into the general case is often a good thing. As an example, the implementation of `Iterator::next` for a type can be as complex as is needed to produce an iterator that the user can use all the rest of the standard library's combinators with. Likewise, isn't `Array` supposed to represent a `Vec` whose memory is owned by the core?

rbran:
> isn't Array supposed to represent a sort of Vec whose memory is owned by the core?

In most cases yes, but the "wrapper" function and a context makes it more similar to a `iter.map(...)` in other cases.

rbran:
close in favor of #5261

PR #5263, opened 2024-04-11
replace Vec<T> with Box<[T]> where Vec::capacity == Vec::len

rbran:
The `assert!(tokens.len() == tokens.capacity());` is not a verification that we can make.

From the `Vec::shrink_to_fit` documentation: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.shrink_to_fit

> The resulting vector might still have some excess capacity...

`Box<[_]>` avoid this problem, because it's basically a non-growing `Vec<_>`, so there is no capacity.

### DIFF
--- a/rust/src/disassembly.rs
+++ b/rust/src/disassembly.rs
@@ -412,8 +412,10 @@ impl Default for DisassemblyTextLine {
 
 impl Drop for DisassemblyTextLine {
     fn drop(&mut self) {
-        let ptr = core::ptr::slice_from_raw_parts_mut(self.0.tokens, self.0.count);
-        let _ = unsafe { Box::from_raw(ptr) };
+        if !self.0.tokens.is_null() {
+            let ptr = core::ptr::slice_from_raw_parts_mut(self.0.tokens, self.0.count);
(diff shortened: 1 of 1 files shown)


rust/src/disassembly.rs
```diff
@@ -307,10 +307,9 @@ impl std::fmt::Display for DisassemblyTextLine {
 }
 
 impl From<Vec<InstructionTextToken>> for DisassemblyTextLine {
-    fn from(mut tokens: Vec<InstructionTextToken>) -> Self {
-        tokens.shrink_to_fit();
+    fn from(tokens: Vec<InstructionTextToken>) -> Self {
+        let mut tokens: Box<[_]> = tokens.into();
```
mkrasnitski, line 311: You can use `Vec::into_boxed_slice` to avoid having to annotate the type here.

mkrasnitski, line 311:
```suggestion
        let mut tokens = tokens.into_boxed_slice();
```
You can use `Vec::into_boxed_slice` to avoid having to annotate the type here.

PR #5265, opened 2024-04-11
Base address detection Triage view widget and C++ API

zznop:


plafosse (changes requested):

examples/triage/baseaddress.cpp
```diff
@@ -0,0 +1,356 @@
...
+	if (str.startsWith("0x"))
+		return str.mid(2).toULongLong(nullptr, 16);
+	return str.toULongLong();
+}
+
+
+void BaseAddressDetectionThread::run()
+{
+	BaseAddressDetectionQtResults results;
+	uint32_t alignment = HexOrDecimalQStringToUint32(m_inputs->AlignmentLineEdit->text());
```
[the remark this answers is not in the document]
zznop: I tried out the `ParseExpression` API. I don't think it's a good fit for this for a couple reasons:
1. It treats all inputs that can be represented as a hex string as hexadecimal. For example, if I set the input to `80` for the alignment value it parses it as `0x80`. I'd like the user to have the option to supply decimal inputs and if they want to supply hex then they can prefix it with `0x`. 
2. `ParseExpression` also accepts symbol expressions / strings. This makes for weird error messages stating that the "symbol can't be found" if a user tries to erroneously input a string value into a numeric field.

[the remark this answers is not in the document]
plafosse: I'll leave it up to you but, they can provide decimal `0n80`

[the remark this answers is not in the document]
plafosse: This is standard throughout our API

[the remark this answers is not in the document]
zznop: I'll change it for the sake of consistency then. My concern is users who want to set the alignment value in base detection to 16 bytes (for example) are probably just going to enter "16" and expect it to be interpreted as decimal 16, as opposed to 22. I also don't like that they can enter a symbol string erroneously into these numeric fields and `ParseExpression` will return true and set the result to the address of the symbol.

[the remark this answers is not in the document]
plafosse: I wonder if we can show that in a preview label to the right:
`0x16 (22)`
Just a thought

examples/triage/baseaddress.cpp
```diff
@@ -0,0 +1,427 @@
...
+			m_status->setText("Completed with no results");
+		m_preferredBase->setText("Not available");
+		m_confidence->setText("Not available");
+	}
+	else
+	{
+		HideResultsWidgets(false);
+		if (results.Status.empty() && !m_worker->IsAborted())
+			m_status->setText("Completed with results");
+		m_preferredBase->setText(QString("0x%1").arg(QString::number(results.Scores.rbegin()->second, 16)));
```
plafosse, line 191: nit: No need to convert to string `.arg` can handle that for you `QString("0x%1").arg(results.Scores.rbegin()->second, 0, 16)`

examples/triage/baseaddress.cpp
```diff
@@ -0,0 +1,427 @@
...
+	}
+
+	m_resultsTableWidget->clearContents();
+	size_t numRows = 0;
+	for (auto rit = results.Scores.rbegin(); rit != results.Scores.rend(); rit++)
+		numRows += results.Reasons.at(rit->second).size();
+
+	m_resultsTableWidget->setRowCount(numRows);
+	size_t row = 0;
+	for (auto rit = results.Scores.rbegin(); rit != results.Scores.rend(); rit++)
```
plafosse, line 205: nit: Better - `for(auto & [score, baseaddr] : results.Scores)`

examples/triage/baseaddress.cpp
```diff
@@ -0,0 +1,427 @@
...
+	grid->addWidget(new QLabel("Max Pointers:"), row, column + 2, Qt::AlignLeft);
+	m_inputs.MaxPointersPerCluster = new QLineEdit("0n128");
+	grid->addWidget(m_inputs.MaxPointersPerCluster, row++, column + 3, Qt::AlignLeft);
+
+	m_advancedSettingsGroup = new ExpandableGroup(grid);
+	m_advancedSettingsGroup->setTitle("Advanced Settings");
+}
+
+
+BaseAddressDetectionWidget::BaseAddressDetectionWidget(QWidget* parent, BinaryNinja::Ref<BinaryNinja::BinaryView> bv)
```
plafosse, line 346: You should be initializing the parent `: QWidget(parent)` otherwise none of the objects stored in here know when to get cleaned up

plafosse (approved):
A couple of little nits but otherwise it looks really good.

examples/triage/baseaddress.cpp
zznop: The `QString::number` call is needed. It's converting the preferred base address to a hex string. If I change it to your recommendation I get the following:

image

zznop: Ah, that did it. I mis-read 😅

examples/triage/baseaddress.cpp
plafosse: Yeah of course...but you know c++20 added: https://en.cppreference.com/w/cpp/ranges/reverse_view

plafosse (approved):
Looks good to merge

PR #5274, opened 2024-04-14
fix metadata impl from using impl from_iterator

rbran:


rbran:
Close in favor of #5276

PR #5306, opened 2024-04-22
implement open/create database with progress function

rbran:


rbran:
Is possible to change the function from a `fn` to a `FnMut`, but I could not locate where this free is implemented, so to avoid leaks I implemented using static functions.

KyleMiles (approved):

PR #5307, opened 2024-04-22
only create FlowGraph with Ref

rbran:


rbran:
It's unclear no me if `BNShowGraphReport` require a owned FlowGraph object or only a reference to it. I assumed only a reference, because the method `BinaryViewExt::show_graph_report` received a `FlowGraph` parameter and not a `Ref<FlowGraph>`

KyleMiles (approved):

PR #5308, opened 2024-04-22
add StructureBuilder missing python methods

rbran:


rbran:
Could the non-mutable setters be problematic? What if the user hold a reference to the builder (or internals) before a setter call?

Maybe a code like this could cause problems:
```rust
let builder: StructureBuilder = todo!();
let members = builder.members();
builder.clear_members();
for member in members.iter() {
 ...
}
```

This could be solved by having all the setters functions require `&mut self` and return an Array tied to the self lifetime. Or simply clone all the members and return a `Vec<StructureMember>` instead.

KyleMiles:
This isn't building locally for me... Complaining about `CoreArrayWrapper` not accepting a lifetime parameter. Not sure why it works in CI, though. I'll look into this more later.

rbran:
This was written before the GAT PR, I'll fix this.

rbran:
@KyleMiles Fixed.

PR #5318, opened 2024-04-24
Remove the Architecture Handle

rbran:
The Architecture handle returns only a pointer to the Architecture itself, it seems redundant.

To facilitate the removal, changed all `CoreArchitecture` into `&'static CoreArchitecture`, as it only exist as 'static ref.


emesare:
Unlikely to happen, currently the architecture objects are static globals but this may change in the not-near future, thank you for the PR.

PR #5338, opened 2024-04-26
Python API and docs for base address detection

### LINKED ISSUE #3391: Automated base address guesser for raw firmware images
Considering one of the first things that needs to be figured out when loading a new firmware is the base address it would make sense that we provide a suggested value. In discussing this idea internally we came up with a potential solution that we believe has a high likely hood of succeeding for single image firmwares.

### LINKED ISSUE #5235: Base address detection for position-dependent raw binaries
Feature that aids in identifying the runtime base address of position-dependent raw binaries. This feature should be accessible through the UI as well as exposed to the headless API.

zznop:
Closes #5235 
Closes #3391 

Currently writing unit tests

### DIFF
--- a/python/basedetection.py
+++ b/python/basedetection.py
@@ -21,7 +21,18 @@
 
 import ctypes
 from typing import Optional
+from dataclasses import dataclass
 from . import _binaryninjacore as core
+
+
+@dataclass
+class BaseAddressDetectionReason:
+    """``class BaseAddressDetectionReason`` is a class that is used to store information about why a base address is a
+    candidate"""
+
+    pointer: int
+    offset: int
+    type: int
 
 
 class BaseAddressDetection:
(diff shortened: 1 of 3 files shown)


python/basedetection.py
```diff
@@ -0,0 +1,319 @@
...
+
+        count = ctypes.c_size_t()
+        reasons = core.BNGetBaseAddressDetectionReasons(self._handle, base_address, ctypes.byref(count))
+        if count.value == 0:
+            return []
+
+        result = list()
+        for i in range(count.value):
+            result.append(BaseAddressDetectionReason(reasons[i].Pointer, reasons[i].POIOffset, reasons[i].POIType))
+        core.BNFreeBaseAddressDetectionReasons(reasons)
```
plafosse, line 267: we typically put `Free` calls in a finally block just to be absolutely sure it gets freed. So you need a `try` scope around your `reasons` lifetime and the free in the finally block

plafosse (approved):
Looks good just a few minor nits

PR #5358, opened 2024-05-03
hide array implementation details

rbran:
This merge and re-split the ArrayProvider traits into `CoreArrayProvider` and `CoreArrayProviderInner`, with only the first being public.


KyleMiles (changes requested):
This looks great to me. Please rebase on top of latest dev and I'll merge it.

KyleMiles (approved):

PR #5370, opened 2024-05-06
simplify BlockContext

rbran:


rbran:
It's possible to completely remove the `BlockContext` and related generics, any `BasicBlock<T>` and `Array<BasicBlock<T>>` can be transformed into `T` and `Array<T>`, but the `Ref` struct get in the way. That can only be fully implemented with something similar to #5238.

emesare:
There is another way to remove the BlockContext by new typing each of the four basic block types. I think that will provide consumers a better experience IMO, I will try and make a PR shortly to show what I mean.

Issue #5362, opened 2024-05-05
Warnings printed when loading binary headlessly


negasora:
Tested with ELF, Mach-O, PE

```python
>>> import binaryninja
>>> for _ in range(5):
...     with binaryninja.load('/bin/true') as bv:
...             pass
... 
[0:8935060525942268291 Settings warn] Setting 'loader.architecture' deprecated by 'loader.platform'.
[0:8935060525942268291 Settings warn] Setting 'loader.architecture' deprecated by 'loader.platform'.
[0:8935060525942268291 Settings warn] Setting 'loader.architecture' deprecated by 'loader.platform'.
[0:8935060525942268291 Settings warn] Setting 'loader.architecture' deprecated by 'loader.platform'.
[0:8935060525942268291 Settings warn] Setting 'loader.architecture' deprecated by 'loader.platform'.
>>> 
```

zznop:
Fixed in build 4.1.5294

PR #5382, opened 2024-05-08
Remove loader.architecture override handling in views

zznop:
Fix for https://github.com/Vector35/binaryninja-api/issues/5362#event-12730665835

It's important to understand that this PR still allows for backwards compatibility and doesn't prevent the use of `loader.architecture` for overriding the architecture. `loader.architecture` has been deprecated for `loader.platform`, so if they use `loader.architecture`, it gets reset to `loader.platform` (which is an override that's handled by all the views). 

Example of the new behavior (if they use the deprecated `loader.architecture` setting):

```
>>> bv = load("test.exe", options={"loader.architecture" : "armv7"})
[0:6451428262981885673 Settings warn] Setting 'loader.architecture' deprecated by 'loader.platform'.
>>> bv.arch
<arch: armv7>
>>> bv.platform
<platform: armv7>
```

How we want users to do it from now on:

```
>>> bv = load("test.exe", options={"loader.platform" : "efi-x86"})
>>> bv.arch
<arch: x86>
>>> bv.platform
<platform: efi-x86>
```

### DIFF
--- a/view/elf/elfview.cpp
+++ b/view/elf/elfview.cpp
@@ -459,13 +459,20 @@ bool ElfView::Init()
 	m_simplifyTemplates = viewSettings->Get<bool>("analysis.types.templateSimplifier", this);
 
 	Ref<Settings> settings = GetLoadSettings(GetTypeName());
-	if (settings && settings->Contains("loader.imageBase") && settings->Contains("loader.architecture")) // handle overrides
+	if (settings)
 	{
-		preferredImageBase = settings->Get<uint64_t>("loader.imageBase", this);
+		if (settings->Contains("loader.imageBase"))
+			preferredImageBase = settings->Get<uint64_t>("loader.imageBase", this);
 
-		Ref<Architecture> arch = Architecture::GetByName(settings->Get<string>("loader.architecture", this));
-		if (!m_arch || (arch && (arch->GetName() != m_arch->GetName())))
-			m_arch = arch;
+		if (settings->Contains("loader.platform"))
+		{
+			Ref<Platform> platformOverride = Platform::GetByName(settings->Get<string>("loader.platform", this));
+			if (platformOverride)
+			{
+				m_plat = platformOverride;
+				m_arch = m_plat->GetArchitecture();
+			}
+		}
 	}
 
 	int64_t imageBaseAdjustment = 0;
@@ -732,13 +739,6 @@ bool ElfView::Init()
 		platform = platform->GetRelatedPlatform(entryPointArch);
 	if (!platform)
 		platform = entryPointArch->GetStandalonePlatform();
-
-	if (settings && settings->Contains("loader.platform")) // handle overrides
-	{
-		Ref<Platform> platformOverride = Platform::GetByName(settings->Get<string>("loader.platform", this));
-		if (platformOverride)
-			platform = platformOverride;
-	}
 
 	SetDefaultPlatform(platform);
 	GetParentView()->SetDefaultPlatform(platform);
@@ -2783,7 +2783,7 @@ Ref<Settings> ElfViewType::GetLoadSettingsForData(BinaryView* data)
 	Ref<Settings> settings = GetDefaultLoadSettingsForData(viewRef);
 
 	// specify default load settings that can be overridden
-	vector<string> overrides = {"loader.architecture", "loader.imageBase", "loader.platform"};
+	vector<string> overrides = {"loader.imageBase", "loader.platform"};
 	if (!viewRef->IsRelocatable())
 		settings->UpdateProperty("loader.imageBase", "message", "Note: File indicates image is not relocatable.");
 
--- a/view/macho/machoview.cpp
+++ b/view/macho/machoview.cpp
@@ -1099,13 +1099,20 @@ bool MachoView::Init()
 
 	SetOriginalBase(initialImageBase);
 	uint64_t preferredImageBase = initialImageBase;
-	if (settings && settings->Contains("loader.imageBase") && settings->Contains("loader.architecture")) // handle overrides
+	if (settings)
 	{
-		preferredImageBase = settings->Get<uint64_t>("loader.imageBase", this);
+		if (settings->Contains("loader.imageBase"))
+			preferredImageBase = settings->Get<uint64_t>("loader.imageBase", this);
 
-		Ref<Architecture> arch = Architecture::GetByName(settings->Get<string>("loader.architecture", this));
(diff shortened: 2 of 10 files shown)


bpotchik (commented):
In the API, mappedview.py is also using `loader.architecture`.

view/elf/elfview.cpp
```diff
@@ -459,15 +459,9 @@ bool ElfView::Init()
...
 
 	Ref<Settings> settings = GetLoadSettings(GetTypeName());
-	if (settings && settings->Contains("loader.imageBase") && settings->Contains("loader.architecture")) // handle overrides
-	{
+	if (settings && settings->Contains("loader.imageBase")) // handle overrides
 		preferredImageBase = settings->Get<uint64_t>("loader.imageBase", this);
 
-		Ref<Architecture> arch = Architecture::GetByName(settings->Get<string>("loader.architecture", this));
-		if (!m_arch || (arch && (arch->GetName() != m_arch->GetName())))
-			m_arch = arch;
```
bpotchik, line 468: m_arch is not being set based on the overridden platform. It is set based on the original platform, which may have changed due to an override.

view/macho/machoview.cpp
```diff
@@ -1099,15 +1099,9 @@ bool MachoView::Init()
...
 	SetOriginalBase(initialImageBase);
 	uint64_t preferredImageBase = initialImageBase;
-	if (settings && settings->Contains("loader.imageBase") && settings->Contains("loader.architecture")) // handle overrides
-	{
+	if (settings && settings->Contains("loader.imageBase")) // handle overrides
 		preferredImageBase = settings->Get<uint64_t>("loader.imageBase", this);
 
-		Ref<Architecture> arch = Architecture::GetByName(settings->Get<string>("loader.architecture", this));
-		if (!m_arch || (arch && (arch->GetName() != m_arch->GetName())))
-			m_arch = arch;
```
bpotchik, line 1108: m_arch is not being set based on the overridden platform. It is set based on the original platform, which may have changed due to an override.

view/pe/peview.cpp
```diff
@@ -597,15 +597,9 @@ bool PEView::Init()
...
 
 		settings = GetLoadSettings(GetTypeName());
-		if (settings && settings->Contains("loader.imageBase") && settings->Contains("loader.architecture")) // handle overrides
-		{
+		if (settings && settings->Contains("loader.imageBase")) // handle overrides
 			m_imageBase = settings->Get<uint64_t>("loader.imageBase", this);
 
-			Ref<Architecture> arch = Architecture::GetByName(settings->Get<string>("loader.architecture", this));
-			if (!m_arch || (arch && (arch->GetName() != m_arch->GetName())))
-				m_arch = arch;
```
bpotchik, line 606: m_arch is not being set based on the overridden platform. It is set based on the original platform, which may have changed due to an override.

view/pe/teview.cpp
```diff
@@ -227,27 +214,29 @@ bool TEView::Init()
...
-				platform = Platform::GetByName("efi-x86_64");
-			if (archName == "aarch64")
-				platform = Platform::GetByName("efi-aarch64");
-			if (!platform)
+			if (settings->Contains("loader.imageBase"))
 			{
-				m_logger->LogError("TE architecture '%s' is not supported", archName.c_str());
-				return false;
+				uint64_t baseOverride = settings->Get<uint64_t>("loader.imageBase", this);
+				if (baseOverride != header.imageBase)
```
bpotchik, line 226: This guard is unnecessary.

view/pe/teview.cpp
```diff
@@ -227,27 +214,29 @@ bool TEView::Init()
...
+				{
+					// TE image bases represent the base of the original PE (before headers are stripped) - make this
+					// adjustment for the user
+					m_imageBase = baseOverride - m_headersOffset;
```
bpotchik, line 230: There is something strange about this code. We should try to keep the definition of "loader.imageBase" consistent across view types.

Also, the comparison above `baseOverride != header.imageBase` would seem to indicate that "loader.imageBase" and header.imageBase have the same meaning. If they mean the same, and we have this: `m_imageBase = header.imageBase;`, then in the override case m_imageBase should be directly assigned baseOverride.

bpotchik: Looking elsewhere in the code, m_image base is used many times like this: `m_imageBase + m_headersOffset`. Perhaps m_imageBase should be header.imageBase - m_headersOffset?

view/md1rom/md1rom.cpp
```diff
@@ -125,42 +125,31 @@ bool Md1romView::Init()
...
 			m_plat = Platform::GetByName("nanomips");
+			m_arch = Architecture::GetByName("nanomips");
+			if (!m_arch)
+			{
+				LogWarn("nanoMIPS architecture not found. Code cannot be disassembled. If you are interested in purchasing "
+						"the nanoMIPS architecture plugin, please contact us via https://binary.ninja/support/");
+			}
 		}
 
+		if (m_arch)
```
bpotchik, line 148: Change to `if (m_arch && m_plat)` ?

view/pe/teview.cpp
zznop: This code is strange because it's accounting for the hell that is the TE file format. See [here](https://uefi.org/specs/PI/1.8/V1_TE_Image.html) for more information.

TL;DR TE's are frankenstein'd PEs. The toolchain starts by building a PE file, then it strips off the MS-DOS, COFF, optional headers, and section headers and replaces them with TE headers (which are smaller in size). The top level TE header contains a "strippedSize" member that tells you how much data was stripped from the original PE. `m_imageBase` represents the base of the original PE image (before it was stripped).

From the spec:

> The StrippedSize should be set to the number of bytes removed from the start of the original image, which will typically include the MS-DOS, COFF, and optional headers, as well as the section headers. This size can be used by image loaders and tools to make appropriate adjustments to the other fields in the TE image header. Note that StrippedSize does not take into account the size of the TE image header that will be added to the image. That is to say, the delta in the total image size when converted to TE is StrippedSize - sizeof (EFI_TE_IMAGE_HEADER). This will typically need to be taken into account by tools using the fields in the TE header.

zznop: For even more context, here are some examples of other loaders that need to make the same adjustments:

https://github.com/mkrasnitski/teloader/blob/master/teloader.py#L68
https://github.com/binarly-io/efiXplorer/blob/1f268587cdd378cd689ae256691434c7a305cc97/efiXloader/pe_ida.h#L1062

PR #5396, opened 2024-05-10
Fix invalid write to uninitialized Vec

rbran:


rust/src/binaryview.rs
```diff
@@ -225,45 +225,23 @@ pub trait BinaryViewExt: BinaryViewBase {
...
-
-            {
-                let dest_slice = ret.get_unchecked_mut(0..len);
-                res = self.read(dest_slice, offset);
-            }
-
-            ret.set_len(res);
-        }
+        let slice = ret.as_mut_slice();
+        let size = self.read(slice, offset);
```
mkrasnitski, line 231:
```suggestion
        let size = self.read(&mut ret, offset);
```
Deref coercion allows to pass `ret` directly here. Also, the extra empty lines could probably be removed.

rust/src/binaryview.rs
```diff
@@ -225,45 +225,23 @@ pub trait BinaryViewExt: BinaryViewBase {
...
-
-            if res > 0 {
-                dest.set_len(starting_len + res);
-            }
-
-            res
-        }
+        dest.resize(starting_len + len, 0);
+        let slice = &mut dest[starting_len..];
+        let read_size = self.read(slice, offset);
```
mkrasnitski, line 242:
```suggestion
        let read_size = self.read(&mut dest[starting_len..], offset);
```

KyleMiles (approved):

KyleMiles:
The new suggestions thing is really cool. Thank you both @rbran and @mkrasnitski.

PR #5420, opened 2024-05-14
neon_intrinsics.cpp: add some unimplemented f32 scalar encodings

comex:
Add f32 cases mirroring the f64 ones.

Caveat: This is far from comprehensive.  It's just a set of instructions I happened to bump into in the binary I'm reverse engineering.

Also, it would be nice if FNEG generated a standard negation instruction instead of an intrinsic, but I didn't try to implement that.


galenbwill:
Closing PR because its changes are incorporated into 68b7ecba

PR #5422, opened 2024-05-15
Allow Rust CoreAPI FFI to implement Default

rbran:
This allow writing simpler and cleaner code when initializing FFI types, avoiding the need of `core::mem::zeroed()`.

It's specially useful to partially initialize structs:
```rust
BNHighlightColor {
  style: BNHighlightColorStyle::StandardHighlightColor,
  color: BNHighlightStandardColor::NoHighlightColor,
  alpha,
  ..Default::default()
}
```

### DIFF
--- a/rust/binaryninjacore-sys/build.rs
+++ b/rust/binaryninjacore-sys/build.rs
@@ -96,6 +96,7 @@ fn main() {
         .clang_arg("c++")
         .size_t_is_usize(true)
         .generate_comments(false)
+        .derive_default(true)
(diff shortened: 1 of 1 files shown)


KyleMiles (approved):

KyleMiles:
Please add an example of this somewhere, if it makes sense in any of our existing examples/docs.

rbran:
It's not part of the public API, because right now the Core API FFI is not exposed. It's mostly to avoid unnecessary unsafe blocks on the API implementation.

PR #5431, opened 2024-05-17
Implement wrappers around the Architecture ID types

rbran:
Instead of using naked ids (u32) for types like: Register, Flag, Intrinsic, etc. This PR replaces those types with wrappers like: RegisterId, FlagId, IntrinsicId, etc.

The objective is to provide a more clear API for the user, once the type RegisterId can't be confused with any other u32.

This pattern also allow the creation of the `ArchitectureExt::get_field` that can receive any kind of id an return the corresponding type, eg:

```rust
let reg_id = get_reg_id();
let flag_id = get_flag_id();
let instrinsic_id = get_intrinsic_id();
let reg: CoreRegister = arch.get_field(reg_id).unwrap();
let flag: CoreFlag = arch.get_field(flag_id).unwrap();
let intrinsic: CoreIntrinsic = arch.get_field(instrinsic_id).unwrap();
```

### DIFF
--- a/rust/examples/pdb-ng/src/symbol_parser.rs
+++ b/rust/examples/pdb-ng/src/symbol_parser.rs
@@ -2010,13 +2010,13 @@ impl<'a, S: Source<'a> + 'a> PDBParserInstance<'a, S> {
                 self.log(|| format!("Register {:?} ==> {:?}", reg, xreg));
                 self.arch
                     .register_by_name(xreg.to_string().to_lowercase())
-                    .map(|reg| reg.id() as i64)
+                    .map(|reg| reg.id().0 as i64)
             }
             Some(AMD64(areg)) => {
                 self.log(|| format!("Register {:?} ==> {:?}", reg, areg));
                 self.arch
                     .register_by_name(areg.to_string().to_lowercase())
-                    .map(|reg| reg.id() as i64)
+                    .map(|reg| reg.id().0 as i64)
             }
             // TODO: Other arches
             _ => None,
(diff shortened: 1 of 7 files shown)


rust/examples/pdb-ng/src/symbol_parser.rs
```diff
@@ -2010,13 +2010,13 @@ impl<'a, S: Source<'a> + 'a> PDBParserInstance<'a, S> {
                 self.log(|| format!("Register {:?} ==> {:?}", reg, xreg));
                 self.arch
                     .register_by_name(xreg.to_string().to_lowercase())
-                    .map(|reg| reg.id() as i64)
+                    .map(|reg| reg.id().0 as i64)
```
emesare, line 2013: `id().into()`?

rbran: Maybe, but not using `Into`. Maybe something like `reg.id().into_inner()` or `reg.raw_id()`.

rbran: The `as i64` can be replaced with `.into()`. But the `.0` can only be replaced with a appropriated named function.

emesare:
This is available with https://github.com/Vector35/binaryninja-api/commit/ace9ba8c641689a69f974060fa20189ea7963d93

However I didn't add that field helper, I find that a little useless, maybe I misunderstand the purpose but it seems unnecessary. Please correct me if I'm wrong.

rbran:
Do you mean the `ArchitectureFromId::into_value`? It was mostly to allow the creation of the function `Architecture::field`, that allow the user to get any kind of field using a single function.

emesare:
This PR was merged with the rust refactor, https://github.com/Vector35/binaryninja-api/commit/6a63c17853009ce7a0071458e39c140a09440230. Thank you @rbran!

PR #5435, opened 2024-05-18
Rust: Allow running doctests and unit tests

mkrasnitski:
Adding the `cargo::rustc-link-arg` option to the build.rs file of the `binaryninja` crate allows doctests and unit tests to be linked against `binaryninjacore.so.1` and ran when invoking `cargo test`.

Also, the single-colon `cargo:` syntax is deprecated as of Rust 1.77, and future features will all use the `cargo::` syntax. Since the MSRV is already Rust 1.77, it's fine to switch for consistency's sake going forward.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,6 +37,10 @@ jobs:
         working-directory: ./rust
         run: cargo test --doc -- --show-output
 
+      - name: cargo unit test
+        working-directory: ./rust
+        run: cargo test --no-run
+
       - name: cargo clippy
         working-directory: ./rust
         run: cargo clippy -- -D warnings
--- a/rust/binaryninjacore-sys/build.rs
+++ b/rust/binaryninjacore-sys/build.rs
@@ -44,7 +44,7 @@ fn link_path() -> PathBuf {
 }
 
 fn main() {
-    println!("cargo:rerun-if-changed=../../binaryninjacore.h");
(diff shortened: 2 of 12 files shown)


mkrasnitski:
I guess CI is failing because binja isn't actually installed as part of the GH action? I wouldn't know how to set it up to allow for the linking to succeed - as far as I know the `build.rs` changes are correct (works on my machine lol)

emesare:
> I guess CI is failing because binja isn't actually installed as part of the GH action?

You are correct we do not install binja on GH actions, we only use it for PR checks.  I think it would be fine to just not run tests in CI.

mkrasnitski:
What's odd is that the doctests compile fine, whereas `cargo test --no-run` fails to even compile. What could be the difference here?

emesare:
> What's odd is that the doctests compile fine, whereas `cargo test --no-run` fails to even compile. What could be the difference here?

The doctests are all annotated with `no_run`

mkrasnitski:
Exactly: they type check and compile just fine but don't run. For some reason though, the unit tests don't even though `--no-run` is passed.

emesare:
This PR is superseded by [rust_break_everything](https://github.com/Vector35/binaryninja-api/tree/rust_break_everything). 

Specifically https://github.com/Vector35/binaryninja-api/commit/8a24a4394cf887728da1f260d7d3a86fb58157aa (Warning: this branch will rewrite history, these commits may not be accurate later)

PR #5439, opened 2024-05-22
fix double-free with `Array<InstructionTextToken>`

rbran:


emesare:
Added in https://github.com/Vector35/binaryninja-api/commit/16b31d9bac4dc3c41799f644b18bea433e4e9460, thanks!

PR #5476, opened 2024-05-30
Arm64 lift scvtf

galenbwill:
Lifts the scalar variants of scvtf


galenbwill:
Squashed and merged manually.

Issue #3092, opened 2022-04-27
function recognizer fails to work on got/plt linux-aarch64


psifertex:
Simple hello-world, prtinf type library is not being appropriately applied.

[hello.zip](https://github.com/Vector35/binaryninja-api/files/8572687/hello.zip)

PR #5479, opened 2024-05-31
Fixed type propagation issue in AARCH64 and MIPS

river-li:
Previously, prtinf type library is not being appropriately applied.

The original issue can be found here: #3092

Before
image

After
image


plafosse (approved):
Code looks good. Though you could make it a little bit more efficient by passing in the `DEFAULT_EXTERNAL_NAMESPACE` as the second parameter.

PR #5481, opened 2024-05-31
Add new setting to skip HEAD requests for localhost servers

joshwatson:
Proxy servers such as [SymProxyCloud](https://github.com/microsoft/SymProxyCloud) don't respond to `HEAD` requests, so this PR adds a new setting that allows the user to skip file existence checks using a `HEAD` request.

Note that this setting only affects `localhost` servers, not other remote servers.

### DIFF
--- a/python/settings.py
+++ b/python/settings.py
@@ -65,6 +65,7 @@ class Settings:
 		"title"               string                                   None                 No         Concise Setting Title
 		"type"                string                                   None                 No         {"array", "boolean", "number", "string"}
 		"elementType"         string                                   "type" is "array"    No         {"string"}
+		"sorted"              boolean                                  "type" is "array"    Yes        Automatically sort list items (default is true)
 		"enum"                array : {string}                         "type" is "string"   Yes        Enumeration definitions
 		"enumDescriptions"    array : {string}                         "type" is "string"   Yes        Enumeration descriptions that match "enum" array
 		"minValue"            number                                   "type" is "number"   Yes        Specify 0 to infer unsigned (default is signed)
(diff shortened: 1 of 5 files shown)


CouleeApps (changes requested):
This seems to break support for `file.ptr` redirection (and would break `.pd_` compression if it were supported) because the symbol search thinks it can resolve the pdb file, but gets a 404 and fails back to the next symbol server instead of trying the next option in the current server. Resolving this might be a bit tricky (I can't believe Rust still doesn't have generators) since handling `file.ptr` involves doing two requests and some parsing. So some redirection of control flow will be needed to get this to work properly, likely moving the call to `read_from_sym_store` out of `load_from_file` and maybe dropping `sym_store_exists` as a concept and just loading it all in one shot. I can implement this if you want, or feel free to do it yourself.

joshwatson:
Not handling `file.ptr` and `foo.pd_` was an explicit choice on my part, because I don't know of a symbol proxy that runs as localhost that will serve those. Further, if there were a localhost server that handled those, my guess would also be that it would handle `HEAD` requests properly (and hence why I gated the behavior behind a new setting).

However, if that logic is a requirement for you, I'm perfectly fine with you adding it; I just don't have an exemplar localhost proxy server to test further changes, haha

CouleeApps:
I'd really prefer to not stealth break functionality if possible, given I have no idea if people are relying on this. I've been testing with `python -m http.server` which is perfectly happy to serve whatever files I throw in it, so it could certainly be being used by someone.

CouleeApps:
Just pushed an update for this that does the implementation I described above. Let me know if it works for the same servers that your commit did, and if so I'm just going to merge that.

joshwatson:
Yep, it still seems to work, thank you!

CouleeApps:
Merged that commit to dev, so closing this now

PR #5483, opened 2024-05-31
implement Project, ProjectFile and ProjectFolder in rust

rbran:
I assumed that `BNProjectFileGetCreationTimestamp` and other time related fields are seconds since the UNIX_EPOCH.

I also notice that the CPP implementation is missing the functions `BNFreeProject`, `BNFreeProjectFile` and `BNFreeProjectFolder`, although it include the implementation for each of these objs. Maybe it's missing the deconstructor and is leaking memory.

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -152,6 +152,7 @@ pub mod logger;
 pub mod metadata;
 pub mod mlil;
 pub mod platform;
+pub mod project;
 pub mod rc;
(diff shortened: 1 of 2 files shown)


rbran:
NOTE TO SELF: Once merged, remember to rebase and PR rbran/impl-extlib

negasora:
Manually rebased and pushed these commits, ty!

PR #5538, opened 2024-06-10
implement rust workflow

rbran:


rust/src/workflow.rs
```diff
@@ -0,0 +1,529 @@
...
+        let result = unsafe {
+            BNWorkflowRegisterActivity(
+                self.as_raw(),
+                activify.as_raw(),
+                subactivities_ptr.as_mut_ptr(),
+                subactivities_ptr.len(),
+            )
+        };
+        unsafe { Activity::from_raw(ptr::NonNull::new(result).unwrap()) }
+    }
```
emesare, line 311: This does not work, given:
```rs
    let work = Workflow::new("testWorkflow");
    let activity = Activity::new_with_action("test", |ctx| {});
    work.register_activity::<Vec<String>>(&activity, vec![]);
    work.register(
        r#"{ 
            "title" : "Rust workflow",
            "description" : "",
            "capabilities" : []
        }"#,
    )
    .expect("workflow is registered");
    work.insert("core.function.translateTailCalls", ["test"]);
```
I don't know *why* but it fails in `Workflow::Validate` but it shouldn't fail if there is no subactivity.

rust/src/workflow.rs
```diff
@@ -0,0 +1,529 @@
...
+        } else {
+            Err(())
+        }
+    }
+
+    /// Register an Activity with this Workflow.
+    ///
+    /// * `activity` - the Activity to register
+    /// * `subactivities` - the list of Activities to assign
+    pub fn register_activity<I>(&self, activify: &Activity, subactivities: I) -> Activity
```
emesare, line 291: typo: activify -> activity

emesare:
Added with https://github.com/Vector35/binaryninja-api/commit/3a947674819c43e8de9902242a1b1748943bb025, thanks!

PR #5611, opened 2024-06-14
implement rust Component

rbran:


negasora:
Manually rebased and pushed these commits, ty!

PR #5634, opened 2024-06-19
Implement Clone for Expression

emesare:
This is the rust equivalent to python's [LowLevelILFunction.copy_expr](https://dev-api.binary.ninja/binaryninja.lowlevelil-module.html#binaryninja.lowlevelil.LowLevelILFunction.copy_expr). This is very useful when replacing expressions with the original expression being a child expression, as replacing the expression without recreating the original expression will lead to the child expression sharing the same expression index, causing any visitor to infinitely loop.

This allows you to `clone` the expression which effectively creates a new expression index for which you can slot into the replaced expression.

### DIFF
--- a/rust/src/llil/expression.rs
+++ b/rust/src/llil/expression.rs
@@ -83,6 +83,36 @@ where
     }
 }
 
+impl<'func, A, V, R> Clone for Expression<'func, A, Mutable, NonSSA<V>, R>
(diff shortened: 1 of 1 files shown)


emesare:
Closing this in favor of https://github.com/Vector35/binaryninja-api/pull/5641

PR #5654, opened 2024-06-23
Rust: remove direct dependency on the `libc` crate

mkrasnitski:
Almost all uses of the `libc` crate were to import types that the standard library already provides, e.g. `c_void`, `c_char`, etc. We can import those from `std::ffi`. In some places they were being imported from `std::os::raw`, so I changed those imports as well, as they are just type aliases for the definitions in `std::ffi` anyways.

One nontrivial case where `libc` was being used was inside `cb_{free_}flag_conditions_for_semantic_flag_group`, which used `libc::{malloc, free}`. This was changed to use `std::alloc::{alloc, dealloc}` and required some hacky layout calculations in order to smuggle the length of the allocated array across the FFI boundary. 

Getting rid of this hack would require for the Core API to change such that a `count: usize` parameter is passed to `cb_free_flag_conditions_for_semantic_flag_group` (and ideally also to `cb_free_register_list`) so that no smuggling would be required and we could just call `Box::from_raw(ptr::slice_from_raw_parts_mut(conds, count))` and be done. I'll open a separate issue for that.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -102,7 +102,6 @@ version = "0.1.0"
 dependencies = [
  "binaryninjacore-sys",
  "lazy_static",
- "libc",
  "log",
  "rayon",
 ]
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -10,7 +10,6 @@ noexports = []
 [dependencies]
 lazy_static = "1.4.0"
 log = "0.4"
-libc = "0.2"
 rayon = { version = "1.8", optional = true }
 binaryninjacore-sys = { path = "binaryninjacore-sys" }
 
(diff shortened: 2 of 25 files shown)


mkrasnitski:
Rebased on the newest `dev`, and ready for re-review.

mkrasnitski:
Before I resolve the new conflicts, I'd love for #5764 to be merged so I could rebase on top.

emesare:
This PR is superseded by [rust_break_everything](https://github.com/Vector35/binaryninja-api/tree/rust_break_everything). 

Specifically https://github.com/Vector35/binaryninja-api/commit/ec43516f6c9eb0482e13626bb205c149d75d1bf4 and https://github.com/Vector35/binaryninja-api/commit/ce5995c3b8c8765811521d9d9f0eadcf3b601415 (Warning: this branch will rewrite history, these commits may not be accurate later)

PR #5665, opened 2024-06-25
Add entry functions to triage UI

river-li:
Added a new group in the triage UI.

image


emesare:
Screenshot 2024-06-25 at 8 20 44 PM

IMO the layout should be something like above, or have it in a tab group with exports.

Screenshot 2024-06-25 at 8 20 44 PM

emesare:
Merged with the change described above:

image

Should be available on `4.3.6538-dev` or later.

Thank you!

PR #5671, opened 2024-06-27
implement rust MainThread

rbran:


rust/src/mainthread.rs
```diff
@@ -0,0 +1,204 @@
...
+//!
+//! The default maximum number of concurrent worker threads is controlled by the
+//! `analysis.limits.workerThreadCount` setting but can be adjusted at runtime via
+//! [set_worker_thread_count].
+//!
+//! The worker threads are native threads, managed by the Binary Ninja core. If
+//! more control over the thread is required, consider using the
+//! [crate::backgroundtask::BackgroundTask] class.
+
+use core::{ffi, mem, ptr};
```
mkrasnitski, line 32: Since there's no realistic chance of ever being `#![no_std]`, these should be imported from `std` instead. Also, types like `NonNull`, `CStr`, `c_char`, etc. should probably be imported here as well to avoid extra `::` when using those types (they're used a lot in this file).

rust/src/mainthread.rs
```diff
@@ -0,0 +1,204 @@
...
+
+impl MainThreadAction {
+    pub(crate) unsafe fn from_raw(handle: ptr::NonNull<BNMainThreadAction>) -> Self {
+        Self { handle }
+    }
+
+    pub(crate) unsafe fn ref_from_raw(handle: &*mut BNMainThreadAction) -> &Self {
+        assert!(!handle.is_null());
+        mem::transmute(handle)
+    }
```
mkrasnitski, line 64: Is this function actually used anywhere? And since `*mut T: Copy`, you could reasonably just call `from_raw` instead and get an owned `Self`. Plus, elsewhere in the codebase, the return type of `ref_from_raw` is usually `Ref<Self>`. I also don't like the use of `transmute` here.

rust/src/mainthread.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    #[allow(clippy::mut_from_ref)]
+    pub(crate) unsafe fn as_raw(&self) -> &mut BNMainThreadAction {
+        &mut *self.handle.as_ptr()
+    }
```
mkrasnitski, line 69:
```suggestion
    pub(crate) fn as_raw(&self) -> *mut BNMainThreadAction {
        self.handle.as_ptr()
    }
```

This function isn't `pub` and passing the raw `*mut BNMainThreadAction` across FFI already requires unsafe, so this function should just be a wrapper around `NonNull::as_ptr`. As far as I can tell, every use of this function coerces the return type from `&mut` to `*mut`.

rust/src/mainthread.rs
```diff
@@ -0,0 +1,204 @@
...
+
+use core::{ffi, mem, ptr};
+
+use binaryninjacore_sys::*;
+
+use crate::string::BnStrCompatible;
+
+#[repr(transparent)]
+pub struct MainThreadAction {
+    handle: ptr::NonNull<BNMainThreadAction>,
```
mkrasnitski, line 40: Reading the documentation for `NonNull` mentions that it's covariant as opposed to `*mut T`. This worries me a little bit especially because much of the Binja Rust bindings offer some amount of interior mutability by having methods take `&self` while calling C functions that mutate the underlying data. I'm unsure of the soundness here, but I think it's probably fine because the only data held on the Rust side is the raw handle.

mkrasnitski (changes requested):
Much of my review here applies to most of the PRs you've opened recently as well, as they follow the same patterns of implementation.

rust/src/mainthread.rs
rbran: This function is used by the `cb_add_action`. The copy is intentionally avoided here, to avoid problems like #5496

rust/src/mainthread.rs
rbran: Similar to above, the `*mut T: Copy` is avoided and lifetime is enforced to avoid citations like #5496.

Although the pointer is usually immediately converted into a pointer, voiding the effort of enforcing the lifetime. In any case, this is a internal function and it could be fixed without changing the API.

rust/src/mainthread.rs
mkrasnitski: In that case, a better alternative to raw transmute is something like:
```rust
unsafe { &*(handle as *const _ as *const Self) }
```
It's more clear what's going on here and it doesn't tap into transmute's full power (which is a good thing)

emesare:
This PR was merged with the rust refactor, https://github.com/Vector35/binaryninja-api/commit/6a63c17853009ce7a0071458e39c140a09440230. Thank you @rbran!

PR #5712, opened 2024-07-09
Remove sysv calling convention

river-li:
Adding `sysv` calling convention introduces a lot of false positives during analysis. Defined in the UEFI Specification, X64 follows the same calling convention as `win64` calling convention.

https://uefi.org/specs/UEFI/2.9_A/02_Overview.html#detailed-calling-conventions

If still possible for users to set the calling convention to `sysv` in function properties widget.


zznop:
LGTM - I'm good with this change as long as the user still has the ability to override CC to be `sysv` if needed.

PR #5764, opened 2024-07-18
Fix clippy warnings and run rustfmt

mkrasnitski:
Clippy was complaning about some uses of `std::mem::transmute`, so I changed them to use simple `as` casting instead. Additionally, ran `cargo fmt --all` to avoid extraneous formatting changes from leaking into future PRs.


emesare:
Due to the nature of formatting all of the rust api, we are instead going to format the code in [`rust_break_everything`](https://github.com/Vector35/binaryninja-api/tree/rust_break_everything). Those refactors also will create large diffs, and doing them both at the same time is better than if we were to go ahead and format now, then follow it shortly after with another large refactor.

I am going to keep this PR up until [`rust_break_everything`](https://github.com/Vector35/binaryninja-api/tree/rust_break_everything) is merged.

PR #5791, opened 2024-07-25
Update Cargo.lock

mkrasnitski:
This PR updates dependencies in `Cargo.lock` in order to fix compilation with Rust 1.80 (see time-rs/time#693). in doing so, `uuid` v1.8.0 introduced a change which broke type inference in a few places in `pdb-ng`, so those places are changed to accomodate. 

Also, deleted every other `Cargo.lock` file because they weren't actually being used. They were located inside workspace member directories, whose dependencies are tracked in the main project's `Cargo.lock` and so served no use. Finally, changed the patched `pdb` dependency from a git submodule to a cargo git-dependency.


emesare:
We are actually going to move away from `Cargo.lock` entirely as its purpose is mute for any dependency (which is what the sys and wrapper crate both are), other than for tests and what not. Keeping it up-to-date is a chore and we should instead run ci on the latest possible version of a cargo dependency (that way we can catch things like `uuid` making a breaking change).

For anyone interested the changes will land on https://github.com/Vector35/binaryninja-api/tree/rust_break_everything for testing

PR #5810, opened 2024-07-31
Fix some reported issues with MIPS LLIL

cryptwhoa:
This addresses some issues reported in the general updates to MIPS.

The original goal was to make sure that 64-bit shifts still work when the architecture uses 32-bit addresses, and add LLIL for some requested instructions.  During the course of this, I found that most places that set the HI and LO registers weren't sign extending properly for 32-bit operations in 64-bit architectures, so I went ahead and fixed those too.

The binary causing issues with 64-bit shifts not being lifted is something I don't have access to, so I wasn't able to test if that issue is fixed.  Several of the system intrinsics also aren't used in any of my sample binaries, so they haven't been tested, but are generally simple.  I was able to test the new TLB lifting and 64-bit sign extension.


fuzyll:
These changes look great, thanks! Looks like they went through our unit tests okay and, from what I was able to test, the output looks much improved. I guess the way I did the merge via the CLI didn't manage to sync up with this PR on the web, but rest assured your commits are currently making their way through our CI right now.

Sorry it took a little bit for me to get to, I was unexpectedly out of the office last week.

PR #5831, opened 2024-08-06
Allow the user to modify the BV during initialization

rbran:
If the method `CustomBinaryView::init` don't accept `&mut self` the user will have difficulty to incorporate `args: Self::Args` into it's data structure.


emesare:
LGTM

PR #5858, opened 2024-08-17
doc: Add `BackgroundTaskThread` python example

riskydissonance:
To allow python plugins to do work not on the UI thread. 

Also allows them to call `update_analysis_and_wait()`.

Tested with a basic plugin.json as a full plugin and works.


CouleeApps:
Nice basic example and works as expected. Thanks.

PR #5868, opened 2024-08-21
Add msp430 architecture

jrozner:


fuzyll:
In the process of merging this. There are a bunch of changes that were required to get this to play nicely with CI and load at the appropriate time as a core plugin, rather than user plugin, but that wasn't bad.

The main problem I'm now working through is that there are a number of crashes while loading even just the [code examples from TI](https://www.ti.com/tool/MSP430-FUNCTION-CODE-EXAMPLES#downloads). One of them is actually in DWARF debug info parsing, which I now have fixed and is our fault. But, there are a number of emitted warnings and a crash on `one_operand!` not handling `Operand::Constant` at all that I am still working through.

jrozner:
I can take a look at it a bit later tonight or tomorrow probably. The disassembly should be very accurate and is well tested. The lifting is less so. Unless there are edge cases that are being missed, fixing the macros should be fairly straight forward and just fix the lifting for all relevant instructions. Most of this was tested on toy firmwares I wrote or micro corruption levels. So there maybe parts that were not exercised extensively.

fuzyll:
I think we might be okay. Your constant generator decoding appears to be accurate, so I just added a `Operand::Constant(val) => $il.store(2, $il.const_int(2, *val as u64), $op).append(),` in the macro and we're cooking. Have a few other things to track down, but as long as we're not crashing, I'm probably close to shipping.

fuzyll:
With some additional fixes, this has now been merged. Unfortunately, a different issue is preventing it from shipping out to customers at the moment. It should be included in all editions except Free starting in builds >= `4.2.5961-dev`. I've also added a test case in our internal unit tests for it.

Thanks!

PR #5880, opened 2024-08-24
Simplify `Guard` API and reduce its use in `CoreArrayProvider`

mkrasnitski:
The value of the lifetime parameter `'a` in `Guard<'a>` can be inferred via type checking and doesn't actually need to be directly linked to an `owner` parameter. Additionally, lots of types whose `Wrapped<'a>` associated type was a `Guard<'a, Self>` can instead be made `#[repr(transparent)]` and their wrapped type changed to `&'a Self`.

A few types still require using `Guard` because they aren't able to be made `#[repr(transparent)]`, namely:
* `BasicBlock`
* `CallingConvention`
* `LinearDisassemblyLine`
* `CodeReference`
* `Array<InstructionToken>`

Also, a few types which weren't able to be marked `#[repr(transparent)]` were changed to return an owned `Self` instead of a `Guard`:
* `NamedTypedVariable`
* `StructureMember`


emesare:
LGTM, committed with https://github.com/Vector35/binaryninja-api/commit/53466cdaba3f6decb6226b7e2abc5195e7e71d76

PR #5884, opened 2024-08-26
implement rust idb_import

rbran:
This plugin allow binja to import `IDB` (`*.idb` and `*.i64`) and `TIL` (`*.til`) files data types using the `Import Debug Info from External File` option.


emesare:
Merged with https://github.com/Vector35/binaryninja-api/commit/fe0d4ae2c3baa26fca84a9a8cb9670955bdfea12, on dev with `>4.2.6013-dev`

PR #5897, opened 2024-09-02
Replace `BnStrCompatible` with `AsCStr` trait

mkrasnitski:
This is a rather large PR, but the essence of the changes is in `string.rs`. Replaces the `BnStrCompatible` trait with an `AsCStr` trait with the following definition:

```rust
pub trait AsCStr {
    fn as_cstr(&self) -> Cow<'_, CStr>;
}
```

Note that this trait borrows `&self` instead of consuming `self`. This simplifies the list of implementing types and reduces the need for cloning. The `as_cstr` method returns a `Cow` in order to provide implementations both for already null-terminated strings (`BnString`) and for non-terminated strings (`&str`/`String`).

Note that it is still possible in some cases to create a pointer to a temporary that gets immediately dropped if calling `s.as_cstr().as_ptr()` inline, same as with `BnStrCompatible`. I was careful to avoid doing so when swapping out the traits.

### DIFF
--- a/plugins/warp/src/matcher.rs
+++ b/plugins/warp/src/matcher.rs
@@ -506,7 +506,7 @@ pub struct PlatformID(u64);
 impl From<&Platform> for PlatformID {
     fn from(value: &Platform) -> Self {
         let mut hasher = DefaultHasher::new();
-        hasher.write(value.name().to_bytes());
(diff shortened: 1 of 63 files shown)


mkrasnitski:
Closing as this has been superseded.

PR #5898, opened 2024-09-03
Replace `lazy_static` crate with `std::sync::LazyLock`

mkrasnitski:
The `LazyLock` type was stabilized in Rust 1.80, and obsoletes the `lazy_static` crate for lazy initialization.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -122,7 +122,6 @@ name = "binaryninja"
 version = "0.1.0"
 dependencies = [
  "binaryninjacore-sys",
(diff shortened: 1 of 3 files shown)


emesare:
Removed it with https://github.com/Vector35/binaryninja-api/commit/1b6b9a3e10dc4640621c29df9173bbdc0fc5a8f6, I also just omitted that in favor of the Display impl so there is no need to create an empty BinaryView.

Blocking on https://github.com/Vector35/binaryninja-api/tree/rust_cleanup_0

PR #5909, opened 2024-09-07
Respect ELF endianness when calculating AArch64 relocations

### LINKED ISSUE #5366: aarch64 Big-Endian data mode doesn't handle relocations in big endian
- Binary Ninja Version: 4.1.5213-dev
- OS: macOS
- OS Version: 14
- CPU Architecture: M1

**Bug Description:**
When opening an aarch64/big endian data mode file, while the lifter totally handles the big endian loads correctly, the relocations are written as little endian and appear backwards as a result.

**Steps To Reproduce:**
1. Open this aarch64be file from dogbolt: [fe62aea37a48c3836cdf5c8431ba155988f375f5cdac062f36d4e6da66ff3165.zip](https://github.com/Vector35/binaryninja-api/files/15217257/fe62aea37a48c3836cdf5c8431ba155988f375f5cdac062f36d4e6da66ff3165.zip)
2. Navigate to 0x410df0
3. Observe all the addresses are byte swapped
4. Check the bytes in hex view (ELF) vs hex view (Raw) and see that the ELF bytes are swapped

**Expected Behavior:**
I expected the relocations to be applied correctly.

**Screenshots:**
image

**Additional Information:**
Apparently aarch64be is totally a file format, it's only big endian for data accesses, not instructions. Very cursed.

mkrasnitski:
Fixes #5366

Queries the view endianness when calculating relocation targets, at least in the case where the target is a simple 16, 32, or 64-bit integer. Some relocation targets are 32-bits wide but use bitfields, and I'm unsure how to respect endianness there cleanly, or if it even matters. Feedback welcome.

### DIFF
--- a/arch/arm64/arch_arm64.cpp
+++ b/arch/arm64/arch_arm64.cpp
@@ -3021,11 +3021,18 @@ class Arm64ElfRelocationHandler : public RelocationHandler
 		auto info = reloc->GetInfo();
 		if (len < info.size)
 			return false;
-		uint64_t* dest64 = (uint64_t*)dest;
-		uint32_t* dest32 = (uint32_t*)dest;
-		uint16_t* dest16 = (uint16_t*)dest;
-		// auto swap = [&arch](uint32_t x) { return (arch->GetEndianness() == LittleEndian)? x :
-		// bswap32(x); };
+
+		BNEndianness endianness = view->GetDefaultEndianness();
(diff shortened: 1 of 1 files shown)


emesare:
Had to swap out the byte swap for a generic version (byteswap.h is compiler specific)

Other than that LGTM!

Before:
Screenshot 2024-09-08 at 5 16 20 PM
After:
Screenshot 2024-09-08 at 5 16 05 PM

Issue #5876, opened 2024-08-23
Need to handle R_MIPS_TLS_* MIPS64 relocations properly


fuzyll:
**What is the feature you'd like to have?**
Currently, in MIPS64, we don't handle MIPS64-specific relocations properly. They appear to be non-standard and are built out of three separate relocations (see the comment and structs [here in the gdb source](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=include/elf/mips.h;h=fb3357666ec93483eee0fa12d7e8cb3211dc490a;hb=HEAD#l938)). If I'm understanding this right, a relocation of 4611 should be something like `(R_MIPS_64 << 8) | R_MIPS_REL32 | 3 = 4611`

**Is your feature request related to a problem?**
Binaries like this `libc` from a router that uses a Cavium OCTEON processor ([libc-2.9.so.zip](https://github.com/user-attachments/files/16733714/libc-2.9.so.zip)) report multiple relocations that are not recognized. It's unclear to me at this point how impactful these are on analysis, but the fact that we don't handle _any_ of the MIPS64 relocations is likely problematic in at least some limited scenarios.

**UPDATE**
As of 4.2 stable release, we now handle all the relocation types mentioned in this issue except for the Thread Local Storage-related relocations (`R_MIPS_TLS_*`).

zznop:
[This PR](https://github.com/Vector35/binaryninja-api/pull/5919) successfully handles the 4611 relocation type. It has been added in builds >= `4.2.6325-dev`. I have tested with the binary linked to this issue and it reduces the number of unhandled relocations from 1000+ to only 12. See below:

```
Unsupported relocation type: 48 (R_MIPS_TLS_TPREL64) @0x189640
Unsupported relocation type: 48 (R_MIPS_TLS_TPREL64) @0x189648
Unsupported relocation type: 48 (R_MIPS_TLS_TPREL64) @0x189650
Unsupported relocation type: 48 (R_MIPS_TLS_TPREL64) @0x189658
Unsupported relocation type: 48 (R_MIPS_TLS_TPREL64) @0x189660
Unsupported relocation type: 48 (R_MIPS_TLS_TPREL64) @0x189668
Unsupported relocation type: 40 (R_MIPS_TLS_DTPMOD64) @0x189670
Unsupported relocation type: 48 (R_MIPS_TLS_TPREL64) @0x189680
[...]

PR #5919, opened 2024-09-10
Fixing various MIPS relocation issues

Whurbin:
Fixes various MIPS relocation issues.

1. Implicit addends in MIPS do not seem to be honored. Adding the addend to the target address to fix relocation target.

2. It is supposed to be the case that R_MIPS_HI16 relocations are followed by one or more R_MIPS_LO16 relocations. This is not always the case when certain compiler optimizations are enabled. Removing this requirement and fulfilling the relocation regardless of the preceding relocation.

3. Adding support for the following relocations:
- R_MIPS_HIGHEST
- R_MIPS_HIGHER
- R_MIPS_GPREL16
- R_MIPS_GPREL32
- (R_MIPS_64 << 8) | R_MIPS_REL32 (i.e. 0x1203)


Whurbin:
This patch should fix the following issue(s):
https://github.com/Vector35/binaryninja-api/issues/5876

plafosse:
Thanks for the PR! Do you happen to have a binary which contains these relocations that you can share?

zznop:
This PR has been accepted and included in builds >= `4.2.6325-dev`. Thank you for the contribution.

zznop:
@Whurbin - this PR caused a regression in handling `LO16` and `HI16` relocation types with certain binaries. I am going to revert back to the old code for those relocation types for now. Do you have a binary that we can use to investigate this behavior:

> It is supposed to be the case that R_MIPS_HI16 relocations are followed by one or more R_MIPS_LO16 relocations. This is not always the case when certain compiler optimizations are enabled. Removing this requirement and fulfilling the relocation regardless of the preceding relocation.

I also sent you a message on Slack.

PR #6013, opened 2024-10-19
adding mips3 support

v1X3Q0:
Adding support for mips3, really not that much different but gives an edge on the 10% of the instructions that are different.


galenbwill:
Can you please share some sample binaries, either here or in slack?

galenbwill:
Merged in builds >= 4.3.6831

PR #6046, opened 2024-10-26
fix: validate input type in TypeLibrary.add_alternate_name 

### LINKED ISSUE #5814: Python TypeLibrary.add_alternate_name(name) crash on null name
- Binary Ninja Version: 4.2.5802-dev
- OS: Linux (arch) / MacOS
- OS Version: 6.6.12-hardened1-1-hardened / Sonoma 14.6
- CPU Architecture: x86_64 / M1 

**Bug Description:**
In the Python API, when calling `TypeLibrary.add_alternate_name(name)` with name as null value, Binary Ninja will crash

**Steps To Reproduce:**
1) Load a Binary (in GUI or through the API)
2) Use the following code:
```py
name = None
tl = TypeLibrary.new(bv.arch, "some_type_lib")
typelib.add_alternate_name(name)
```
3) Crash

**Expected Behavior:**
I would expect it to raise an error -- something along the lines of "Expected a string, got TYPE" ?

Zerotistic:
Fixes #5814

### DIFF
--- a/python/typelibrary.py
+++ b/python/typelibrary.py
@@ -183,6 +183,8 @@ class TypeLibrary:
 
 	def add_alternate_name(self, name: str) -> None:
 		"""Adds an extra name to this type library used during library lookups and dependency resolution"""
+		if type(name) != str:
+			raise ValueError(f"Expected name to be str, got {type(name)}")
 		core.BNAddTypeLibraryAlternateName(self.handle, name)
 
 	@property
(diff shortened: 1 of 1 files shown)


python/typelibrary.py
```diff
@@ -183,6 +183,8 @@ def alternate_names(self) -> List[str]:
 
 	def add_alternate_name(self, name: str) -> None:
 		"""Adds an extra name to this type library used during library lookups and dependency resolution"""
+		if type(name) != str:
```
galenbwill: This is not idiomatic python.

Please change to 
```python
        if not isinstance(name, str):
```

galenbwill (changes requested):

python/typelibrary.py
Zerotistic: Changed that, my bad

galenbwill (approved):

Issue #6003, opened 2024-10-17
RISC-V: Lifting bug in `JALR rd, rs1, imm` when `rd == rs1`


jeanmicheldeva:
See here:
https://github.com/Vector35/binaryninja-api/blob/7d0b6bc3f49070a10d6257628bbdf93e52d87fed/arch/riscv/src/lib.rs#L1220-L1234

If `rd == rs1`, but is neither `zero` or `ra` (`x0` or `x1` resp.), the above code will lift the `jalr rd, rs1, imm` instruction as follows:
```
# let's say rd == rs1 == t1
# target = t1 + imm
t1 = pc + 4            # inst_len==4
jump(target) <=> jump(t1 + imm) <=> jump(pc + 4 + imm)
```
Whereas the intended code should be lifted as:

```
tmp_register = t1 + imm
t1 = pc + 4
jump(tmp_register)
```

jeanmicheldeva:
I patched it as follows, and it seems to work:
```Rust
let target = il.add(max_width, Register::from(rs1), imm).build();
match (rd.id(), rs1.id(), imm) {
    (0, 1, 0) => il.ret(target).append(),  // jalr zero, ra, 0
    (1, _, _) => il.call(target).append(), // indirect call
    (0, _, _) => il.jump(target).append(), // indirect jump
    (_, _, _) => {
        // indirect jump with storage of next address to non-`ra` register
        if rd.id() == rs1.id() {
            let tmp: llil::Register<Register<D>> = llil::Register::Temp(0);
            il.set_reg(max_width, tmp, target).append();
            il.set_reg(
                max_width,
                Register::from(rd),
                il.const_ptr(addr.wrapping_add(inst_len)),
            )
            .append();
            il.jump(tmp).append();
        } else {
            il.set_reg(
                max_width,
                Register::from(rd),
                il.const_ptr(addr.wrapping_add(inst_len)),
            )
            .append();
            il.jump(target).append();
        }
    }
}
```

However, I encountered another problem: "functions" that are called with `jalr rd, rs1, imm` usually end with `jr rd`. Obviously, the control-flow is the target "function" is unresolved, since `rd` does not have defined value in the context of the function. 

As a matter of fact, this target "function" (ending with `jr rd`) is most probably just one or multiple basic blocks that are shared between multiple calling functions for space optimization.

Is there any way to tell BN to treat these target "functions" as such ? (i.e. not functions, but basic blocks to reattach to different parent functions).
[...]

PR #6064, opened 2024-10-30
RISC-V JALR: Handle cases where `rd`==`rs1`, and `rd`!=`zero`

jeanmicheldeva:
Solves #6003

### DIFF
--- a/arch/riscv/src/lib.rs
+++ b/arch/riscv/src/lib.rs
@@ -711,6 +711,8 @@ impl<D: 'static + RiscVDisassembler + Send + Sync> architecture::Architecture fo
                     };
 
                     res.add_branch(branch_type, None);
+                } else {
+                    res.add_branch(BranchInfo::Unresolved, None);
                 }
             }
             Op::Beq(ref b)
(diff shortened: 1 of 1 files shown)


emesare:
_Adding the unresolved branch_

Before:
image
After:
image

emesare:
So half of this PR was merged with https://github.com/Vector35/binaryninja-api/pull/6213. Your other change, adding the unresolved branch to JALR, is being considered. I am not entirely sure if there was a reason why the branch was left out which is why I can't merge that part now.

emesare:
image

For an example this is the issue with not adding a branch and what this PR would fix.

emesare:
Talking to @rssor 

>so it was intentional, and when rd is ra especially we don't want it (those are, in fact, calls)
>plt stubs are always going to be super bespoke and, generally speaking, we don't want calls to them to be showing up anyways (since they can't be meaningfully analyzed)

If you look at https://github.com/Vector35/binaryninja-api/pull/6064#issuecomment-2509481139 you can see an example of the function calls.

Looking at your comment https://github.com/Vector35/binaryninja-api/issues/6003#issuecomment-2426061810 it seems you are running into a different set of issues resulting from the omitted branch info, if you can share more information about what was fixed by this change maybe their is some alternative solution.

emesare:
I am going to close this as done, the comment above explains my reasoning for not including your unresolved branch for the JALR instruction is in the above comments. If you have anything further please comment on this PR. Thanks again.

PR #6078, opened 2024-11-04
RISC-V: Implements MULH, MULHU, MULHSU lifting

jeanmicheldeva:
I hope this pull request implements the three instructions correctly. I only tested it with a 32 bits binary having a MULHU instruction. 

image
See https://www.cs.sfu.ca/~ashriram/Courses/CS295/assets/books/rvbook.pdf


emesare:
Added with https://github.com/Vector35/binaryninja-api/commit/fce3b5ba8e08f22b0d1604ac350348d5077e8e90 and on `4.3.6491-dev`.

Thank you!

PR #6124, opened 2024-11-13
[SharedCache] Fix some memory leaks in shared cache loading

bdash:
This fixes three sources of leaks:
1. `MMappedFileAccessor::ReadBuffer` was returning a heap-allocated `DataBuffer`, but no callers were ever deleting it. There does not appear to be any reason to heap allocate the `DataBuffer` as the type is effectively a smart pointer wrapper around `BNDataBuffer`. It is now returned by value instead.
2. `MMappedFileAccessor::ReadBuffer` was allocating a buffer, copying data into it, and then handing that allocation to the `DataBuffer` constructor. The constructor copies data into a new allocation it owns so this allocation is unnecessary and was being leaked.
3. `SharedCache::Store` was appending a JSON array to itself rather than to a similarly named variable. This cyclic data structure was not correctly freed.

Before this change loading a macOS shared cache would leak over 1GB of memory:
```
leaks Report Version: 4.0, multi-line stacks
Process 34177: 30805047 nodes malloced for 11333150 KB
Process 34177: 2512405 leaks for 1087361648 total leaked bytes.
```

After this change the leaks are in the 10s of kilobytes and appear to be unrelated to shared cache loading:
```
leaks Report Version: 4.0, multi-line stacks
Process 69005: 25357627 nodes malloced for 9319529 KB
Process 69005: 240 leaks for 18400 total leaked bytes.
```

Additionally, the memory footprint of `binaryninja` after loading the macOS shared cache drops from 8.3GB to 6.9GB.

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -223,7 +223,7 @@ void SharedCache::PerformInitialLoad()
 
 	m_baseFilePath = path;
 
-	DataBuffer sig = *baseFile->ReadBuffer(0, 4);
+	DataBuffer sig = baseFile->ReadBuffer(0, 4);
 	if (sig.GetLength() != 4)
 		abort();
 	const char* magic = (char*)sig.GetData();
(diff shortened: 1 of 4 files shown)


bdash:
@0cyn

bdash:
Looks like these changes were cherry-picked into dev in 0d27f74  and d553396.

PR #6129, opened 2024-11-14
[SharedCache] Use basic copy-on-write for viewStateCache

bdash:
Copying the state from the cache into a new `SharedCache` object is done with a global lock held and is so expensive that it results in much of the shared cache analysis running on a single thread, with others blocked waiting to acquire the lock.

The cache now holds a `std::shared_ptr` to the state. New `SharedCache` objects take a reference to the cached state and only create their own copy of it the first time they perform an operation that would mutate it. The cached copy is never mutated, only replaced, so there is no danger of modifying the state out from under a `SharedCache` object. Since the copy happens at first mutation, it is performed without any global locks held. This avoids blocking other threads.

This cuts the initial load time of a macOS shared cache from 3 minutes to 70 seconds, and cuts the time taken to load and analyze AppKit from multiple hours to around 14 minutes. The process now consistently uses 8-14 CPU cores rather than being limited to 1 core.

---

A couple of notes:

1. This depends on #6127. Since GitHub doesn't really support dependent changes that is also included on this branch. It can be dropped once #6127 is merged. You can use <https://github.com/bdash/binaryninja-api/compare/dsc-serialization...bdash:binaryninja-api:dsc-cash-cow?expand=1> to view the diff excluding the serialization changes.
2. This approach to copy-on-write is very naive, but easy to reason about. Copying the entirety of `State` when performing the first mutation is still quite expensive. There are a number of large vectors and hash maps whose items are not cheap to copy. There's more scope for improving performance by copying only the parts of the state being mutated. I went down that path but it was hard to enforce invariants (ensuring that everything we're mutating has been copied) and so harder to reason about correctness.
3. Rethinking how this data is used or the data structures involved may be a better long-term solution.


plafosse:
Thank you so much for your work on this!

plafosse:
Just wanted to give you a heads up on our plan for your PRs. We're in the process of releasing, 4.2 and this won't make the cut off. We're going to accept this PR after the release. So it might be a week or 2 before it gets accepted. Again thanks for the great commits!

bdash:
https://github.com/bdash/binaryninja-api/commit/dsc-persistent-data-structures takes copy-on-write one step further by using persistent data structures from [immer](https://github.com/arximboldi/immer) for the various vectors and maps that make up the in-memory state related to the shared cache. Their copy-on-write and structural sharing significantly reduces the amount of copying that occurs, cutting the time spent loading from the shared cache by nearly 50% vs this PR. It is a more invasive change that could benefit from some refactoring.

0cyn:
Merged via [8024cbe3](https://github.com/Vector35/binaryninja-api/commit/8024cbe3eb257c39d17e77b1f445ceb28d02b1ed)

PR #6139, opened 2024-11-15
[SharedCache] Switch to SAX-based writer API for JSON serialization

bdash:
Building up an in-memory representation of the JSON document was performing a lot of temporary memory allocations. Using the SAX-based writer API avoids this work by directly writing the desired types. This cuts the time spent serializing the state to JSON by around half.

One additional benefit of the SAX-based writer API is that it's now possible for serialization to operate on individual types rather than having to serialize a complex type in a single operation in order to assign it to an object field. `Serialize` is updated to work on a single value at a time, with function templates for types like `std::pair`, `std::vector`, and `std::unordered_map` delegating to `Serialize` overloads for the types they contain. This removes the repetition that was previously required for implementing serialization arrays or maps of different types.

Deserialization continues to use the document-based API as using the SAX-based reader API is cumbersome.

This branch builds on the work in #6127 and should be compared against it. You can use https://github.com/bdash/binaryninja-api/compare/dsc-serialization...bdash:binaryninja-api:dsc-serialization-2?expand=1 to view the diff excluding the serialization changes.


0cyn:
Merged via [a2e5d061](https://github.com/Vector35/binaryninja-api/commit/a2e5d06195c6ec725fa311220ea7037abbb9ce69)

PR #6149, opened 2024-11-17
[SharedCache] Preserve flags when splitting a memory region

bdash:
The flags field was being left uninitialized which could result in the region being mishandled during later analysis depending on what random value its memory happened to take.

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -846,21 +846,19 @@ void SharedCache::PerformInitialLoad()
 						// Part before the overlap
 						if (regionStart < segmentStart)
 						{
-							MemoryRegion newRegion;
+							MemoryRegion newRegion(*it);
 							newRegion.start = regionStart;
 							newRegion.size = segmentStart - regionStart;
-							newRegion.prettyName = it->prettyName;
-							newRegions.push_back(newRegion);
+							newRegions.push_back(std::move(newRegion));
 						}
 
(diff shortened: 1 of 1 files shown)


bpotchik:
Perhaps unrelated to this pull request but I have two observations so I'm just going to plop this here:

1) I'm unfamiliar with the DSC source, however there appears to be a MemoryRegion object which is part of the DSC namespace. BinaryNinja also support MemoryRegions with the new MemoryMap. Currently the objects themselves persist int he core and APIs exist to get the data. It also supports overriding flags, etc.

I'm sure there is probably some overlap of the between the two. I'm curious if it's possible for DSC MemoryRegions to be replaced by those provided by BinaryNinja.

2) When adding AutoSegments or UserSegments, we provide a `BeginBulkAddSegments` and `EndBulkAddSegments`. Modifying the segments causes a new MemoryMap to be generated which can be an expensive operation if there are a large number of segments. Worth considering for DSC if there are many segments being added.

bdash:
`MemoryRegion` holds state about a region in the shared cache, the vast majority of which have not been loaded into the `BinaryView` (and likely never will be). There are 20-30,000 regions in a typical shared cache. The half dozen or so regions for an individual image are loaded into the view when a user choose to load that image from the shared cache.

0cyn:
Merged via [90af5bdc](https://github.com/Vector35/binaryninja-api/commit/90af5bdc87977c4c39ed7594f882dfad41756fb5)

PR #6172, opened 2024-11-21
[SharedCache] Fix slide info parsing

WeiN76LQh:
The issue this commit fixes was causing `SharedCache::ParseAndApplySlideInfoForFile` to completely fail to work with v5 slide info, which had a lot of knock on effects, i.e. lots of Objective-C analysis was failing due to invalid pointers which hadn't been fixed up.

`dyld_cache_slide_info5` has 4 bytes of padding before `value_add`. Whilst `value_add` is not actually being read from, `SharedCache::ParseAndApplySlideInfoForFile` will read at a location in the file based on the size of the structure `dyld_cache_slide_info5`. This being off by 4 bytes basically broke v5 slide info fixups.

With this fix many more Objective-C functions have names and a lot more `msgSend` calls are fixed up.

### DIFF
--- a/view/sharedcache/core/ObjC.cpp
+++ b/view/sharedcache/core/ObjC.cpp
@@ -1087,16 +1087,16 @@ void DSCObjCProcessor::ApplyMethodTypes(Class& cls)
 	}
 }
 
-void DSCObjCProcessor::PostProcessObjCSections(VMReader* reader, std::string baseName)
+void DSCObjCProcessor::PostProcessObjCSections(VMReader* reader)
 {
 	auto ptrSize = m_data->GetAddressSize();
-	if (auto imageInfo = m_data->GetSectionByName(baseName + "::__objc_imageinfo"))
+	if (auto imageInfo = m_data->GetSectionByName("__objc_imageinfo"))
 	{
 		auto start = imageInfo->GetStart();
 		auto type = Type::NamedType(m_data, m_typeNames.imageInfo);
 		m_data->DefineDataVariable(start, type);
 	}
-	if (auto selrefs = m_data->GetSectionByName(baseName + "::__objc_selrefs"))
+	if (auto selrefs = m_data->GetSectionByName("__objc_selrefs"))
 	{
 		auto start = selrefs->GetStart();
 		auto end = selrefs->GetEnd();
@@ -1119,7 +1119,7 @@ void DSCObjCProcessor::PostProcessObjCSections(VMReader* reader, std::string bas
 			DefineObjCSymbol(DataSymbol, type, "selRef_" + sel, i, true);
 		}
 	}
-	if (auto superRefs = m_data->GetSectionByName(baseName + "::__objc_classrefs"))
+	if (auto superRefs = m_data->GetSectionByName("__objc_classrefs"))
 	{
 		auto start = superRefs->GetStart();
 		auto end = superRefs->GetEnd();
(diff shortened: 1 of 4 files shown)


WeiN76LQh:
I forgot there was another issue with Objective-C processing that meant a bunch was being skipped. This is fixed by the 2nd commit.

Something which I didn't fully touch on in the commit is the issue of processing Objective-C sections which may require data from another image that is not yet loaded. I think the naive but simplistic approach would be to go through all Objective-C sections after loading an image and re-processing all of them. I think this would solve the problem but it would be quite costly, particularly if you're loading alot of sections. In that case it would make sense to have the ability to batch load images and then defer Objective-C processing until the end. Additionally this may mess with changes that the user has made, although I think its unlikely the user would be messing about with Objective-C sections unless they're using a script, in which case they could re-run their script.

What probably makes the most sense in the long run is to keep track of any bits of Objective-C data that failed analysis and then re-analyze those bits after loading of images.

bdash:
I took this branch for a quick test and found that it breaks Obj-C message names when loading from the macOS 15.1 shared cache.

Screenshot 2024-11-21 at 15 03 26

I think this is tickling another existing bug in how `dyld_cache_slide_info5` is handled rather than being the fault of your change.

`dyld` checks `page_start[i]` for  `DYLD_CACHE_SLIDE_V5_PAGE_ATTR_NO_REBASE`. If not equal, it divides `page_start[i]` by `sizeof(uint64_t)` and uses that value as the initial offset into the page.

The code in `ParseAndApplySlideInfoForFile` for `dyld_cache_slide_info5` leaves `page_start[i]` as-is. It works correctly when `page_start[i]` is 0, which is admittedly the most common case, but generates garbage otherwise.

```diff
-                                       pageStartFileOffsets.push_back(mapping.mappingInfo.fileOffset + (pageSize * i) + pageStart);
+                                       pageStartFileOffsets.push_back(mapping.mappingInfo.fileOffset + (pageSize * i) + (pageStart / 8));
```

WeiN76LQh:
> ```diff
> -                                       pageStartFileOffsets.push_back(mapping.mappingInfo.fileOffset + (pageSize * i) + pageStart);
> +                                       pageStartFileOffsets.push_back(mapping.mappingInfo.fileOffset + (pageSize * i) + (pageStart / 8));
> ```

I'm not 100% sure this is enough/or correct, I still had issues with v5 slides at least. I've gone and reworked the slide parsing code and think I've got it fixed. The way "starts" were pulled out into a vector and then processed below in another loop seemed to be making things difficult and disjointed for no visible benefit. Instead I'm processing "starts" in the loops where they are read and largely mimicking exactly what `dyld` does. I'm just trying to test things to ensure I've got it correct.

WeiN76LQh:
The latest commit re-writes some of the slide parsing stuff. More details in the commit message but I think at this point it might be working. Not everything has been thoroughly tested but I'm pretty confident v3 and v5 are now working as intended. v2 should be but less testing of it has been done.

bdash:
I did some testing against the macOS 15.1 dyld shared cache I was seeing problems with yesterday. I've confirmed the slide handling exactly matches what dyld is doing, so this change looks good.

Sadly I'm still seeing bogus Obj-C selector names. I'm guessing that's a result of Obj-C data being processed that wasn't before and that there's a pre-existing bug in parsing Obj-C metadata that wasn't visible due to the data being skipped. I'll look into that separately.

WeiN76LQh:
> Sadly I'm still seeing bogus Obj-C selector names. I'm guessing that's a result of Obj-C data being processed that wasn't before and that there's a pre-existing bug in parsing Obj-C metadata that wasn't visible due to the data being skipped. I'll look into that separately.

Does this happen to be a result of [this comment on issue 6173](https://github.com/Vector35/binaryninja-api/issues/6173#issuecomment-2495196535)?

WeiN76LQh:
I've removed the fixes for the Objective-C code because it didn't make sense to be part of this pull request. I'll re-add it in another one which is focussed on some Objective-C changes

0cyn:
Hi! Thank you so much for the PR. We're looking at getting this merged in today assuming testing goes well

0cyn:
Merged via [8b212f86](https://github.com/Vector35/binaryninja-api/commit/8b212f865c2da667dfe791ecf4d994e56228ea85)

PR #6174, opened 2024-11-21
[SharedCache] Serialize `SharedCache::m_symbolInfos`

WeiN76LQh:
`SharedCache::m_symbolInfos` isn't being serialized but there is an attempt to deserialize it. This commit adds in the code to serialize it.

I slightly modified the format because I didn't really understand how it was expected to be serialized based on the deserialization code. The deserialization code looked wrong to me but its likely a misunderstanding on my part. I kept it similar to how `m_exportInfos` is serialized.


0cyn:
Hi! I'll be taking a look at this PR in conjunction with https://github.com/Vector35/binaryninja-api/pull/6210, seeing as removing symbol info serialization entirely in favor of using User Symbols may be a better approach.

PR #6182, opened 2024-11-23
fix shrx, shlx, sarx, rorx

yrp604:
All had the same off by 1 in addressing operands.

ref: 
- https://www.felixcloutier.com/x86/sarx:shlx:shrx
- https://www.felixcloutier.com/x86/rorx


emesare:
Added with https://github.com/Vector35/binaryninja-api/commit/34bd2341f9fb8d5e8bc57639cb0ec547c3b8596b and on `4.3.6491-dev`.

Thank you!

PR #6184, opened 2024-11-24
[SharedCache] Have VM object track ranges rather than pages

bdash:
There are typically only a few dozen mappings, while there can be millions of pages. This reduces the amount of time spent populating the mapping from region to file accessor, along with the memory usage of the same.


0cyn:
Merged via [c70afdbe](https://github.com/Vector35/binaryninja-api/commit/c70afdbe472b80151cbe65395b95283848ade62e)

0cyn:
If there is ever a situation where we are using VM address translation much more intensely this might need to be reconsidered, as the original per-page implementation was optimized for read performance, at the cost of initialization performance. But currently I don't think this is typically the case.

PR #6194, opened 2024-11-25
[SharedCache] Remove unnecessary lock in `SharedCache` destructor

WeiN76LQh:
From what I can tell the lock being taken in `SharedCache::~SharedCache` was purely for the decrement of `sharedCacheReferences`, however its an atomic so a lock isn't necessary. The lock being taken is extremely contentious and therefore often slow to be acquired. This resulted in a surprising amount of execution time spent in the `SharedCache` destructor. Nothing hugely significant but a quick and easy win to remove this single line of code.


0cyn:
Thanks for the PR! I'll look at getting this merged sometime this week.

PR #6195, opened 2024-11-25
[SharedCache] Cache references to type libraries

WeiN76LQh:
`SharedCache::FindSymbolAtAddrAndApplyToAddr` significantly impacts loading and analysis of DSC libraries. Its a hot path and the main cause of its slow execution time is the loading of a type library by name. By caching references to type libraries this function spends 90%+ less time executing.

The cache is a global because `SharedCache` gets destroyed and re-created alot so I think this is the best place to store cached references. The downside to this is it'll cause a memory leak as I map entries using the view session ID. Entries won't be removed once the view is closed as I'm unsure how to do this. That said I think the performance benefits far outway the potential memory leak which will likely have no noticeable affect on the user experience or BN memory consumption.

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -83,10 +83,17 @@ std::unordered_map<uint64_t, BNDSCViewLoadProgress> progressMap;
 
 struct ViewSpecificMutexes {
 	std::mutex viewOperationsThatInfluenceMetadataMutex;
-	std::mutex typeLibraryLookupAndApplicationMutex;
 };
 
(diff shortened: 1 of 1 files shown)


bdash:
I was thinking about this after you mentioned it was a bottleneck on https://github.com/bdash/binaryninja-api/pull/2. I noticed that the way that some of the current view-specific state, including the mutexes, are accessed has a data race.  I ended up implementing a very similar change (#6196) that fixes those data races and is able to free memory when the shared cache is closed.

WeiN76LQh:
Closed in favour of https://github.com/Vector35/binaryninja-api/pull/6196

PR #6197, opened 2024-11-25
[SharedCache] Use `m_exportInfos` as an export list cache

WeiN76LQh:
`SharedCache::ParseExportTrie` is getting called a lot during DSC library loading and analysis. In large part due to the hot path `SharedCache::FindSymbolAtAddrAndApplyToAddr`. Its unnecessary for it to be being called more than once per DSC header as the export list symbol information is stored in `SharedCache::m_exportInfos`.

This commit adds the function `SharedCache::GetExportListForHeader`, which will either return the header's list of symbol information cached in `SharedCache::m_exportInfos` or call `SharedCache::ParseExportTrie` and cache the results in `SharedCache::m_exportInfos`.

This should also improve the execution time of `SharedCache::LoadAllSymbolsAndWait`.

Further improvement here would be to add locking to `SharedCache::GetExportListForHeader` so that races don't result in redundant parsing of the export trie for the same header if multiple threads call `SharedCache::GetExportListForHeader` at the same time for the same header. This only really matters during initial loading of the first image because from what I can tell that results in parsing all the export trie's anyway.

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -57,7 +57,7 @@ int count_trailing_zeros(uint64_t value) {
 
 struct SharedCache::State
 {
-	std::unordered_map<uint64_t, std::unordered_map<uint64_t, Ref<Symbol>>>
+	std::unordered_map<uint64_t, std::shared_ptr<std::unordered_map<uint64_t, Ref<Symbol>>>>
 		exportInfos;
 	std::unordered_map<uint64_t, std::vector<std::pair<uint64_t, std::pair<BNSymbolType, std::string>>>>
 		symbolInfos;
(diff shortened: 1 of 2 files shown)


view/sharedcache/core/SharedCache.cpp
```diff
@@ -2779,6 +2779,31 @@ std::vector<Ref<Symbol>> SharedCache::ParseExportTrie(std::shared_ptr<MMappedFil
...
+		return it->second;
+	}
+	else
+	{
+		std::shared_ptr<MMappedFileAccessor> linkeditFile = provideLinkeditFile();
+		if (!linkeditFile)
+			return std::vector<std::pair<uint64_t, std::pair<BNSymbolType, std::string>>>();
+
+		auto exportList = SharedCache::ParseExportTrie(linkeditFile, header);
+		std::vector<std::pair<uint64_t, std::pair<BNSymbolType, std::string>>> exportMapping(exportList.size());
```
bdash: Since this is now the only caller of `ParseExportTrie`, it'd be beneficial to have it produce data in the format that this expects. This will eliminate the need for `ParseExportTrie` / `ReadExportNode` to allocate a bunch of `Symbol` instances only for this code to copy data out of them and then discard them. All those memory allocations and copying all of the names around is surprisingly expensive.

WeiN76LQh: Yes that was lazy of me, I'll add another commit to do that

view/sharedcache/core/SharedCache.cpp
```diff
@@ -2779,6 +2779,31 @@ std::vector<Ref<Symbol>> SharedCache::ParseExportTrie(std::shared_ptr<MMappedFil
 	return symbols;
 }
 
+
+std::vector<std::pair<uint64_t, std::pair<BNSymbolType, std::string>>> SharedCache::GetExportListForHeader(SharedCacheMachOHeader header, std::function<std::shared_ptr<MMappedFileAccessor>()> provideLinkeditFile)
```
bdash: This returns a copy of this chonky vector each time it is called. If the thread-safety story permits it, it would be better to return a const reference instead.

It might be worth replacing this whacky nested `std::pair` with a struct with named fields that describe what they contain. That'd clear up a lot of the confusing `pair.first` / `pair.second.first` stuff.  There's an `ExportNode` structure declared in this file that looks to be intended for this purpose, but is currently never used for anything.

view/sharedcache/core/SharedCache.cpp
```diff
@@ -2794,30 +2819,32 @@ std::vector<std::pair<std::string, Ref<Symbol>>> SharedCache::LoadAllSymbolsAndW
...
-		{
-			m_logger->LogWarn("Serious Error: Failed to open export trie %s for %s", header->exportTriePath.c_str(), header->installName.c_str());
-			continue;
-		}
-		auto exportList = SharedCache::ParseExportTrie(mapping, *header);
-		std::vector<std::pair<uint64_t, std::pair<BNSymbolType, std::string>>> exportMapping;
+		auto exportList = GetExportListForHeader(*header, [&]() {
+			try {
+				auto mapping = MMappedFileAccessor::Open(m_dscView, m_dscView->GetFile()->GetSessionId(), header->exportTriePath)->lock();
+				doSave = true;
```
bdash: Is this the only difference between the various "open a file" lambdas passed to `GetExportListForHeader`? If so you could simplify these call sites by having `GetExportListForHeader` take `bool* didModifyExportList = nullptr`. Callers that are interested in knowing that `m_exportInfos` was modified could pass a pointer to bool in. Others would omit it. This would let you move the file opening logic into `GetExportListForHeader`.

view/sharedcache/core/SharedCache.cpp
WeiN76LQh: I think this should be possible as once a vector is created it shouldn't be reallocated by `m_exportInfos`?

view/sharedcache/core/SharedCache.cpp
WeiN76LQh: I would have done this but the call in `SharedCache::InitializeHeader` has a different way of loading a mapped file.

`bool* didModifyExportList = nullptr` - I didn't do this just cause it seemed this is the only instance where it mattered so it seemed 50/50 whether the extra parameter made sense or what I did. For future stuff though it might make more sense to have the parameter so its clear to a caller that they need to do a save if they didn't know.

view/sharedcache/core/SharedCache.cpp
bdash: The thread safety problems arise when `m_exportInfos` is modified. If the hash table needs to grow it will reallocate and moves its contents. This would invalidate any existing references to data it stores.

One solution that doesn't require any additional locking would be to add a layer of indirection: have `m_exportInfos` store a `std::unique_ptr` to the vector. Then the vector itself will never move, only the pointer that owns it. You could then safely return a reference to the pointed-at vector from this function.

view/sharedcache/core/SharedCache.cpp
bdash: Ah, I missed that `SharedCache::InitializeHeader` is doing something different. Bummer!

view/sharedcache/core/SharedCache.cpp
WeiN76LQh: I'm not sure using `std::unique_ptr` is pratical here because `m_exportInfos` in `ViewStateCacheStore` causes compiler errors due to copying mechanics when `m_exportInfos` has a `unique_ptr` type in its template. `shared_ptr` works fine and I don't see much of an issue in using that instead.

WeiN76LQh:
I think these latest commits fix all the issues @bdash mentioned in the comments. Also there maybe some race issues here with accessing `m_exportInfos` in `SharedCache::GetExportListForHeader` but there's a lot to reason about because callers may be holding locks that make it fine. In some quick tests its working fine. I didn't add any locks in because ultimately I think PR https://github.com/Vector35/binaryninja-api/pull/6129 solves that problem anyway.

view/sharedcache/core/SharedCache.cpp
```diff
@@ -2990,7 +2998,7 @@ bool SharedCache::SaveToDSCView()
 		c.m_dyldDataRegions = m_dyldDataRegions;
 		c.m_nonImageRegions = m_nonImageRegions;
 		c.m_baseFilePath = m_baseFilePath;
-		c.m_exportInfos = m_exportInfos;
+		//c.m_exportInfos = m_exportInfos;
```
bdash: I guess you commented this out while testing the `std::unique_ptr` approach.

WeiN76LQh: Yup 🤦

0cyn:
Thank you so much for this PR, this was the intended original functionality of the export trie list in memory! Glad to merge this sometime over the next couple of weeks, may need to be looked at in conjunction with the things listed on https://github.com/Vector35/binaryninja-api/pull/6210.

0cyn:
Hi! Would you be able to rebase this on the latest commit on dev?

WeiN76LQh:
Yes, I'll sort that out now

PR #6202, opened 2024-11-26
[SharedCache] Reduce lock contention in `SharedCache::LoadSectionAtAddress`

WeiN76LQh:
When loading a number of sections, for instance when an image is loaded, many of the analysis threads bottleneck in `SharedCache::LoadSectionAtAddress` trying to acquire the `viewOperationsThatInfluenceMetadataMutex` lock. This is a very expensive lock to try and acquire as it is used in a lot of places and held for long durations.

This commit improves performance in `SharedCache::LoadSectionAtAddress` by not acquiring the `viewOperationsThatInfluenceMetadataMutex` lock until absolutely necessary, i.e. when something actually needs to be loaded. Often what happens is many threads try to load the same sections but ending up queuing up to do this one at a time. The commit adds per memory region locking so that threads only block waiting for the memory region they require to be loaded. In most cases though, if the region is already loaded they won't wait at all because no lock is required to determine if this is the case.

**Note:** this PR is built on @bdash's PR https://github.com/Vector35/binaryninja-api/pull/6196 due to the fixes and improvements in view-specific state, as this commit adds view-specific mutexes. I thought it would be prudent to make use of those changes.

To be honest I'm not 100% sure about the use of a map of mutexes to provide per memory region locking. It just feels a bit excessive but it works and the memory consumption of it won't be anything significant.

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -1577,8 +1577,8 @@ bool SharedCache::LoadSectionAtAddress(uint64_t address)
 
 				// The region appears not to be loaded. Acquire the loading lock, re-check 
 				// that it hasn't been loaded and if it still hasn't then actually load it.
-				std::unique_lock<std::mutex> memoryRegionLoadingLockslock(ViewSpecificStateForView(m_dscView)->memoryRegionLoadingMutexesMutex);
-				auto& memoryRegionLoadingMutex = ViewSpecificStateForView(m_dscView)->memoryRegionLoadingMutexes[stubIsland.start];
+				std::unique_lock<std::mutex> memoryRegionLoadingLockslock(m_viewSpecificState->memoryRegionLoadingMutexesMutex);
+				auto& memoryRegionLoadingMutex = m_viewSpecificState->memoryRegionLoadingMutexes[stubIsland.start];
 				// Now the specific memory region's loading mutex has been retrieved, this one can be dropped
 				memoryRegionLoadingLockslock.unlock();
 				// Hold this lock until loading of the region is done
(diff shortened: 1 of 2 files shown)


view/sharedcache/core/SharedCache.cpp
```diff
@@ -1584,6 +1574,46 @@ bool SharedCache::LoadSectionAtAddress(uint64_t address)
 				{
 					return true;
 				}
+
+				// The region appears not to be loaded. Acquire the loading lock, re-check 
+				// that it hasn't been loaded and if it still hasn't then actually load it.
+				std::unique_lock<std::mutex> memoryRegionLoadingLockslock(ViewSpecificStateForView(m_dscView)->memoryRegionLoadingMutexesMutex);
```
bdash: The view-specific state can be found in `m_viewSpecificState` rather than using `ViewSpecificStateForView(…)` to look it up in the global hash table.

0cyn:
Thank you for the PR! We're looking into getting this merged over the next couple of weeks

WeiN76LQh:
I'm currently doing some rather sweeping changes to the code in regards to the usage of locking. I'm not 100% confident in some of the changes I've made in this PR and I actually think I have a better version in the works. I've got it so I can do bulk parallelised loading of images which can dramatically bring down load times to very acceptable levels. Although it is somewhat reliant on #6271 being fixed and the UI freezes if many changes are made that quickly so you have to do it in the background.

WeiN76LQh:
In light of #6393 (and the comment I made on it) and my other concerns mentioned above I'm going to close this PR. This issue can be revisited in the future once that PR has landed.

PR #6213, opened 2024-11-30
Fix RISC-V JALR lift when rs1==rd

ehntoo:
When the source and destination register are the same, the current implementation of lifting for `jalr` in RISC-V results in the jump target being overwritten by the return address. This PR avoids the issue by copying the jump target to a LLIL temporary register before writing the destination register.

This situation comes up in certain binaries when LLVM optimizes for size and performs compiler outlining of register stack spills. The attached zipfile contains a build of a rust RISC-V Embassy project that demonstrates the issue in a function at 0x00000b08.
[riscv-jalr-issue.zip](https://github.com/user-attachments/files/17964210/riscv-jalr-issue.zip)

### DIFF
--- a/arch/riscv/src/lib.rs
+++ b/arch/riscv/src/lib.rs
@@ -1242,6 +1242,8 @@ impl<D: 'static + RiscVDisassembler + Send + Sync> architecture::Architecture fo
                     (1, _, _) => il.call(target).append(), // indirect call
                     (0, _, _) => il.jump(target).append(), // indirect jump
                     (_, _, _) => {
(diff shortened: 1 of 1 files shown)


emesare:
The lifting change is good, however will require changes to the function recognizer, see the second pair of images.

Before:
image
After:
image

Before:
image
After:
image

emesare:
I went ahead and added the required changes, thank you for the PR!

Also thank you to @jeanmicheldeva, apologies that your PR wasn't merged, I am going to try and get your other changed merged shortly.

Relevant PR: https://github.com/Vector35/binaryninja-api/pull/6064

PR #6236, opened 2024-12-06
MachO/ObjC: Avoid generating names containing non-printable chars

droe:
Avoid generating ObjC symbol names containing ASCII control characters for both 8 and 16 bit character encodings.

Behaviour is like before for each, except that ASCII control characters (0..0x1f) are replaced by ?, except for carriage return, line feed, and tabulator characters, which are replaced by \r, \n and \t respectively.

For an example CF string of `"Hello world!\r\n"`, with this change, the symbol would be named `cfstr_Hello_world!\r\n`, instead of w/o this change `cfstr_Hello_world!  ` with embedded CR/LF that display as slightly weird spaces.

Tested on latest stable by rebasing this change on top of [v4.2.6455-stable](https://github.com/Vector35/binaryninja-api/releases/tag/v4.2.6455-stable).

### DIFF
--- a/view/macho/objc.cpp
+++ b/view/macho/objc.cpp
@@ -1441,12 +1441,22 @@ void ObjCProcessor::ProcessCFStrings()
 					uint8_t* rawData = static_cast<uint8_t*>(data.GetData());
 					uint8_t* offsetAddress = rawData + bufferOff;
 					uint16_t c = *reinterpret_cast<uint16_t*>(offsetAddress);
-					if (c == 0x20)
+					if (c == 0x20) {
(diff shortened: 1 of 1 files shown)


0cyn:
Thank you so much for this PR, I'm currently working through several PRs related to sharedcache processing and will be taking a look at this one within the next couple of weeks as we finish working through those.

droe:
FWIW, the issue this PR addresses is exacerbated by latest dev builds displaying symbol names with newlines as multi-line symbol names in the symbol list, leading to a lot of wasted vertical space.

PR #6243, opened 2024-12-10
Add fmt impl for QualifiedNameAndType

pr0me:
Currently, only `Type` has a `fmt::Debug` implementation, with the effect that the printer has no knowledge of the type name.
This is undesirable for, e.g., imported types.

This PR adds a debug print for `QualifiedNameAndType` that prints the full type declaration as it would appear in an exported header file.

If a (cleaner) way to achieve this already exists, please let me know.

### DIFF
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -2582,6 +2582,75 @@ unsafe impl CoreArrayProviderInner for QualifiedNameAndType {
     }
 }
 
+impl fmt::Debug for QualifiedNameAndType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Ok(lock) = TYPE_DEBUG_BV.lock() {
+            if let Some(bv) = &*lock {
+                let container = unsafe { BNGetAnalysisTypeContainer(bv.handle) };
+
+                let printer = if f.alternate() {
+                    unsafe { BNGetTypePrinterByName(c"_DebugTypePrinter".as_ptr()) }
+                } else {
+                    unsafe { BNGetTypePrinterByName(c"CoreTypePrinter".as_ptr()) }
+                };
+                if printer.is_null() {
+                    return Err(fmt::Error);
+                }
+
+                let mut name = self.name().clone();
+
+                let mut lines: *mut BNTypeDefinitionLine = null_mut();
+                let mut count: usize = 0;
+
+                unsafe {
+                    BNGetTypePrinterTypeLines(
+                        printer,
+                        self.type_object().as_ref().handle,
+                        container,
+                        &mut name.0,
+                        64,
+                        false,
+                        BNTokenEscapingType::NoTokenEscapingType,
+                        &mut lines,
+                        &mut count,
+                    )
+                };
+                unsafe {
+                    BNFreeTypeContainer(container);
+                }
+
+                if lines.is_null() {
+                    return Err(fmt::Error);
+                }
+
+                let line_slice: &[BNTypeDefinitionLine] =
+                    unsafe { slice::from_raw_parts(lines, count) };
+
+                for (i, line) in line_slice.iter().enumerate() {
+                    if i > 0 {
+                        writeln!(f)?;
+                    }
+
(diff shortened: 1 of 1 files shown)


emesare:
That impl was always a little hacky, for now we are just going to omit it in favor of the Display impl.

See my commit here: https://github.com/Vector35/binaryninja-api/commit/1b6b9a3e10dc4640621c29df9173bbdc0fc5a8f6

Please let me know if this is satisfactory, specifically the following:

https://github.com/Vector35/binaryninja-api/blob/1b6b9a3e10dc4640621c29df9173bbdc0fc5a8f6/rust/src/types.rs#L907-L923j

CouleeApps:
@emesare I'd be up for the Display impl using GetTypeString but for Debug I've explicitly chosen to use GetLines so it prints the members of structures. The correct answer to that awful bv hack I wrote is a way to construct an empty type container without needing a bv, and then passing that to get_lines(). I can write such a method if need be (would have to be in the core).

pr0me:
Hey, thanks for looking into this!
The current state of the `rust_cleanup_0` is not really satisfactory in terms of what I want to use `fmt` / `Debug` for `QualifiedNameAndType` for.

The debug output of `Type` produces a C-compliant (besides it missing the name) representation of the type and also for structs and their sub-types, e.g.:
```
enum : uint16_t
{
    PE_ROM_IMAGE = 0x107,
    PE_32BIT = 0x10b,
    PE_64BIT = 0x20b
};
```

Looking at your commits, I didn't find an obvious way of achieving comparable output for `QualifiedNameAndType` which would be nice to have.
That was also the intention of this PR.


The reason I'm trying to do this is that I want to export header files, pretty much what the 'Anlaysis > Export Header File...' is doing. 
Is there an easier way of doing this with the Rust API?

emesare:
I see, I am going to add an example that I think does what you want, alb-eight with the side effect of requiring you to first gather all your types and have a valid binary view. Until @CouleeApps can add the required changes to the core.

```rs
  let type_printer = CoreTypePrinter::default();
  let my_structure = Type::structure(
      &Structure::builder()
          .insert_member(
              &StructureMember::new(
                  Type::int(4, false).into(),
                  "my_field".to_string(),
                  0,
                  MemberAccess::PublicAccess,
                  MemberScope::NoScope,
              ),
              false,
          )
          .finalize(),
  );

  let printed_types = type_printer.print_all_types(
      [("my_struct", my_structure)],
      // We assume you have access to a BinaryView.
      &bv,
      4,
      TokenEscapingType::NoTokenEscapingType,
  );
  
  println!("Printed types: {:?}", printed_types);
```

```c
//------------------------------------------------------------------------------
// Types for /bin/cat
//
// This header file generated by Binary Ninja Ultimate 4.3 Ultimate
//------------------------------------------------------------------------------

#ifndef BN_TYPE_PARSER
#include <stdint.h>
#include <stddef.h>
#include <stdlib.h>
#include <stdbool.h>
#include <wchar.h>

#define __packed
#define __noreturn
#define __convention(name)
#define __syscall(number)
#define __offset(...)
#define __padding
#define __named(name)
#define __inherited
#define __base(name, offset)
#define __ptr_offset(offset)
#define __data_var_refs
#define __vtable
#define __pure
#define __ptr_width(width)
#define __ptr8
#define __ptr16
// __ptr32 and __ptr64 are real keywords on MSVC
// #define __ptr32
// #define __ptr64
#define __based(...)
typedef uint16_t wchar16;
typedef uint32_t wchar32;
#endif

//------------------------------------------------------------------------------
// Forward Declarations of Structures
//------------------------------------------------------------------------------

struct my_struct;
//------------------------------------------------------------------------------
// Type Definitions
//------------------------------------------------------------------------------

// "my_struct"
struct my_struct
{
        uint32_t my_field;
};
```

I am _very_ against having the Debug impl use the type printing facilities. But I see no reason as to why you could not just use them yourself directly (assuming that which is implemented in the `rust_cleanup_0` branch is merged).

emesare:
The changes above have been added with https://github.com/Vector35/binaryninja-api/commit/6a63c17853009ce7a0071458e39c140a09440230

PR #6256, opened 2024-12-14
WARP: Update README with macOS library path instructions

cyberkaida:
The README said to use `DEP_BINARYNINJACORE_PATH` but this was not referenced.

I updated the `build.rs` to use this variable, and as a treat I added the default install locations from the [User Guide](https://docs.binary.ninja/guide/)

I think this works on Windows too, but I do not have a Windows install to test. Thank you for the hard work on WARP I am excited to use it! 💜💜🥷

### DIFF
--- a/plugins/warp/build.rs
+++ b/plugins/warp/build.rs
@@ -1,6 +1,7 @@
 #![allow(unused_imports)]
 use std::path::PathBuf;
 use std::process::Command;
+use std::path::Path;
 
 #[cfg(feature = "test")]
 fn compile_rust(file: PathBuf) -> bool {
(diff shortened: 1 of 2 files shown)


cyberkaida (commented):
This might be better as part of the BinaryNinja install instructions or a part of the `sigem` build process. Maybe you can adjust the `LC_LOAD` segment path?

For now I have just quickly added this to the README.

emesare:
Hi, thank you for the PR.

I don't know why the README says to set `DEP_BINARYNINJACORE_PATH`, that env var is only used on a [separate branch](https://github.com/Vector35/binaryninja-api/commit/ab06d72c6e35a9a2fdd0bd0ce28f8946b67ea233#diff-d4626306ca381c0fc0457e7651413225718fa235057022cbd8673d1426e8fe22).

Anyways what you want is to set `BINARYNINJADIR` instead.

Also this might interest you: https://github.com/Vector35/warp/pull/22/ (sigem docker image)

plugins/warp/build.rs
```diff
@@ -18,10 +19,48 @@ fn compile_rust(file: PathBuf) -> bool {
 }
 
 fn main() {
-    if let Some(link_path) = option_env!("BINARYNINJADIR") {
+
+    let binja_base_dir: &str = option_env!("DEP_BINARYNINJACORE_BASE_DIR").unwrap_or_else(|| {
```
emesare, line 23: Our build system needs to set its own paths, so unfortunately we can't do lookups like this.

emesare: This would work fine: https://github.com/Vector35/binaryninja-api/commit/5239ad6e8345e691fde31fd9f513fcf200e942aa

emesare:
Correct the README: https://github.com/Vector35/binaryninja-api/commit/487fa4b240ae2c2d5a712dd6e00071f8de57d0fc
Add auto linking: https://github.com/Vector35/binaryninja-api/commit/5239ad6e8345e691fde31fd9f513fcf200e942aa

Let me know if those two suffice.

FWIW we store a `lastrun` artifact on your system which can be used instead of guessing for the install.

Thank you for the PR!

psifertex:
> Correct the README: [487fa4b](https://github.com/Vector35/binaryninja-api/commit/487fa4b240ae2c2d5a712dd6e00071f8de57d0fc) Add auto linking: [5239ad6](https://github.com/Vector35/binaryninja-api/commit/5239ad6e8345e691fde31fd9f513fcf200e942aa)
> 
> Let me know if those two suffice.
> 
> FWIW we store a `lastrun` artifact on your system which can be used instead of guessing for the install.
> 
> Thank you for the PR!

Lastrun is documented here for those curious: 
https://docs.binary.ninja/guide/index.html#user-folder (it lives in your user folder)

cyberkaida:
> Correct the README: https://github.com/Vector35/binaryninja-api/commit/487fa4b240ae2c2d5a712dd6e00071f8de57d0fc
> 
> Add auto linking: https://github.com/Vector35/binaryninja-api/commit/5239ad6e8345e691fde31fd9f513fcf200e942aa
> 
> 
> 
> Let me know if those two suffice.
> 
> 
> 
> FWIW we store a `lastrun` artifact on your system which can be used instead of guessing for the install.
> 
> 
> 
> Thank you for the PR!

This looks much better, thank you!

emesare:
Closing as author is satisfied with the changes mentioned above, re-open if that isn't the case. Thanks again!

PR #6257, opened 2024-12-17
Remove IntoActivityName trait

mkrasnitski:
This trait seems relatively useless, and can be replaced by bounding on `BnStrCompatible` instead. Then, instead of passing an `activity`, users can just pass `activity.name()`, or any other string-like.


emesare:
This PR was merged with the rust refactor, https://github.com/Vector35/binaryninja-api/commit/6a63c17853009ce7a0071458e39c140a09440230. Thank you @mkrasnitski!

PR #6280, opened 2024-12-27
arm64: Lift FCVTZS Wd,Dn,#n to LLIL

droe:
Lift `FCVTZS Wd,Dn,#n` to LLIL as an intrinsic.

The intrinsic is made up in the sense that I could not find an intrinsic corresponding to this instruction variant in [Arm Neon Intrinsics Reference](https://arm-software.github.io/acle/neon_intrinsics/advsimd.html) nor in `arm_neon.h` shipping with Xcode's clang.  There is precedent for such made up intrinsics: The unsigned version of this, `FCVTZU Wd,Dn,#n`, was added as part of c3040ec.

I've added two tests to `arm64test.py` that pass (after disabling the `tests_tbl` test group that's currently failing for reasons unrelated to this PR).  Also tested that it resolves the unimplemented instruction error with an in-the-wild binary that makes use of this instruction variant (not sharable).


galenbwill:
As you point out, the intrinsics are generated from the ARM C Language Extensions at https://github.com/ARM-software/acle/

This instruction variant, like all fixed-point instruction variants, is not included in that spec, and only appears in our generated intrinsics because it was manually added to our post-processed version of the spec that feeds our internal generation script.

I'm not going to merge this PR for two reasons:
1. It is missing input and output type information
2. The change would be overwritten the next time the intrinsics code is regenerated

I am however, going to leave the PR open as a reminder to add this instruction variant to the spec the next time we run the generator, but I am marking it "Won't Fix" for now.

galenbwill (approved):
See my comment in the conversation section for more information on why this isn't getting merged.

psifertex:
Closing this per the previous comment.

PR #6297, opened 2025-01-03
[ObjC] Create a shared ObjC processor for Macho and DSC views

WeiN76LQh:
NOTE: this builds on top of #6192

Both the Macho and DSC views need to process Objective-C but have separate processor classes. It would appear that the DSC version was largely a copy and paste of the Macho view one, with some modifications. The majority of code overlaps between the 2 so it doesn't make sense to maintain 2 and copy and paste improvements/fixes between them.

This PR fixes that by creating a base Objective-C processor that contains the shared code. View specific code is implemented in the respective subclasses for the views. Although there is very little view specific code for each.

I think something similar would be worth doing for the [Objective-C workflow](https://github.com/Vector35/workflow_objc) as I believe things like the `objc_msgSend` workflow stuff is being shared/copied and pasted from there. However I'm not sure how to do that myself as I'm unsure the desired coding structure because it would probably require that plugin to be imported into this API repo.


plafosse:
This is great I was asking @0cyn to do this just the other day 🥇

emesare:
Rebased version here with changes to allow building on all platforms:

https://github.com/Vector35/binaryninja-api/commit/5a1ab53e77d110145a940c3416721b24b96af06b

emesare:
This PR exists on the https://github.com/Vector35/binaryninja-api/tree/test_shared_cache_stuff_prime_0 branch.

fuzyll:
Should be addressed in builds >= 5.0.7148-dev.

Issue #67, opened 2015-11-23
keyboard navigation improvement (press enter selecting cross-reference should jump to that location)


psifertex:
See subject.

rlc4:
Some suggestions for keyboard shortcuts:

arrow-up - up one line of asm
arrow-down - down one line of asm
cmd/meta-arrow-up - to calling function (assuming only one available)
cmd/meta-arrow-down - follow currently selected jmp / call 
n - name selected symbol / address / function
N - name/rename current function
cmd/meta +/- - Zoom view


Commands already found:
H - view in Hex Editor
; - add comment / note
N - rename symbol (function)  (I suggest limiting this to function naming / renaming, use lower n for a general renaming, or the other way around if people have already started using N as rename of currently selected item)
P - create function at symbol

D0ntPanic:
Fixed in dev

PR #6302, opened 2025-01-04
[SharedCache] Inline `CFString`s

WeiN76LQh:
Basically just copied from the [workflow objc plugin](https://github.com/Vector35/workflow_objc) with [PR #67](https://github.com/Vector35/workflow_objc/pull/67) applied.

Ideally this would not be a copy and these 2 codebases would share the code but its unclear if Vector35 would want to import the Objective-C workflow plugin into the BN API repo and what that would look like. So this is an intermediary solution to get `CFString` inlining working in DSC.

### DIFF
--- a/view/sharedcache/workflow/SharedCacheWorkflow.cpp
+++ b/view/sharedcache/workflow/SharedCacheWorkflow.cpp
@@ -515,14 +515,68 @@ void fixObjCCallTypes(Ref<AnalysisContext> ctx)
 }
 
 
+void inlineCFStrings(Ref<AnalysisContext> ac)
+{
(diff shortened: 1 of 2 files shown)


0cyn:
While I'd eventually like to add functionality used in workflow-objc here, we're looking to use a different approach for CFString inlining and other objc representations. (see pseudo-language representation stuff added in latest major release). 

The CFString inlining approach used in the old workflow was very hacky and it's actually disabled currently due to causing a lot of problems with IL output since we were misusing intrinsic pretty badly. I think the current variable naming approach should be sufficient for now.

Thank you for this PR though!

PR #5773, opened 2024-07-19
impl rust Collaboration and Remote API

rbran:
Unimplemented and other notes:
* `BNRemoteFileDownload` - Returns the c++ `std::vector` from the core API, but don't provide any way to free it.
* `BNCollaborationSnapshotDownloadSnapshotFile` - Returns the c++ `std::vector` from the core API, but don't provide any way to free it.
* `BNRemoteRequest` - Parameters are complex c++ types, and no direct way to manage it.
* `BNAnalysisMergeConflictGetPathItem` - Implementation with downcast to u64 is unclear.
* `Remote::connect` - Not fully implemented because it depends on [enterprise](https://github.com/Vector35/binaryninja-api/pull/5430) and [SecretsProvider](https://github.com/Vector35/binaryninja-api/pull/5630).
* `Remote` - Methods in python preemptively calls `pull_files`/`pull_folders`/`pull_user_permissions`/`open`, but C++ don't. Python was followed.
* `RemoteProject::upload_new_file` - Was ported from `remotebrowser.cpp`, this file is not public, so I decided not to implement it.

PR #5465, opened 2024-05-27
Implement python HLIL related functions

rbran:

PR #6309, opened 2025-01-11
Rust bindings refactor

### LINKED ISSUE #6294: Add `binaryninjacore-sys` to online docs
Currently, https://dev-rust.binary.ninja hosts the rust api docs, but not the `binaryninjacore-sys` docs. Tagging the `-sys` crate with the correct `html_root_url` tag and uploading it under the same domain would yield major benefits to rust users. For one, it would allow to see the auto generated bindings without building them locally, and second, it would automatically link type aliases of the wrapper crate to the original source. That way enum values and the likes can be inspected without, again, building the docs for the `-sys` crate locally.

**Are there any known examples of people using this API/concept?**
All that needs to be done is to add the `binaryninjacore-sys` crate to the list of crates to be built for docs and uploaded to the same html root folder that is also used to serve the `binaryninja` crate docs. The 2 crates do not need to have their docs built at the same time for this to work.

emesare:
Opening this PR for public discussion. This needs more testing and also lacks a few major changes that make sense to land alongside the already existing set of breaking changes.

The Rust bindings are being heavily refactored for the 4.3 release. This is to address many of the shortcomings and historical issues, namely around maintaining and safety.

Some of the additions of this branch include:
- Revised project structure
- Consistent formatting (with `rustfmt`)
- Many UB fixes
- Unit tests
- More/updated documentation 
- Real rust examples (i.e. `cargo run --example`)
- Many of the already opened PR's
- Additional APIs
- Updated and idiomatic linking to the core
- Consolidation of API's (removal of historical API's)
- Rust main thread handler
- Type driven API's (ex. `RegisterId`)
- Updated CI

# Reviewer Guide
**TODO**
- Explain the CI and add a bunch of links to relevant GitHub Actions documentation
- Explain the project structure (this isn't really necessary for consumers)
- Explain the usage of Ref
- Explain the usage of BnString (and strings in general)
- Explain how to review PR's
- Explain how to run tests


emesare:
Documentation needs to be updated to reflect that linking to the core will no longer be optional for plugins. This should be done proactively before we start versioning, because once that is done there _will_ be a strict requirement to depend directly on the `-sys` crate.

emesare:
Mentioning people who have code that will be moved with this PR, if you have unmerged code (i.e. a branch with fixes) please let me know and I can try to incorporate the changes alongside this PR. Otherwise you shouldn't need to do anything.

@CouleeApps The pdb import plugin was moved to `/plugins`
@negasora The dwarf plugins were moved to `/plugins`
@rbran The idb import plugin was moved to `/plugins`

emesare:
This PR will also close the following:
- https://github.com/Vector35/binaryninja-api/pull/6243
- https://github.com/Vector35/binaryninja-api/pull/5898
- https://github.com/Vector35/binaryninja-api/pull/5671
- https://github.com/Vector35/binaryninja-api/pull/5584
- https://github.com/Vector35/binaryninja-api/pull/5431
- https://github.com/Vector35/binaryninja-api/pull/5465
- https://github.com/Vector35/binaryninja-api/pull/6257
- https://github.com/Vector35/binaryninja-api/pull/5773

mkrasnitski:
How come the jobs are duplicated? Seems like `push` and `pull_request` workflows would be mutually exclusive.

emesare:
> How come the jobs are duplicated? Seems like `push` and `pull_request` workflows would be mutually exclusive.

They aren't mutually exclusive, we might want to restrict `push` to just the main branch, not entirely sure what the best way to go about it is.

emesare:
We need to add a doc file explaining the core linkage more in depth, and how to control the ABI version used independently of the rust API. 

I added an example here https://github.com/emesare/simple_rust_binja/blob/master/Cargo.toml but obviously needs to be expanded on and explained.

We should also have a simple "best practices" that is a TLDR, but basically boils down to, please pin your code to a commit or branch, or something other than "dev".

emesare:
As for can we document what API's are going to be changing, I don't have an exhaustive list, but in the case of a user wanting to know if they should use the rust api for a certain task, over lets say, C++ or python, It would be nice to have.

For an example: User wants to use the flow graph API, however there is not a consumer of that necessarily (other than dwarfdump) so the API might change once there is a consumer. 

The easiest way to determine if an API is subject to many changes is to see if there is a consumer. If there is no consumer we could document, either at the module level, or in some documentation that said API would be "unstable". I don't _really_ like doing it the other way around, that is, saying that an API is stable.

emesare:
The structure grouper is failing for the pdb import plugin, I am not entirely sure what I did to cause that, but the effect in a real binary is palpable.

Test: https://github.com/Vector35/binaryninja-api/actions/runs/12780281192/job/35626277295

The issue in a real binary:
image

Left is before, and right is this branch

emesare:
We have a real issue with code coverage in the rust api, a large portion of the api is not used at all.

The solution is unit tests and more examples, which is what this PR is working towards. However measuring that is difficult, we might want to look into including some code coverage tool in CI.

- https://lib.rs/crates/cargo-llvm-cov
- https://github.com/xd009642/tarpaulin
- https://github.com/mozilla/grcov

emesare:
What is stopping this PR from being merged?

1. Some tests are failing for pdb import
2. Need to run through some binaries again to make sure there are no regressions performance-wise, this was done after some changes to the architecture api but has not been done again

emesare:
> As for can we document what API's are going to be changing, I don't have an exhaustive list, but in the case of a user wanting to know if they should use the rust api for a certain task, over lets say, C++ or python, It would be nice to have.
> 
> For an example: User wants to use the flow graph API, however there is not a consumer of that necessarily (other than dwarfdump) so the API might change once there is a consumer.
> 
> The easiest way to determine if an API is subject to many changes is to see if there is a consumer. If there is no consumer we could document, either at the module level, or in some documentation that said API would be "unstable". I don't _really_ like doing it the other way around, that is, saying that an API is stable.

For this I think both a little foreword in the README.md explaining what API's are actively being used, and therefor more reliable and better suited for consumers that are not comfortable with API's that weren't tested/or API's with the possibility of breaking changes. Aside from the README.md we might also want to add a warning on modules like MLIL and HLIL that they are subject to breaking changes.

The goal for this is to make it clear for those that do not want the API to change as frequently, if rust fits the bill, or if they should possibly use C++/Python api instead.

emesare:
Are there any more planned API changes for this PR?

No, I don't have anything else planned, and the changes that are occurring are on functions that previously had _no_ visible consumer.

emesare:
GitHub has no easy way to set deployment retention, this is insane behavior.

https://github.com/Vector35/binaryninja-api/deployments/testing

Every push to the rust code will trigger this, we might need to setup _another_ workflow to run on some arbitrary schedule to delete old deployments.

emesare:
image

Left is this PR, right is current dev. Any match that is traversing the IL fails. (Because the IL traversal is borked)

At some point during the last few days, IL traversal has been borked. Need to fix that obviously before this is merged.

emesare:
Next priority (due to the above) is to add LLIL unit tests. We need to have the `visit_tree` functions tested thoroughly to prevent regressions.

Most likely the WARP is not visiting the expressions, so first look at the `visit_tree` set of functions.

emesare:
> Next priority (due to the above) is to add LLIL unit tests. We need to have the `visit_tree` functions tested thoroughly to prevent regressions.
> 
> Most likely the WARP is not visiting the expressions, so first look at the `visit_tree` set of functions.

This has been fixed with https://github.com/Vector35/binaryninja-api/pull/6309/commits/2b7557dee756b76f034cf00645585936f958e03e

emesare:
Noting that we need to move the mlil and hlil modules to their respective full name counterparts

emesare:
We still need more coverage of the llil api, mostly regarding the visitors.

The MLIL and HLIL api's I want to mark as unstable in the docs because I will need a second pass for those, I don't have time to rework them. Specifically so that we can rewrite MLIL and HLIL functions.

The LLIL lifter needs to be split apart and also have the Lifted restriction removed, we can emit regular LLIL through function workflows, meaning we also probably need to rename that part of the LLIL API to something other than "lifting", I am partial to something like "emitter" or something.

emesare:
> Are there any more planned API changes for this PR?
> 
> No, I don't have anything else planned, and the changes that are occurring are on functions that previously had _no_ visible consumer.

Clearly this is no longer the case :/

I wish I could stop changing the API but I know it will be a while before we can break API's again, so...

emesare:
I need to add a plugin example, I just don't know where to put it. I really dislike using the "examples" directory for something that isnt actually a rust example. I think a plugin_example subcrate (same directory placement of binaryninjacore-sys) would be the most appropriate. We need **_ONE_** plugin example, not 10. Moving forward we write application examples, people want examples they can just run, not something that requires them to build it and open binary ninja and figure out what the example is trying to do. They need to be self contained.

emesare:
We should also do a once over of the old rust_break_everything branch to make sure there wasn't any missing commits that were not rolled up into this branch.

emesare:
On the topic of imports:

- I am going to normalize all module names (i.e. `debuginfo` will become `debug_info`)
- I am _not_ going to add a prelude module

For why I am deciding against adding a prelude see [this reddit post](https://www.reddit.com/r/rust/comments/q6stlt/massive_preludes_why/?chainedPosts=t3_a7pcp2). Namely, we have very generic names that would end up existing within the prelude, things such as `Type` and `Database`, either we assume that our library is the only library a large portion of users will import with those names, in which case a prelude would be fine, _or_ we don't make that assumption.

If anyone disagrees with this please say something!

emesare:
I don't know if `rstest` can have fixtures depend on other fixtures, but we need to have the session available at the time of binary view initialization. I _think_ that is what is occurring in CI.

emesare:
Before this is merged we should add a migration guide somewhere, most likely in the body of the first PR message.

We need to wrap this up this week, there are a few downstream items that depend on this, namely WARP (see: type containers / type archives)

We should put a pin in any further breaking changes and document the modules that are going to be changed in future releases (after next stable).

rbran:
One suggestion is to remove the generic `A: Architecture` from the `CallingConvention`: https://github.com/Vector35/binaryninja-api/compare/rust_cleanup_0...rbran:binaryninja-api:cc-core-arch?expand=1

For two main reasons:

First the implementation seems to only use CoreArchitecture by default anyways, so a custom architecture is never specified, adding a complexity with no benefits. Also the `CallingConvention` struct don't make use of the function `BNGetCallingConventionArchitecture`. 

Second to allow the user to define one Custom-CallingConvention and implement it to multiple architectures. The currently implementation only allow one `CallingConvention` (struct) per architecture, because `CallingConventionBase` define the `Arch` as an associated type

rbran:
Another good to have feature is to `BinaryViewReader` implementing Read correctly. The [`idb-import`](https://github.com/Vector35/binaryninja-api/blob/da1b04d09832d3a827fd4d97ccdc6818c59b48d8/plugins/idb_import/src/lib.rs#L76) plugin creates a custom implementation of `std::io::Read` for `BinaryView` instead of using `BinaryViewReader` because it don't implement `SeekFrom::End` correctly.

This can be fixed by adding a function that returns the `BinaryView` from the `BinaryViewReader` (or at least the size of it) is added to the core API, like this:

```
pub fn BNGetBinaryReaderView(stream: *mut BNBinaryReader) -> *mut BNBinaryView;
// or at least the size of the view
pub fn BNGetBinaryReaderBinaryViewLen(stream: *mut BNBinaryReader) -> u64;
```

The same applies to `BinaryViewWriter`

emesare:
@rbran both of those suggestions seem completely reasonable, I will look into integrating both with this PR. Thank you!

emesare:
Rolling up this PR is gonna be unfortunate from a git blame perspective but that was always going to be the case.

All contiguous "More rust cleanup" commits will most likely be squashed with an incrementing number.

- Rust refactor
- Some commit
- Rust refactor 2
- Another commit
- Rust refactor 3

That or we just squash this PR down to 1 commit. I don't really mind either.

emesare:
@rbran does this look reasonable for reusing a calling convention across architectures?

https://github.com/Vector35/binaryninja-api/pull/6309/commits/b1693fd81fb94c512e8e18da187d273b2cbbef67

emesare:
```
test struct_grouper::test_dispatcher_header ... FAILED
test struct_grouper::test_bool_modifier ... FAILED
```

These were failing prior to my changes. However there still seems to be some regression when running through the python tests regarding the PDB import plugin.

rbran:
@emesare In my view yes, You implement a CC, then register it for each architecture.

eg: 
```rust
let my_cc = MyCC::new();
let my_cc_x86 = register_calling_convention(arch_x86, "MyCC-x86", my_cc.clone());
let my_cc_arm = register_calling_convention(arch_arm, "MyCC-arm", my_cc);
```

Although it depends on how binja understand the "cc-arch" associations. Maybe it requires a unique CC per arch somewhere inside.

emesare:
> Although it depends on how binja understand the "cc-arch" associations. Maybe it requires a unique CC per arch somewhere inside.

The cc provides the register id so as long as the arch keeps the same register id mapping everything should be ok.

All of the same checks are performed that were done prior to this change so while in theory you could receive a bad register mapping, you cannot produce a bad mapping (i.e. rust api will prohibit you from giving the core a bad register id in your CC)

emesare:
The last known bug has been fixed with https://github.com/Vector35/binaryninja-api/pull/6309/commits/2c378990dd92d536dc2623d0ae1ab9c51962841c

So the next steps as I see it are:
1. Figure out why the view is failed to be created in CI but not when ran locally
2. Performance testing
3. More documentation?

This PR will be wrapped up by week's end. So if anyone has any go backs now is the last chance.

emesare:
@rbran I made the changes you suggested in https://github.com/Vector35/binaryninja-api/pull/6309/commits/16e8ad422e7e25d4723869b1264ccae9c5071efb let me know if that covers your use case. I also changed it for the BinaryWriter as well.

emesare:
Because I have tomorrow and Friday to finish this up I went ahead and pulled in https://github.com/Vector35/binaryninja-api/pull/5773, I think I should have enough time to refactor it and add some tests and examples.

rbran:
@emesare Thanks that solves my use case.

BTW I hope in the future we can add a function to the coreAPI to recover the `BinaryView` from a `BinaryReader`/`BinaryWriter`, this allow rust to implement `#[repr(transparent)]` for `BinaryReader/Writer` to `BNBinaryReader/Writer`. But that's a task for other PR.

emesare:
The rolled up commit will be on https://github.com/Vector35/binaryninja-api/tree/rust_refactor

emesare:
Community rust plugins to update:
- https://github.com/starfleetcadet75/binaryninja-superh
- https://github.com/unknowntrojan/binja_coolsigmaker

emesare:
Merged with https://github.com/Vector35/binaryninja-api/commit/6a63c17853009ce7a0071458e39c140a09440230

blacktop:
@emesare are you going to add a load with progress func for the headless session?

emesare:
> @emesare are you going to add a load with progress func for the headless session?

That sounds like a good idea, I'll add that today! Thank you.

emesare:
@blacktop added with https://github.com/Vector35/binaryninja-api/commit/98276031969e67229d6cc6344f1b9debdc211e8c

I should note that it appears the core only calls that progress callback for loading BNDBs, IIRC there was a comment somewhere about that in the rust code. I could not find that comment so I added it to the respective functions with https://github.com/Vector35/binaryninja-api/commit/0263957dc920a10785af6874ab3040a529547d70

blacktop:
Thank you! 

I know this is most likely an inappropriate place to ask, but I'm trying to understand how to run a plugin via the Rust API (BinExport) specifically.  Are there any examples anywhere on how to do this? I can also ask on Slack if that's a better place?

emesare:
No problem, by run you mean run a command that BinExport registered?

Like one of these:
image

If so than we would need to add the rust equivalent of `PluginCommand` https://github.com/Vector35/binaryninja-api/blob/8767400b712f12f459556844e27d9b9f5ff037bd/binaryninjaapi.h#L14146-L14168

In rust we can register commands but we cannot call them yet.

I unfortunately have a few things in front of adding that, specifically I need to implement the BinaryDataNotification API. If you would like to make an issue for fleshing out plugin commands to support calling them from rust, I can assign it to myself.

emesare:
Also this is assuming the plugin command that you want to call does not then invoke some user interaction, that would require us to also add the interaction callback API.

There is a TODO for that in the flow graph example https://github.com/Vector35/binaryninja-api/blob/6a63c17853009ce7a0071458e39c140a09440230/rust/examples/flowgraph.rs#L52-L53

blacktop:
@emesare I guess this is how it currently works in python? https://infosec.exchange/@binaryninja/113827475796582876

blacktop:
> No problem, by run you mean run a command that BinExport registered?
> 
> Like one of these: image
> 
> If so than we would need to add the rust equivalent of `PluginCommand`
> 
> https://github.com/Vector35/binaryninja-api/blob/8767400b712f12f459556844e27d9b9f5ff037bd/binaryninjaapi.h#L14146-L14168
> 
> In rust we can register commands but we cannot call them yet.
> 
> I unfortunately have a few things in front of adding that, specifically I need to implement the BinaryDataNotification API. If you would like to make an issue for fleshing out plugin commands to support calling them from rust, I can assign it to myself.

Ah I see, thank you for this explanation. I will wait for PluginCommand running to be added to the Rust API. Thanks again!

emesare:
Yea that would be through the `PluginCommand`, if this is something you need now than you can call `BNGetAllPluginCommands` yourself until I or someone else can get to it.

Here are the issues:
- https://github.com/Vector35/binaryninja-api/issues/6362
- https://github.com/Vector35/binaryninja-api/issues/6361 (not required to solve your use case)

PR #6310, opened 2025-01-12
Fix clippy lints

mkrasnitski:
Ran `cargo clippy` with default lints. A few small cleanups included as well.

### DIFF
--- a/rust/build.rs
+++ b/rust/build.rs
@@ -5,7 +5,7 @@ fn main() {
     let _ = std::fs::create_dir("target/doc");
     let _ = std::fs::copy("../docs/img/favicon.ico", "target/doc/favicon.ico");
     let _ = std::fs::copy(
-        "under_construction.png",
+        "assets/under_construction.png",
         "target/doc/under_construction.png",
     );
     let _ = std::fs::copy("../docs/img/logo.png", "target/doc/logo.png");
--- a/rust/clippy.toml
+++ b/rust/clippy.toml
@@ -1,0 +1,1 @@
+too-many-arguments-threshold = 8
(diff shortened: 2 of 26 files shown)


rust/src/architecture.rs
emesare, line 1: I find the before to be more readable

rust/clippy.toml
emesare, line 1: we can just clippy::allow the too many arguments lint in the lib.rs

rust/build.rs
```diff
@@ -5,7 +5,7 @@ fn main() {
     let _ = std::fs::create_dir("target/doc");
     let _ = std::fs::copy("../docs/img/favicon.ico", "target/doc/favicon.ico");
     let _ = std::fs::copy(
-        "under_construction.png",
+        "assets/under_construction.png",
```
emesare, line 8: Whoops forgot about that one! Thanks

emesare:
Merged with https://github.com/Vector35/binaryninja-api/pull/6309/commits/3ea9c4eb4eef2949b2fc41065cb00f187e3d564d thank you! I also went ahead and added clippy to the CI, it should only warn however. Mostly just good for visibility in PR's!

rust/src/architecture.rs
mkrasnitski: yeah, it would be best if the definition of `BNInstructionInfo` was the following instead:

```rust
pub struct BNInstructionInfo {
    pub length: usize,
    pub branchCount: usize,
    pub archTransitionByTargetAddr: bool,
    pub delaySlots: u8,
    pub branchInfo: [BNBranchInfo; 3usize],
}

pub struct BNBranchInfo {
    pub branchType: BNBranchType,
    pub branchTarget: u64,
    #[doc = " If null, same architecture as instruction"]
    pub branchArch: *mut BNArchitecture,
}
```

Then we could iterate over all the branch info at once instead of by-index. This would require changing the C++ definition though.

PR #6311, opened 2025-01-12
MIPS r5900 (PS2 EE) support

0cyn:
This additionally oneshots delay slot warnings as they were causing severe UI hangs in LogView on certain binaries that triggered the warning frequently. 

Not especially committed to that change, it was just on my local copy and seemed worth including.


plafosse:
@0cyn We're going to need a test file for this so we can add it to the unit tests

Souzooka:
This MR does not seem to properly handle R5900's multiplication instructions during lifting (and likely other caveats involving the hi/lo registers). The biggest issue involving `mult` is that since the MULT IL is untouched, the extended destination register is not set when a multiplication occurs.

PR #6315, opened 2025-01-14
[SharedCache] Simplify MMappedFileAccessor::Read* methods

bdash:
This change is mostly motivated by simplifying the code, but it also brings minor correctness and performance benefits.

1. The pointer returned by `mmap` is stored as a `uint8_t*` rather than `void*` as that is how it is used. This eliminates a bunch of casts.
2. Read methods for primitives delegate to a new `Read` template function that in turn delegates to the general-purpose `Read(void* dest, size_t address, size_t length)`. This improves the consistency of bounds checking and simplifies the code. The compiler is more than willing to inline this so we get less repetition and consistent bounds checking with no additional overhead.
3. `ReadNullTermString` now uses `std::find` to find the NUL byte and directly constructs the string from that range of bytes. This removes an unnecessary allocation that was previously being forced by the use of `reserve` followed by `shrink_to_fit`. It also avoids repeated reallocation for longer strings as they grew past the the reserved size as they were being built up a character at a time.

### DIFF
--- a/view/sharedcache/core/VM.cpp
+++ b/view/sharedcache/core/VM.cpp
@@ -149,12 +149,12 @@ void MMAP::Map()
 		return;
 	}
 
-	_mmap = MapViewOfFile(
+	_mmap = static_cast<uint8_t*>(MapViewOfFile(
 		hMapping,                    // handle to the file mapping object
 		FILE_MAP_COPY,         		 // desired access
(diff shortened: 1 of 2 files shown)


bdash:
This was cherry-picked into dev as c9bbe933dd53fff5d54c2e4f218e4a89f6c58fcd.

PR #6318, opened 2025-01-14
[SharedCache] Improvements and fixes for `MMappedFileAccessor` stuff

WeiN76LQh:
`MMappedFileAccessor::Open` wasn't really behaving as I think it was intended. It tried to use `fileAccessorSemaphore` to limit the number of concurrent files Binary Ninja could open. It didn't really confirm it had acquired a reference on `fileAccessorSemaphore` but then would always release one back when a `MMappedFileAccessor` was deallocated. This could actually inflate the count of `fileAccessorSemaphore` meaning it could end up with a higher file limit than originally set. Also using a worker to deallocate a `MMappedFileAccessor` on another thread felt like a bit of a gross hack and made synchronizing dropping a ref and acquiring one on `fileAccessorSemaphore` harder than it needed to be. This resulted in a bit of a rewrite of `MMappedFileAccessor::Open` to mitigate these issues. It now should respect the file limit set on startup but can go over it temporarily in extreme circumstances that I don't expect to occur in real world experiences.

Additionally there seemed to be some insufficient use of locking to synchronize accesses to certain data structures. I believe this commit has cleaned those up. They were causing very intermittent crashes.

### DIFF
--- a/view/sharedcache/core/VM.cpp
+++ b/view/sharedcache/core/VM.cpp
@@ -52,13 +52,36 @@
 
 void VMShutdown()
 {
-	std::unique_lock<std::mutex> lock2(fileAccessorsMutex);
-	std::unique_lock<std::mutex> lock(fileAccessorDequeMutex);
-
 	// This will trigger the deallocation logic for these.
 	// It is background threaded to avoid a deadlock on exit.
-	fileAccessorReferenceHolder.clear();
-	fileAccessors.clear();
+
+	// Do not grab both locks at the same time otherwise a deadlock will occur 
+	// due to `fileAccessorReferenceHolder.clear()` triggering 
+	// `SelfAllocatingWeakPtr<MMappedFileAccessor>` deallocation routines that 
+	// need to acquire `fileAccessorsMutex`.
+	{
+		// Some additional complexity with this lock is that dropping all these references likely 
+		// will cause the final reference to a DSC binary view to be dropped. In doing so 
+		// `~DSCView` will call `MMappedFileAccessor::CloseAll` which requires the same lock. This 
+		// occurs on the same thread, resulting in it trying to acquire a lock it already has. 
+		// This solution copies all the deques from `fileAccessorReferenceHolder`, therefore 
+		// keeping all the references alive temporarily. Then `fileAccessorReferenceHolder` can be 
+		// cleared and the `fileAccessorDequeMutex` lock can be dropped. When exiting this scope 
+		// the desctructor for `accessorDequesToDrop` will drop all the `MMappedFileAccessor` 
+		// references whilst holding no locks.
+		std::unique_lock<std::mutex> lock(fileAccessorDequeMutex);
+		std::vector<std::deque<std::shared_ptr<MMappedFileAccessor>>> accessorDequesToDrop;
+		for (auto& [_, fileAccessorDeque] : fileAccessorReferenceHolder)
+		{
+			accessorDequesToDrop.push_back(fileAccessorDeque);
+		}
+		fileAccessorReferenceHolder.clear();
+		lock.unlock();
+	}
+	{
+		std::unique_lock<std::mutex> lock(fileAccessorsMutex);
+		fileAccessors.clear();
+	}
 }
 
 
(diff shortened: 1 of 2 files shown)


bdash:
As an FYI, https://github.com/Vector35/binaryninja-api/pull/6316 also changes `MMappedFileAccessor::Open` and will conflict with this. It shouldn't be too bad to resolve as none of the behavior it changes is relevant to what you're fixing here.

WeiN76LQh:
Yes I hadn't seen your PRs until after I pushed this one. As you state it doesn't really conflict with any of your changes. If an updated PR is required in the future then I can do that.

WeiN76LQh:
I've identified another bug in this area of code to do with the [`blockedSessionIDs`](https://github.com/Vector35/binaryninja-api/blob/362a420eb027bd1de0a172f1823ba7e4dd806a3a/view/sharedcache/core/VM.h#L123) variable. If a user opens DSC using `Open Selected Files With Options...` and then changes a setting (I think any setting) and then clicks `Open`, in the background a `DSCView` is destroyed and a new one is created. The destruction of the first one calls [`MMappedFileAccessor::CloseAll`](https://github.com/Vector35/binaryninja-api/blob/362a420eb027bd1de0a172f1823ba7e4dd806a3a/view/sharedcache/core/VM.cpp#L260) which adds the current session ID to the [`blockedSessionIDs`](https://github.com/Vector35/binaryninja-api/blob/362a420eb027bd1de0a172f1823ba7e4dd806a3a/view/sharedcache/core/VM.h#L123) list. However the newly created `DSCView` has the same session ID so during initial load it is constantly creating and destroying `MMappedFileAccessor`s due to [this check](https://github.com/Vector35/binaryninja-api/blob/362a420eb027bd1de0a172f1823ba7e4dd806a3a/view/sharedcache/core/VM.cpp#L247) preventing `MMappedFileAccessor`s being added to [`fileAccessorReferenceHolder`](https://github.com/Vector35/binaryninja-api/blob/362a420eb027bd1de0a172f1823ba7e4dd806a3a/view/sharedcache/core/VM.h#L122). The performance degradation is pretty crazy, causing the initial load to take minutes. During which Binary Ninja is unresponsive so its likely for the average user they will just kill the process after a bit of time of seemingly nothing happening.

I would provide a fix in this PR via an additional commit but I'm unsure how to proceed as I don't fully understand the mechanics of everything in this context and the levers that can be worked with.  [`blockedSessionIDs`](https://github.com/Vector35/binaryninja-api/blob/362a420eb027bd1de0a172f1823ba7e4dd806a3a/view/sharedcache/core/VM.h#L123) existence feels like a hack to solve a problem which I don't really understand exactly how it exists. If this check is still required then another solution is going to be needed otherwise I guess the code can just be removed?

bdash:
There's another issue with how `MMappedFileAccessor::Open` works / is used. Every call to it takes in a `postAllocationRoutine` function. This is used by `SharedCache` to apply the slide when the file is opened. But `MMappedFileAccessor::Open` caches the _first_ lazy accessor (and its `postAllocationRoutine`) created for any given path. If the first lazy accessor does not include a `postAllocationRoutine`, no slide will be applied if the `MMappedFileAccessor` is closed and subsequently reopened. `SharedCache::PerformInitialLoad` never specifies a `postAllocationRoutine` when opening files, and it touches most (all?) of the files in the shared cache.

In practice the file descriptor limit seems high enough that files for a single shared cache tend to stay open indefinitely, avoiding this problem. I guess if you open several iOS shared caches with macOS's default low file descriptor limit you might hit the problem.

I was able to trigger it by hard-coding a file descriptor limit of 8 and opening an iOS shared cache. I can see files being closed then reopened, with no slide being applied.

Perhaps the approach to file accessor caching should be revisited?

A lot of the hoops that the existing code jumps through to try to limit the number of open files could be avoided by using `setrlimit(RLIMIT_NOFILE, …)` to bump the soft file descriptor limit to a more reasonable value. It may still be desirable to close files that haven't been accessed recently, but that would become a memory optimization rather than being necessary for correctness.

bdash:
I sent https://github.com/Vector35/binaryninja-api/pull/6392 which removes the complex synchronization in `MMappedFileAccessor::Open` rather than trying to fix it. It solves the underlying issues by:
1. Using `setrlimit` to increase the file descriptor limit.
2. Using an LRU cache to impose a soft limit on the number of open `MMappedFileAccessor` objects. This is primarily a memory optimization since closing the accessors lets the dirty pages corresponding to those files be discarded.
3. Allowing caching of file accessors opened by sessions corresponding to views that have been closed (to avoid the issue mentioned in a comment above: https://github.com/Vector35/binaryninja-api/pull/6318#issuecomment-2593981486).

WeiN76LQh:
Closing because this is no longer relevant after #6392 was merged

PR #6321, opened 2025-01-15
[SharedCache] Optimize parsing and applying of slide info

bdash:
Slide info is parsed and applied on the main thread, below the `SharedCache` constructor, when a shared cache is opened. Time spent applying the slide is time that the main thread is blocked.

These changes eliminate some unnecessary overhead so what work remains is dominated by kernel work (paging in data, copying pages when the first modification is made).

The changes are:
1. Read slide pointers via `ReadULong` rather than the variable-length `Read` method. The compiler isn't able to eliminate the call to `memcpy` in the variable-length `Read` function, and the function call overhead is noticeable given the small size of the read and number of times it is called.
2. Remove the file member from `MappingInfo`. Slide info is applied for a single file at a time so there's no reason to track the file it belongs to. This removes unnecesssary reference counting on the `std::shared_ptr` that holds the `MMappedFileAccessor`.

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -1078,7 +1078,6 @@ void SharedCache::ParseAndApplySlideInfoForFile(std::shared_ptr<MMappedFileAcces
 		MappingInfo map;
 
 		file->Read(&map.mappingInfo, baseHeader.mappingOffset + sizeof(dyld_cache_mapping_info), sizeof(dyld_cache_mapping_info));
-		map.file = file;
 		map.slideInfoVersion = slideInfoVersion;
 		if (map.slideInfoVersion == 2)
 			file->Read(&map.slideInfoV2, slideInfoOff, sizeof(dyld_cache_slide_info_v2));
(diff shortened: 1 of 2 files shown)


bdash:
Updated after 27eb30dcfcb5f897aed7c7b128e7c50a2020a6a1 removed use of the `rewrites` vector as the earlier version of this change did.

emesare:
Fixes for this are included on the https://github.com/Vector35/binaryninja-api/tree/test_shared_cache_stuff_prime_0 branch.

We no longer have a shared_ptr to the accessor but I did not swap out the read, so I went ahead and did that: https://github.com/Vector35/binaryninja-api/commit/8d9dd61c605ec509df8deb9453872e665fa88c53

https://github.com/Vector35/binaryninja-api/blob/8d9dd61c605ec509df8deb9453872e665fa88c53/view/sharedcache/core/SlideInfo.cpp#L48

https://github.com/Vector35/binaryninja-api/blob/8d9dd61c605ec509df8deb9453872e665fa88c53/view/sharedcache/core/SlideInfo.cpp#L85

PR #6358, opened 2025-01-24
bugfix: ARM64: missing break in il.cpp's switch

themaks:
A `break` was missing in the `switch`, making the `LDXRH` instruction generate 2 IL instructions instead of 1.

I did not actually encountered the bug dynamically nor tested the fix, but it all seems pretty straightforward.

Cheers


galenbwill:
Thanks for the PR! Merged to dev, will go out in a future dev build, version >= 4.3.6848

PR #6365, opened 2025-01-27
Fix rust EnumerationBuilder method return types

VisualEhrmanntraut:
methods `append`, `insert`, `replace`, and `remove` were returning `&Self` instead of `&mut Self`, but accepting `&mut Self`, so you could add/remove a member only once.

### DIFF
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -1217,7 +1217,7 @@ impl EnumerationBuilder {
         unsafe { Enumeration::ref_from_raw(BNFinalizeEnumerationBuilder(self.handle)) }
     }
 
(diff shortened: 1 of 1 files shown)


emesare:
Thanks for noticing that, we only ever used them unchained so never ran into that. I'll add some more unit tests today for this and other builders

VisualEhrmanntraut:
Hey @emesare, I'm in the BN slack, I've got an inquiry about the Rust API other than this issue. Can you please check the rust-api channel? I would appreciate it.

VisualEhrmanntraut:
As in, another problem with the API, not a question

PR #6402, opened 2025-02-08
[SharedCache] Introduce a SymbolInfo type

bdash:
Previously `SharedCache` used a mixture of `pair<uint64_t, pair<BNSymbolType, std::string>>` and `Ref<Symbol>` to represent a symbol. The former is hard to understand due to the nesting and lack of names. The latter is expensive to create and its getters are expensive due to the multiple allocations + `strlen` calls performed when converting between core's string representation and `std::string`.

`SymbolInfo` wraps the same data that the nested pairs contained in a friendly struct with named fields. Beyond that it, it exposes an `AsSymbol` function to construct a `Ref<Symbol>` where that is what is required.

Code within `SharedCache` is updated to work with `SymbolInfo` until it needs to call into a core API that requires a `Symbol`.

This change is only a small performance improvement at the moment. It's more noticable once `SharedCache` no longer copies so much of its state (#6393).

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -57,10 +57,8 @@ int count_trailing_zeros(uint64_t value) {
 
 struct SharedCache::State
 {
-	std::unordered_map<uint64_t, std::vector<std::pair<uint64_t, std::pair<BNSymbolType, std::string>>>>
-		exportInfos;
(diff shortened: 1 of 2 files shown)


bdash:
I'll have to look at whether this is still worth doing given #6197. It may not be necessary.

bdash:
Constructing a `Symbol` is still a little more expensive than I'd like, but the changes on the [test_api_string_ref branch](https://github.com/Vector35/binaryninja-api/tree/test_api_string_ref) make the `Symbol` getters significantly more efficient. That's a better path forward than introducing a custom type.

PR #6403, opened 2025-02-08
[SharedCache] Make MetadataSerializable::Load static

bdash:
`MetadataSerializable` is updated to allow a subclass to optionally specify the return type of `Load`. If not specified it defaults to the subclass itself. `SharedCache` specifies `std::optional` to represent the case where the serialized metadata is in a format or version it does not recognize.

By making `Load` a static member function and having it return a new object it becomes easier to reason about the state of objects when a deserialization failure occurs. It can return `std::optional` to indicate failure and there is no object left in an unknown state.

Prior to this, `Load` worked on an existing instance of an object. It was unclear what state the object would be in if a deserialization failure occurred (its original state prior to `Load`? some partially-loaded state? a null state?). The failure itself also had to be communicated out of band.

This is a prerequisite of #6393.

### DIFF
--- a/view/sharedcache/api/sharedcache.cpp
+++ b/view/sharedcache/api/sharedcache.cpp
@@ -204,9 +204,8 @@ namespace SharedCacheAPI {
 		BNFreeString(outputStr);
 		if (output.empty())
 			return {};
-		SharedCacheMachOHeader header;
-		header.LoadFromString(output);
-		return header;
+
+		return SharedCacheMachOHeader::LoadFromString(output);
 	}
 
 	std::optional<SharedCacheMachOHeader> SharedCache::GetMachOHeaderForAddress(uint64_t address)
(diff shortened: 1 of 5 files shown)


CouleeApps:
FYI this requires /std:c++20 to compile on MSVC due to the use of designated initializers. That's _probably_ fine and we will likely need to be updating the entirety of the binja api to C++20 anyway, but for the moment it's a compile error. You can either rewrite parts of this to not use those, or we can have The Discussion internally for whether that is something we are doing now.

CouleeApps:
On further inspection, seems like Mason already beat me to it in 360160b88f35b331f7fdb67264bb16402e34000b

bdash:
Thanks for pointing that out. It's a little annoying that Clang lets C++20 constructs slide in C++17 language mode without so much as a warning.

bdash:
Looks like this was cherry-picked into `dev` as c20bf355152d9a715dae3c19777c3d4a1046174b.

PR #6409, opened 2025-02-11
[SharedCache] Fix performance loading images from iOS 14 shared cache

bdash:
A logic error in `FindSymbolAtAddrAndApplyToAddr` meant that a code path that would attempt to re-use existing symbol definitions in the view would still fall back to searching the exported symbol list.

`FindSymbolAtAddrAndApplyToAddr` is updated to return early if a symbol already exists at the target location and the target location is the same as the symbol location. Additionally, in the event a symbol was found in the view at the symbol location, it is applied to the target location and then the function returns.

`WillMutateState` is moved after these initial checks so that is will only be called in the codepaths that attempt to locate the symbol amongst the exported symbols. This helps significantly as the copying this state is quite expensive.

With this change it is now possible to load QuartzCore from the iOS 14 shared cache in around 70 seconds. Previously it would run for hours without the load completing.

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -2933,85 +2933,85 @@ Ref<TypeLibrary> SharedCache::TypeLibraryForImage(const std::string& installName
 void SharedCache::FindSymbolAtAddrAndApplyToAddr(
 	uint64_t symbolLocation, uint64_t targetLocation, bool triggerReanalysis)
 {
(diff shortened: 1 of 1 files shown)


bdash:
This was cherry-picked into dev as 798598596352f46856b1d1926e63baac0b8ac084.

emesare:
Thank you for closing the other PR's for me, and thank you for your continued contributions!

PR #6417, opened 2025-02-14
Remove m_reader / m_builder members from AnalysisContext

bdash:
Constructing these objects is very expensive compared to the cost of creating the `AnalysisContext`.

`m_reader` appears to be entirely unused so it is simply removed.

`m_builder` is only used within `Inform`. It may be appropriate to move it to a local variable within that function, but I'm not sure how often `Inform` is called. Given that the `Json::StreamWriterBuilder` is rather expensive to construct I've instead opted to make it static so it is initialized only on first use and then reused by later calls. `Json::StreamWriterBuilder::newStreamWriter` does not mutate any state so this seems ok in terms of thread-safety.

### DIFF
--- a/binaryninjaapi.h
+++ b/binaryninjaapi.h
@@ -10017,8 +10017,7 @@ namespace BinaryNinja {
 	class AnalysisContext :
 	    public CoreRefCountObject<BNAnalysisContext, BNNewAnalysisContextReference, BNFreeAnalysisContext>
 	{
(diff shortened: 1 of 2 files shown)


bpotchik:
Thanks for highlighting this. I ended up swapping jsoncpp for rapidjson, and cleaning up the constructor. The Inform API also is changed to eliminate the std::variant as well.

PR #6432, opened 2025-02-19
[SharedCache] Fix edge cases in handling of exported symbol nodes

bdash:
1. Correctly handle symbols whose address is the start of a section. The address range logic failed to match the appropriate section and so these symbols would end up with their type being determined based on an uninitialized `flags` variable.
2. Handle symbols of kind `EXPORT_SYMBOL_FLAGS_KIND_ABSOLUTE`. These symbols have values that do not point within the current image. There is no way to determine if they point to data or a function.

    In practice the handful of places these are present in the shared cache use these to represent constants rather than real addresses (`_swift_async_extendedFramePointerFlags` with value 0x1000000000000000, `_objc_absolute_packed_isa_class_mask` with value 0x7ffffffffffff8).
    
    I'm not sure what the "best" way to represent these is so for now they're created as symbols with an absolute address.

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -2756,33 +2756,59 @@ void SharedCache::ReadExportNode(std::vector<Ref<Symbol>>& symbolList, const Sha
 		uint64_t flags = readValidULEB128(current, end);
 		if (!(flags & EXPORT_SYMBOL_FLAGS_REEXPORT))
 		{
-			uint64_t imageOffset = readValidULEB128(current, end);
-			if (!currentText.empty() && textBase + imageOffset)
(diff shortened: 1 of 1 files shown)


emesare:
The fix for this is on https://github.com/Vector35/binaryninja-api/tree/test_shared_cache_stuff_prime_0 branch.

Diff:
https://github.com/Vector35/binaryninja-api/commit/ff2f0bfc9a994a8c9bea3e56db11f69c75e905a1#diff-e31b7bdb182131d21f4a254a3c397b79b7dbad50403e80a210cd76678de21d4fR565
Line ref:
https://github.com/Vector35/binaryninja-api/blob/ff2f0bfc9a994a8c9bea3e56db11f69c75e905a1/view/sharedcache/core/MachO.cpp#L554-L599

PR #6436, opened 2025-02-20
[SharedCache] Apply objc_msgSend call type overrides in more places

bdash:
`fixObjCCallTypes` is updated to handle tail calls as well as regular calls.

Additionally, if the selector address is not valid it now looks for the `sel_` symbols that `DSCObjCProcessor::ReadMethodList` adds for selectors whose names reside in regions that are not yet mapped. This allows call type overrides to be applied even when a selector's name is defined in a different image. This is the common case in recent macOS shared caches at least.

### DIFF
--- a/view/sharedcache/workflow/ObjCActivity.cpp
+++ b/view/sharedcache/workflow/ObjCActivity.cpp
@@ -54,16 +54,16 @@ void ObjCActivity::AdjustCallType(Ref<AnalysisContext> ctx)
 
 	const auto rewriteIfEligible = [bv, ssa](size_t insnIndex) {
 		auto insn = ssa->GetInstruction(insnIndex);
-		if (insn.operation != LLIL_CALL_SSA)
+		if (insn.operation != LLIL_CALL_SSA && insn.operation != LLIL_TAILCALL_SSA)
 			return;
 
 		// Filter out calls that aren't to `objc_msgSend`.
-		auto callExpr = insn.GetDestExpr<LLIL_CALL_SSA>();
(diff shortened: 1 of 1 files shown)


bdash:
This works, but Binary Ninja seems to get confused about the state of some functions.

In some cases the override doesn’t appear to take. When I scroll the function containing the override to the right place in the view I see a little progress spinner appear at the top of the view, where the “Reanalyze function” arrow would usually be. It just spins and spins. If I right-click on one of the objc_msgSend calls that has had its type overridden and select “Override Call Type…” it shows the expected overridden type. If I hit Accept the call (and others in the same function) all update to reflect the overrides.

Is this an error in `fixObjCCallTypes`, or a bug elsewhere?

emesare:
Can you give an example of a function that is improved, just so i can verify to make sure the refactor did not mess this up

bdash:
In MediaLibrary.framework, look for:
```
-[MLMediaLibrary init]
-[MLMediaLibrary outboundRequestQueue]
-[MLMediaLibrary outboundRequestCondition]
-[MLPMRContext cppInstrument:]
-[MLPMRContext applicationName]
-[MLPMRContext logPath]
```

They each have a tail call to `objc_msgSend` that is displayed as `return _objc_msgSend() __tailcall` with no arguments without this change.

emesare:
Before:
image

After:
image

LGTM

emesare:
Added with https://github.com/Vector35/binaryninja-api/commit/a6c5c22f08032374481aa060b59ef3492d124200

On dev with `>7183`

PR #6437, opened 2025-02-20
[SharedCache] Make it faster to look up which image contains an address

bdash:
`MemoryRegion` now tracks the start address of the image to which it belongs. `SharedCache::HeaderForAddress` uses the existing address range map to quickly find the `MemoryRegion` for a given address, and from there can cheaply look up the image via its start address.

Some extra logic is added to `CacheInfo::Load` to populate the image start address on `MemoryRegion` when loading from metadata that predates this change.

This eliminates `HeaderForAddress` as a bottleneck within `FindSymbolAtAddrAndApplyToAddr`.


emesare:
Tracking of the image in a given shared cache region exists on the https://github.com/Vector35/binaryninja-api/tree/test_shared_cache_stuff_prime_0 branch.

https://github.com/Vector35/binaryninja-api/blob/2e5182e2d28f12d7bc19cea82b27b24c6c62dab8/view/sharedcache/core/SharedCache.h#L49-L50

Issue #6411, opened 2025-02-11
Allow for sections to have duplicate names


WeiN76LQh:
**What is the feature you'd like to have?**
Sections with the same name should be allowed.

**Is your feature request related to a problem?**
Currently if multiple sections are added to a binary view with the same name only the first instance is displayed in the UI. This means that only 1 will appear in the sections list at the top of the linear view. This applies to the sections table under the `Memory Map` panel. Also there should be text where a section starts and ends in the linear view, as well as a sticky header showing the current section name as you scroll through it. This is not the case for sections with duplicate names (except for the first instance).

Loading `libsystem_c.dylib` from the DYLD Shared Cache in IDA provides the following sections:
```
Name	Start	End	R	W	X	D	L	Align	Base	Type	Class	AD	T	DS
libsystem_c:HEADER	000000018AB60000	000000018AB61A5C	R	.	X	.	L	byte	07	public	DATA	64	00	07
libsystem_c:__text	000000018AB61A5C	000000018ABD8300	R	.	X	.	L	dword	01	public	CODE	64	00	01
libsystem_c:__auth_stubs	000000018ABD8300	000000018ABD9300	R	.	X	.	L	dword	02	public	CODE	64	00	02
libsystem_c:__const	000000018ABD9300	000000018ABDB968	R	.	X	.	L	para	03	public	DATA	64	00	03
libsystem_c:__cstring	000000018ABDB968	000000018ABDEBD6	R	.	X	.	L	byte	04	public	DATA	64	00	04
libsystem_c:__oslogstring	000000018ABDEBD6	000000018ABDEC34	R	.	X	.	L	byte	05	public	DATA	64	00	05
libsystem_c:__unwind_info	000000018ABDEC34	000000018ABDFFF4	R	.	X	.	L	dword	06	public		64	00	06
libsystem_c:__got	00000001E1209578	00000001E12095B8	R	.	.	.	L	qword	08	public	DATA	64	00	08
libsystem_c:__const	00000001E12095B8	00000001E120AEC8	R	.	.	.	L	qword	09	public	DATA	64	00	09
libsystem_c:__data	00000001E4DBA558	00000001E4DBAA28	R	W	.	.	L	qword	0F	public	DATA	64	00	0F
libsystem_c:__crash_info	00000001E4DBAA28	00000001E4DBAA68	R	W	.	.	L	qword	10	public		64	00	10
[...]

PR #6454, opened 2025-02-26
[SharedCache] Avoid data memory region name conflicts & reduce section naming ones

WeiN76LQh:
The problems this commit addresses is that Binary Ninja does not allow for sections with duplicate names; the sections are added but they are unnamed. It also does not allow memory map data regions to be added with the same name and will throw an error when attempting to do so.

Prior to this commit the DSC plugin would use the base name of the image path plus the segment name, as a name for a segment region when it is added to the binary view memory map when loading an image. The problem with this is that some images in the DSC share base names, i.e. `TSUtility`. To avoid this conflict the full image path is used followed by the segment name for memory map data region names.

Additionally this same problem occurs with sections. Prior to this commit the DSC plugin used the image base name plus the section name, however even within an image there can be multiple sections with the same name. This commit changes it so that the segment name is also included in the section name. Analysis shows that images do not contain duplicate named segments and segments do not contain duplicate named sections.

The only problem remaining is the issue of duplicate image base names used in section names. So section names can still be a problem if images with the same base name are loaded. I filed issue #6411 to request sections can have duplicate names. In part this commit actually mitigates half the issue that made me file the feature request but it doesn't seem unreasonable to add support for duplicate section names and it will help fully fix the remaining issue. Arguably there another solution could be provided here for duplicate image base names, like including the image index. Although that information is not particularly useful to the user, but it does differentiate the sections that belong to a given image. I don't think it makes sense to include the full image path in section names because its overly verbose although it does inform the user the most, its just too much text.

### DIFF
--- a/view/sharedcache/core/ObjC.cpp
+++ b/view/sharedcache/core/ObjC.cpp
@@ -1124,13 +1124,13 @@ void DSCObjCProcessor::ApplyMethodTypes(Class& cls)
 void DSCObjCProcessor::PostProcessObjCSections(VMReader* reader, std::string baseName)
 {
 	auto ptrSize = m_data->GetAddressSize();
-	if (auto imageInfo = m_data->GetSectionByName(baseName + "::__objc_imageinfo"))
+	if (auto imageInfo = m_data->GetSectionByName(baseName + ".__objc_imageinfo"))
 	{
 		auto start = imageInfo->GetStart();
 		auto type = Type::NamedType(m_data, m_typeNames.imageInfo);
 		m_data->DefineDataVariable(start, type);
 	}
-	if (auto selrefs = m_data->GetSectionByName(baseName + "::__objc_selrefs"))
+	if (auto selrefs = m_data->GetSectionByName(baseName + ".__objc_selrefs"))
 	{
 		auto start = selrefs->GetStart();
 		auto end = selrefs->GetEnd();
@@ -1153,7 +1153,7 @@ void DSCObjCProcessor::PostProcessObjCSections(VMReader* reader, std::string bas
 			DefineObjCSymbol(DataSymbol, type, "selRef_" + sel, i, true);
 		}
 	}
-	if (auto superRefs = m_data->GetSectionByName(baseName + "::__objc_classrefs"))
+	if (auto superRefs = m_data->GetSectionByName(baseName + ".__objc_classrefs"))
 	{
 		auto start = superRefs->GetStart();
 		auto end = superRefs->GetEnd();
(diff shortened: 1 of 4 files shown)


WeiN76LQh:
image

Section names look like this with this PR applied

WeiN76LQh:
Need to make some more changes to support the new naming scheme. Things that rely on sections names being formatted in a certain way won't work.

WeiN76LQh:
I think this is ready to go now

emesare:
I think latest dev will have less conflicts now that the imageless regions include their start address, however images that share a name with another will still conflict.

I think the best way to fix that one is to just try adding the image section and if it fails because of name then we just prepend the start address

WeiN76LQh:
The main issue this PR addresses is the conflicts with section names. Checking the latest dev 5.0.7159-dev (70a27b05) it seems that section name conflicts is still a problem. Within an image itself it has multiple sections that have the same name but are part of different segments. For instance an image can have multiple `__const` sections. I think my PR adequately addresses the issue with a nice readable format involving the segment name as part of the section name. Something that LLDB does.

For images with duplicate names you could do what you suggest and prepend the load address but I think thats kind of ugly. Given section naming conflicts are valid, from what I can tell, implementing #6411 would be the preferred solution to conflicting image names.

emesare:
image

I have gone ahead and replicated your change for image section names to include the segments with commit: https://github.com/Vector35/binaryninja-api/commit/f902f22843146fb04dddadc5765f6582e3096a18

With that being done I think we just have the issue of duplicate image names? We really need to support duplicate section names :/

bdash:
https://github.com/Vector35/binaryninja-api/commit/f902f22843146fb04dddadc5765f6582e3096a18 broke Objective-C metadata parsing since it relies on sections being named in a specific way: https://github.com/Vector35/binaryninja-api/blob/73e7d21d029a1d3d1160cf6a7bc4a4faf41fe221/objectivec/objc.cpp#L1122-L1132

emesare:
I committed this and in the back of my head "surely no one relies on this behavior".

😮‍💨 Need to make an issue to remove the image name concept from the objective c processor.

But for now ill just write a quick fix to include the region.

emesare:
I just went ahead and pushed https://github.com/Vector35/binaryninja-api/commit/a45466fc3227c615c72e77eb7368d7e83220b325 I thought the image name was more prolific in the objective-c processor but it turned out to be easy enough to just make a getter for sections to override in the shared cache processor.

I think that is the last divergence in the objective-c base processor (divergence between MACH-O and shared cache views).

PR #6455, opened 2025-02-26
[SharedCache] Add the ability to install the plugin using CMake

WeiN76LQh:
Using `-t install` currently doesn't install the plugin into the Binary Ninja user plugin directory. This commit fixes that.

Simple change, useful for dev.


CouleeApps:
Convenient and simple, nice addition.

PR #6459, opened 2025-02-27
[SharedCache] Limit sections being added in `SharedCache::InitializeHeader` to the regions being loaded

WeiN76LQh:
The logic before would default to adding a section unless it was from a region in the `regionsToLoad` argument and its header had already been initialized. However if a user does `Load Section by Address` then there's only one region in `regionsToLoad`. All sections not in that region would be automatically added, even if they had already been or the user hadn't requested them to be loaded.

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -2424,13 +2424,13 @@ void SharedCache::InitializeHeader(
 
 	for (size_t i = 0; i < header.sections.size(); i++)
 	{
-		bool skip = false;
+		bool skip = true;
 		for (const auto& region : regionsToLoad)
 		{
 			if (header.sections[i].addr >= region->start && header.sections[i].addr < region->start + region->size)
 			{
-				if (MemoryRegionIsHeaderInitialized(lock, *region))
-					skip = true;
+				if (!MemoryRegionIsHeaderInitialized(lock, *region))
+					skip = false;
 				break;
 			}
 		}


WeiN76LQh:
I actually wanted to do a lot more with this PR in regards to section loading but I'm unsure what the desire is and a reasonable amount of code will need modifying.

Currently the user can `Load Section by Address` with a right-click action but this is misleading as it results in a segment being loaded which can load multiple sections, in some cases, quite a few. Also when right-clicking a address for memory which has not been loaded the user is offered to load a segment not a section.

It makes sense to be more granular as some segments do contain a number of sections but this will realistically require the metadata to be modified to track when a memory region (representing an image segment) has only be partially loaded due to not all of its sections being loaded.

I think either `Load Section by Address` should become `Load Segment by Address`, to more accurately represent what will happen, or the code should be modified to only load a section. The latter, to me, is more desireable. Additionally it would make sense to modify the right-click on an unloaded memory address to offer to load a section rather than a segment.

This is personal opinion so it would be good to hear what the devs think o this before I push a PR.

0cyn:
I went ahead and merged this as the fix is solid.

> I think either Load Section by Address should become Load Segment by Address, to more accurately represent what will happen, or the code should be modified to only load a section. The latter, to me, is more desireable. Additionally it would make sense to modify the right-click on an unloaded memory address to offer to load a section rather than a segment.

I think it would be good to make an issue regarding your thoughts on this to prompt further discussion, I've been moving towards segment loading as it made the code much easier to deal with (and left this feature out of the KernelCache loader entirely.)

If we do target single section rather than segment, a large swathe of stuff needs to be rethought to work with sections, e.g. how we build the `SharedCacheCore::MemoryRegion` map (currently it's by segment) and how we interact with `BinaryNinjaAPI::MemoryMap` (which creates non-file-backed Segments). 

The language absolutely needs updated in certain areas. 

I'd love to get more input from people using this specific feature.

PR #6475, opened 2025-03-10
Add function caller_sites verification in python api

### LINKED ISSUE #3722: caller_sites is not reliable
It's possible for `caller_sites` to return something that is NOT a call when accessing the IL of the returned code refs.

This should be fixed by adding a small amount of logic in:

https://github.com/Vector35/binaryninja-api/blob/dev/python/function.py#L3185-L3194

that verifies the LLIL is indeed a call

SmoothHacker:
Fixes #3722

### DIFF
--- a/python/function.py
+++ b/python/function.py
@@ -3066,7 +3066,9 @@ class Function:
 		:return: List of ReferenceSource objects of the call sites to this function
 		:rtype: list(ReferenceSource)
 		"""
-		return self.view.get_code_refs(self.start)
+		for x in self.view.get_code_refs(self.start):
+			if isinstance(x.llil, lowlevelil.LowLevelILCall):
+				yield x
 
 	@property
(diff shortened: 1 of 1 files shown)


psifertex:
Thanks for the PR! It's a good idea however there's actually another approach we might need to implement but I'd have to double check the implementation. I believe the correct approach is to actually walk the IL in the first place of each caller_function and first check if it's an LLIL call and then return it. I'd have to dig up a sample that replicates it though to verify whether that is a better approach or not. Either way there's no reason not to at least add this.

Do you mind renaming `x` to `ref` though to be a bit more explicit? Super minor, sorry.

SmoothHacker:
No worries, I was tired when I made this PR

psifertex:
I'm adding this separately with the requested variable name change. Thanks for the PR.

PR #6513, opened 2025-03-23
[SharedCache] Add a section for sectionless segments

WeiN76LQh:
In Binary Ninja it can be quite confusing when a segment is added that doesn't have any sections. `libobjc.A.dylib::__OBJC_RW` is an example of this. When scrolling through that segment there is no bar at the top of the linear view indicating where the data came from. When looking at the list of segments there are no names and with the different segment and section lists it can be hard to eyeball where a segment lands within the list of sections.

There is an argument here that the better solution is to change the way Binary Ninja displays information. For instance when scrolling through the linear view through a segment with no section, the segment information should be displayed at the top. Additionally segment names could be displayed in the segment list.

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -2422,6 +2422,7 @@ void SharedCache::InitializeHeader(
 	if (settings && settings->Contains("loader.dsc.processFunctionStarts"))
 		applyFunctionStarts = settings->Get<bool>("loader.dsc.processFunctionStarts", view);
 
+	std::set<const MemoryRegion*> regionsWithNoSections(regionsToLoad.begin(), regionsToLoad.end());
 	for (size_t i = 0; i < header.sections.size(); i++)
 	{
 		bool skip = false;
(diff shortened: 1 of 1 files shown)


0cyn:
I am of the opinion that the solution here is to:
* Add support for names on segments (MemoryRegions, which dsc uses instead of segments, already adds names)
* Add an Address display mode that supports segments
* Add display of segment names in the Memory Map
* Add a proper display of the segment name in the LinearView sticky header. 

I think the golden standard would be sections in the view accurately reflecting sections in memory. There _could_ be analysis repercussions from this. I'll invite further discussion on the topic.

WeiN76LQh:
I agree with all of that, this was just a stop gap solution because I didn't see any movement in that space in the time being and this didn't feel too problematic to implement for the time being.

0cyn:
Hi! Thank you so much for the PR! After some internal discussion we've decided that we'd like to continue working towards a proper full solution for this rather than a bandaid fix, and discussion/progress can be tracked further via #7602 . 

Thank you again for the PR

PR #6517, opened 2025-03-23
[SharedCache] Add context menu option `Load Image by Address`

WeiN76LQh:
I find there are times where I have an address and I'd like to load the entire image that the address exists in. Usually I right-click `Load Section by Address` and then look at what image gets loaded and then load the full image by name. Adding the option to load a full image based on an address seems like a useful addition.


fuzyll:
This wound up not merging super cleanly after the refactor, so I reimplemented in terms of what's now there in d222db8bd49f33716a18d0f9fc47833c2f802553. Builds >= 5.0.7165-dev (which should be available in about an hour, CI willing) should have this available.

Thank you!

PR #6540, opened 2025-03-31
Shared Cache Refactor

### LINKED ISSUE #6056: Loading an image from DSC often causes a crash related to `BNForgetUndoActions`
- Binary Ninja Version: 4.1.5902 (f2165c5d) and 4.2.6228-dev (312c3c99)
- OS: macOS
- OS Version: 15.0
- CPU Architecture: M1

**Bug Description:**
The following crash frequently occurs when loading images in DSC:

```
* thread #3, name = 'Worker PRI ', stop reason = signal SIGABRT
  * frame #0: 0x000000018b512600 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x000000018b54af70 libsystem_pthread.dylib`pthread_kill + 288
    frame #2: 0x000000018b457908 libsystem_c.dylib`abort + 128
    frame #3: 0x000000018b36167c libsystem_malloc.dylib`malloc_vreport + 896
    frame #4: 0x000000018b3894a8 libsystem_malloc.dylib`malloc_zone_error + 100
    frame #5: 0x000000018b38833c libsystem_malloc.dylib`free_tiny_botch + 40
    frame #6: 0x00000001037a20cc libbinaryninjacore.1.dylib`___lldb_unnamed_symbol29109 + 84
    frame #7: 0x00000001037a22b8 libbinaryninjacore.1.dylib`___lldb_unnamed_symbol29111 + 292
    frame #8: 0x0000000102fd4428 libbinaryninjacore.1.dylib`___lldb_unnamed_symbol14231 + 140
    frame #9: 0x0000000102fbfb98 libbinaryninjacore.1.dylib`___lldb_unnamed_symbol13841 + 280
    frame #10: 0x0000000102fc307c libbinaryninjacore.1.dylib`BNForgetUndoActions + 140
    frame #11: 0x000000010be1f44c libsharedcache.dylib`___lldb_unnamed_symbol2967 + 5300
    frame #12: 0x000000010be44f0c libsharedcache.dylib`___lldb_unnamed_symbol3314 + 1220
    frame #13: 0x000000010be54c64 libsharedcache.dylib`___lldb_unnamed_symbol3845 + 36
    frame #14: 0x00000001037d322c libbinaryninjacore.1.dylib`___lldb_unnamed_symbol29525 + 2160
    frame #15: 0x00000001037d26e4 libbinaryninjacore.1.dylib`___lldb_unnamed_symbol29520 + 12
    frame #16: 0x000000018b54b2e4 libsystem_pthread.dylib`_pthread_start + 136
```

I'm struggling to identify an exact DSC and image combination that always causes this crash. But it happens frequent enough that continually loading images will eventually cause it to happen. If I figure one out I'll post a comment with the information.

In the mean time running this script will often cause the crash eventually, although it can take some time due to how long analysis for each image takes:

```python
import argparse
import binaryninja
import binaryninja.sharedcache

parser = argparse.ArgumentParser()
parser.add_argument('dsc')

args = parser.parse_args()

print("Loading DSC")
with binaryninja.load(args.dsc) as dsc_bv:
    dsc = binaryninja.sharedcache.SharedCache(dsc_bv)
    print("Waiting for symbols to load")
    # Unsure how to wait to ensure `dsc.images` is populated below. Calling `update_analysis_and_wait` doesn't seem to help either
    dsc.load_all_symbols_and_wait()
    if len(dsc.images) == 0:
        # Sometimes it doesn't work
        print("Failed to load DSC images, re-run the script")
    print("Waiting for initial analysis to complete")
    dsc_bv.update_analysis_and_wait()
    for image in dsc.images:
        print(f"Loading image '{image.name}'")
        dsc.load_section_at_address(image.headerAddress)
        print(f"Waiting for auto analysis to complete for image '{image.name}'")
        dsc_bv.update_analysis_and_wait()
```

All it does is go through each image in the DSC and load it, wait for the auto analysis to complete and then load the next image. This may take a while to cause a crash due to analysis times for loading images in the DSC.

Tbh this script in general is pretty good at finding current crashes in the DSC analysis just because there seem to be a couple that occur regularly.

**Steps To Reproduce:**
1. Load DSC in Binary Ninja
2. Wait for initial analysis to complete
3. Load an image
4. Wait for analysis to complete
5. Repeat steps 3 and 4 until Binary Ninja crashes

**Binary:**
Extract the DSC from any iOS 18+ IPSW. Not sure if its an iOS 18 related issues, its just the ones I've been testing on.

### LINKED ISSUE #6060: DSC analysis appears to be limited to a single thread
- Binary Ninja Version: 4.2.6304-dev (65f49c87)
- OS: macOS
- OS Version: 15.0
- CPU Architecture: M1

**Bug Description:**
There seems to be performance issues with Binary Ninja when it comes to analyzing the DSC. The initial loading of `/usr/lib/system/libsystem_c.dylib` isn't too terrible, but further loading of images seems to get progressively slower. Depending on the image being loaded analysis phase 1 can have 40,000+ items to get through and thats with loading a 2nd image (the first one being `libsystem_c.dylib`). Further loading of images seems to increase the amount by 10s of thousands more each time, again depending on the image. I'm not sure if this just keeps going up and up, because there's a number of crashes making it harder to load more than 3 or 4 images. This may not actually be a problem and just be the reality of analyzing DSC but it certainly exacerbates the issue.

Something I've noticed is that analysis seems to only ever be using a single CPU core. Binary Ninja CPU utilisation during this time ranges from 100%-200%, where 100% is the equivalent to an entire CPU core being used. Often it is closer to 100%. So its likely analysis is single threaded which I guess is probably where a lot of performance is lost. With the increasing number of items to do in phase 1 each time a new image is loaded, the time to load an image quickly becomes problematic. It maybe infeasible to load something like 100 images from DSC, which should be more than possible.

When I do analysis of a standalone library extracted from DSC using [ipsw](https://github.com/blacktop/ipsw) it completes extremely quickly with analysis mode on `control flow` and still very fast when its on `full`. For instance a standalone copy of `/usr/lib/system/libsystem_c.dylib`takes 3 seconds to do `control flow` analysis on with an Apple M2 Ultra, and 15 seconds with analysis mode on `full`. By comparison the initial loading analysis of `/usr/lib/system/libsystem_c.dylib` when analyzing DSC in Binary Ninja, takes 2 minutes with analysis mode on `control flow` and 4 on `full`. Which makes sense if you're doing single core vs multi core analysis (for context the Apple M2 Ultra has 16 performance cores and 8 efficiency cores). That may not seem too bad but many of the core libraries take 10s of minutes to analyze on `control flow` mode when loaded as an image from DSC.

**Expected Behavior:**
I would expect that loading an image in DSC should be similar to analyzing that same library as an individual binary once it has been extracted from DSC using [ipsw](https://github.com/blacktop/ipsw). I recognise that performance won't be quite the same due to the tight interlinking of data in DSC, but currently it feels far from where it should be at.

**Binary:**
Extract the DSC from any iOS 18+ IPSW. Not sure if its an iOS 18 related issues, its just the ones I've been testing on.

### LINKED ISSUE #6097: hang on shutdown (or crash) when initial DSC analysis is underway
Tested on 4.2.6376.

If a slice is being loaded during shutdown, a hang or crash will occur. Example sample:

[dsc-shutdown-sample.txt](https://github.com/user-attachments/files/17655416/dsc-shutdown-sample.txt)

### LINKED ISSUE #6300: [SharedCache] Crash in `SymbolTableView` constructor background thread
- Binary Ninja Version: 4.3.6635-dev (6a04b0e5)
- OS: macOS
- OS Version: 15.1.1
- CPU Architecture: M1

**Bug Description:**
Closing a DSC triage view can cause a crash in the [`SymbolTableView` constructor](https://github.com/Vector35/binaryninja-api/blob/3c05103cfaca688f3961b78043e01b1353a238c7/view/sharedcache/ui/dsctriage.cpp#L501) at the linked line. It appears to be a UAF due to the `SymbolTableView` being deleted and then the lambda being executed. `BackgroundThread` seems to state that it guarantees this can't happen by passing the `SymbolTableView` as the owner to the `BackgroundThread`. However, in this case, it seems to not be true. No idea why.

**Steps To Reproduce:**
1. Open a copy of DSC in Binary Ninja.
2. Open the `DSCTraige`/`Dyld Shared Cache Triage` view.
3. Split that view so it opens another `DSCTraige`/`Dyld Shared Cache Triage` view next to it and then immediately close the new one.
5. Wait for a bit, the crash should occur within 10 seconds, how long it takes can vary a bit.

### LINKED ISSUE #6325: [SharedCache] Reconsider use of `BeginUndoActions` / `ForgetUndoActions` during shared cache loading / analysis
- Binary Ninja Version: 4.3-dev (current HEAD)
- OS: macOS
- OS Version: 15.1.1
- CPU Architecture: arm64

**Bug Description:**
The dyld shared cache plug-in contains a couple of locations where it does work that it brackets with `BeginUndoActions` / `ForgetUndoActions`, such as: https://github.com/Vector35/binaryninja-api/blob/09ab8876570d0da38e9a8c9d9e0c3c9076846ac1/view/sharedcache/core/SharedCache.cpp#L2954-L2969 and: https://github.com/Vector35/binaryninja-api/blob/09ab8876570d0da38e9a8c9d9e0c3c9076846ac1/view/sharedcache/core/ObjC.cpp#L1433-L1437

From discussion on Slack, undo actions are tracked at the `BinaryView` level. Forgetting undo actions will forget actions performed outside of the local code, such as simultaneous operations performed by the user or by code running on other threads.

Additionally, `BeginUndoActions` posts work to the main thread and waits for it to complete. This can pose a performance issue if the main thread is not idle.

It would be worth revisiting what the use of `BeginUndoActions` / `ForgetUndoActions` is trying to accomplish and seeing if there's an alternative solution that doesn't have these drawbacks (perhaps auto sections / symbols rather than user?).

Slack discussions that touch on this:

https://binaryninja.slack.com/archives/C1E616ZSQ/p1734975321394819
https://binaryninja.slack.com/archives/C1E616ZSQ/p1735840209591709

emesare:
Opening this PR for further discussions.

The refactor is mostly complete, just need to close out a few minor issues. This will be in the 5.0 release.

This branch will be merged as soon as possible as testing needs to be done _asap_ for inclusion in the upcoming release.

Aside from general code cleanup there were a number of specific changes that I want to highlight:

- `SharedCache` is no longer a super object that retains the state of the shared cache information and the view information.
- `SharedCache` information is populated in the view initialization, this is where it always belonged, this alleviates that issue of initialization that has caused a large number of hacks.
- Usage of core allocated Binary Ninja objects has been removed from any of the shared cache information. This removes a large swath of unneeded memory allocations and indirection (through `Ref`)
- Metadata has been greatly reduced due to the information being repopulated on load through the view init. This will help with database load/save times.
- Untangled the mess regarding `Virtual Memory` and File Accessors, we restrict usage of the file accessor to the entry file setup, and after only reference through the virtual memory.
- Virtual memory region persisted writes. Allowing us to simplify accesses to virtual memory, where writes will be persisted across file lifetime boundaries.

TODO: Add more highlights.

Left to do:
- (Done) Remove some debug logs
- (Done) Re-enable part of the shared cache workflow
- (Done) Re-enable project support
- Parallel slide info & image processing?

Testing to be done:
- More testing with older shared cache versions (ones previously supported)
- More testing with objective-c images
- Performance regressions

On performance:
The entire premise for this refactor is to open up _more_ opportunities to improve performance, this is expected to be done through parallel processing. We do expect there to be edge cases in the coming days after this is merged where there might be regressions, however these all should be easy to fix.

On backwards compatibility:
The current shared cache implementation will not be supported going forwards, however where possible we do still want to make sure BNDB's created with older versions still are usable. This is much easier to accomplish with the approach to shared cache state we took with this implementation.

On memory usage/memory allocations:
Allocations were arguably the worst bottleneck of the prior implementation, we tried our best to avoid duplicating data and reuse existing information. Moreover we removed usages of core owned/allocated objects (such as symbols) to allow for large lists (what most of our data is) to be created without causing excess memory fragmentation.


emesare:
For a high level overview, I recommend looking at the C++ API:

https://github.com/Vector35/binaryninja-api/blob/bef618a35c6ebd304270c3dabf2e0628c6564479/view/sharedcache/api/sharedcacheapi.h#L252-L327

And also the SharedCacheController header:

https://github.com/Vector35/binaryninja-api/blob/test_shared_cache_stuff_prime_0/view/sharedcache/core/SharedCacheController.h

emesare:
Example python scripts:

```py
from binaryninja import sharedcache
controller = sharedcache.SharedCacheController(bv)

# Load all direct dependencies of the loaded images.
for img in controller.loaded_images:
	for dep in controller.get_image_dependencies(img):
		dep_img = controller.get_image_with_name(dep)
		controller.apply_image(current_view, dep_img)
```

```py
from binaryninja import sharedcache
controller = sharedcache.SharedCacheController(bv)

img1 = controller.get_image_with_name("/usr/lib/system/libsystem_c.dylib")
img2 = controller.get_image_at(img1.header_address)
img3 = controller.get_image_containing(img2.header_address + 0x100)
region_start = img3.region_starts[0]
region1 = controller.get_region_at(region_start)
assert img1.header_address == region1.image_start
```

Also a verification script located in `/tests` to check and make sure all regions are correctly associated with its image: https://github.com/Vector35/binaryninja-api/blob/test_shared_cache_stuff_prime_0/view/sharedcache/tests/validate_images.py

bdash:
Is there a build that has the new core symbols that this can be built against (`BNDefineAutoUserSymbol` / `BNUndefineAutoUserSymbol`)?

emesare:
> Is there a build that has the new core symbols that this can be built against (`BNDefineAutoUserSymbol` / `BNUndefineAutoUserSymbol`)?

Oops! Forgot those were still in there. I am running down another issue rn but in the meantime you can just remove the references to those, this iteration no longer uses them so it should be easy to delete. I will push more changes later tonight.

emesare:
@bdash this version dropped the ABI changes, I will note I am trying to fix the stub call resolution right now, specifically the part where the unresolved value gets resolved to the pointer value within the referenced data variable. Don't know exactly why me loading the region does not get picked up by analysis and the constant propagated thereafter. So theres that :/

Screenshot 2025-03-31 at 1 59 36 AM

Anyways thanks for checking the branch out so quickly, I tried to add you as a reviewer but I for some reason can't.

objectivec/objc.cpp
```diff
@@ -0,0 +1,1574 @@
...
+}
+
+void ObjCProcessor::ReadListOfMethodLists(ObjCReader* reader, ClassBase& cls, std::string_view name, view_ptr_t start)
+{
+	reader->Seek(start);
+	method_list_t head;
+	head.entsizeAndFlags = reader->Read32();
+	head.count = reader->Read32();
+
+	// TODO(WeiN76LQh): probably can be removed at this point
```
0cyn: This sanity check probably shouldn't ever be removed, a bad value being in head.count here could irreversibly hang the program, we want to _really_ avoid that failure mode

objectivec/objc.cpp
```diff
@@ -0,0 +1,1574 @@
...
+			m_logger->LogDebug("ReadMethodList: Unknown method list type at 0x%llx: %d", start, start & 0x3);
+			return;
+	}
+
+	reader->Seek(start);
+	method_list_t head;
+	head.entsizeAndFlags = reader->Read32();
+	head.count = reader->Read32();
+
+	// TODO(WeiN76LQh): probably can be removed at this point
```
0cyn: ditto

0cyn: if we want do anything to these, the log text could probably be better.

objectivec/objc.h
```diff
@@ -0,0 +1,337 @@
...
+			QualifiedName cls;
+			QualifiedName category;
+			QualifiedName protocol;
+			QualifiedName protocolList;
+			QualifiedName ivar;
+			QualifiedName ivarList;
+		} m_typeNames;
+
+		bool m_isBackedByDatabase;
+		// TODO(WeiN76LQh): this is to avoid a bug with defining a classes protocol list in the DSC plugin. Remove once fixed
```
0cyn, line 281: Do we have an issue for this

objectivec/objc.cpp
```diff
@@ -0,0 +1,1574 @@
+#include "objc.h"
```
0cyn, line 1: Non-Actionable: I still really wish we could've done something with virtuals, etc to avoid having to cram so much sharedcache compensation into the Objective-C code that Mach-O view uses. It is probably fine.

view/sharedcache/api/python/sharedcache.py
```diff
@@ -1,389 +1,293 @@
...
-	def __str__(self):
-		return repr(self)
+class CacheRegion:
+    region_type: SharedCacheRegionType
+    name: str
+    start: int
+    size: int
+    image_start: int
+    # TODO: Might want to make this use the BN segment flag enum?
+    flags: sccore.SegmentFlagEnum
```
0cyn, line 20: TODO

view/sharedcache/api/python/sharedcache.py
```diff
@@ -1,389 +1,293 @@
...
-		- The target address has already been given a name by this function
-
-		:param symbol_address: Symbol address to check
-		:param target_address: Target address to apply symbol name and type to
-		:param trigger_reanalysis: Whether to reanalyze the function at target_address if the function already existed.
-		:return: None
-		"""
-		sccore.BNDSCFindSymbolAtAddressAndApplyToAddress(self.handle, symbol_address, target_address, trigger_reanalysis)
+class CacheSymbol:
+    symbol_type: sccore.SymbolTypeEnum
```
0cyn, line 42: Can we use core's `enums.SymbolType` here? Is there a reason we need our own?

view/sharedcache/api/python/sharedcache_enums.py
```diff
@@ -1,20 +1,38 @@
...
-	Loaded = 1
-	LoadedWithImages = 2
+class SharedCacheRegionType(enum.IntEnum):
+	SharedCacheRegionTypeImage = 0
+	SharedCacheRegionTypeStubIsland = 1
+	SharedCacheRegionTypeDyldData = 2
+	SharedCacheRegionTypeNonImage = 3
+
+
+class SymbolType(enum.IntEnum):
```
0cyn, line 29: binaryninja-api/python/enums.py:1300

view/sharedcache/core/MachOProcessor.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+	if (header.linkeditPresent && m_vm->IsAddressMapped(header.linkeditSegment.vmaddr))
+	{
+		if (m_applyFunctions && header.functionStartsPresent)
+		{
+			auto targetPlatform = m_view->GetDefaultPlatform();
+			auto functions = header.ReadFunctionTable(*m_vm);
+			for (const auto& func : functions)
+			{
+				// TODO: Check to make sure the func exists in a loaded region?
```
0cyn: `if (m_data->IsValidOffset(func->GetStart())`

view/sharedcache/core/MachOProcessor.cpp
```diff
@@ -0,0 +1,321 @@
...
+				// TODO: ^ this check existed prior so we should add it back.
+				m_view->AddFunctionForAnalysis(targetPlatform, func);
+			}
+		}
+
+		auto typeLib = m_view->GetTypeLibrary(header.installName);
+		m_view->BeginBulkModifySymbols();
+
+		// TODO: Why does this need to only happen in linkeditSegment?
```
0cyn, line 56: It doesn't, but if a binary doesn't have a linkedit segment mapped, symbol and export processing will fail, so we check early.

view/sharedcache/core/MachOProcessor.cpp
```diff
@@ -0,0 +1,321 @@
...
+			const auto symbols = header.ReadSymbolTable(*m_view, *m_vm);
+			for (const auto& sym : symbols)
+				ApplySymbol(m_view, typeLib, sym.ToBNSymbol());
+		}
+
+		// Apply symbols from export trie.
+		if (header.exportTriePresent)
+		{
+			// NOTE: This table is read relative to the link edit segment file base.
+			// TODO: Remove this and use the m_symbols in the cache?
```
0cyn, line 72: TODO: Can we do this? This is good for perf, I'm fairly confident we were doing it before.

view/sharedcache/core/MachOProcessor.cpp
```diff
@@ -0,0 +1,321 @@
...
+	{
+		if (initSection(header.sections[i], header.sectionNames[i]))
+			addedSections++;
+	}
+	return addedSections;
+}
+
+void SharedCacheMachOProcessor::ApplyHeaderDataVariables(SharedCacheMachOHeader& header)
+{
+	// TODO: By using a binary reader we assume the sections have all been mapped.
```
0cyn, line 214: - this shouldn't be called if the header isn't mapped
- if it is anyways, it should hard fail, as it's at best a ton of useless symbols added to Symbol List 
- thus, BinaryReader is fine here

objectivec/objc.cpp
```diff
@@ -0,0 +1,1574 @@
...
+				classRO.reserved = reader->Read32();
+			classRO.ivarLayout = ReadPointerAccountingForRelocations(reader);
+			classRO.name = ReadPointerAccountingForRelocations(reader);
+			classRO.baseMethods = ReadPointerAccountingForRelocations(reader);
+			classRO.baseProtocols = ReadPointerAccountingForRelocations(reader);
+			classRO.ivars = ReadPointerAccountingForRelocations(reader);
+			classRO.weakIvarLayout = ReadPointerAccountingForRelocations(reader);
+			classRO.baseProperties = ReadPointerAccountingForRelocations(reader);
+		}
+		catch (...)
```
0cyn, line 397: We modified every `catch (ReadException& ex)` to `catch (...)`, do we want to be doing that? Was there a reason for that?

view/sharedcache/api/sharedcache.cpp
```diff
@@ -4,231 +4,324 @@
...
+using namespace BinaryNinja;
+using namespace SharedCacheAPI;
+
+BNSharedCacheImage ImageToApi(CacheImage image)
+{
+	BNSharedCacheImage apiImage;
+	apiImage.name = BNAllocString(image.name.c_str());
+	apiImage.headerAddress = image.headerAddress;
+	apiImage.regionStartCount = image.regionStarts.size();
+	// TODO: If we alloc then core cannot delete
```
0cyn: FIXME tag at least

view/sharedcache/api/sharedcache.cpp
```diff
@@ -4,231 +4,324 @@
...
+}
+
+BNSharedCacheEntry EntryToApi(const CacheEntry& entry)
+{
+	BNSharedCacheEntry apiEntry;
+	apiEntry.path = BNAllocString(entry.path.c_str());
+	apiEntry.entryType = entry.entryType;
+	const auto& mappings = entry.mappings;
+	apiEntry.mappingCount = mappings.size();
+	// TODO: If we alloc then the core cannot delete.
```
0cyn, line 84: FIXME

view/sharedcache/api/sharedcache.cpp
0cyn: How badly do we even need this field? Could it be replaced by a separate API call that takes headerAddress as input?

view/sharedcache/api/sharedcache.cpp
```diff
@@ -4,231 +4,324 @@
...
+CacheMappingInfo MappingFromApi(BNSharedCacheMappingInfo apiMapping)
+{
+	CacheMappingInfo mapping;
+	mapping.vmAddress = apiMapping.vmAddress;
+	mapping.size = apiMapping.size;
+	mapping.fileOffset = apiMapping.fileOffset;
+	return mapping;
+}
+
+BNSharedCacheEntry EntryToApi(const CacheEntry& entry)
```
0cyn: `BNSharedCacheEntry` ->

- `BNBackingCacheEntry` 
- `BNSubCacheEntry` 

or something that indicates this is the on-disk subcaches split out across files would be ideal

view/sharedcache/api/sharedcache.cpp
0cyn: How much are we leaking here?

view/sharedcache/api/sharedcache.cpp
0cyn: I'm unsure where this is used but it'd leak ~ `7*2500*sizeof(uint64_t)` (ballpark 140kb?) every time this was called on the list of images if we can't free the list of region starts

view/sharedcache/core/FileAccessorCache.cpp
```diff
@@ -0,0 +1,111 @@
+#include "FileAccessorCache.h"
```
0cyn, line 1: for these files, a general "has this been tested for Project support?"

docs/dev/workflows.md
```diff
@@ -145,7 +145,7 @@ You can query the list of all registered workflows or filter them by type using
 # List all Module and Function workflows
```
0cyn, line 145: The tabs/spaces usage across this PR is very inconsistent 

`find -E . -regex '<DIRECTORY>\.(cpp|hpp|cc|cxx)' -exec clang-format -style=file -i {} \;` Run clang format on everything touched in this PR please. There's a .clang_format file in the API repo it should hit. Now is the best time to do this.

view/sharedcache/core/MachO.cpp
```diff
@@ -0,0 +1,660 @@
...
+// TODO: This is like 90% of the runtime.
+bool SharedCacheMachOHeader::ProcessLinkEditTrie(std::vector<CacheSymbol>& symbols, const std::string& currentText, const uint8_t* begin, const uint8_t* current, const uint8_t* end) const
+{
+	if (current >= end)
+		return false;
+
+	uint64_t terminalSize = readValidULEB128(current, end);
+	const uint8_t* child = current + terminalSize;
+
+    // The terminal is an export symbol.
```
0cyn: .clang_format

view/sharedcache/core/MachO.cpp
```diff
@@ -0,0 +1,660 @@
...
+	if (terminalSize != 0)
+	{
+		// Add the export symbol is applicable.
+		auto symbol = AddExportTerminalSymbol(currentText, current, end);
+		if (symbol.has_value())
+			symbols.push_back(*symbol);
+	}
+
+    // TODO: Make this look better
```
0cyn: .clang_format

view/sharedcache/core/MachO.cpp
```diff
@@ -0,0 +1,660 @@
...
+			return false;
+		if (!ProcessLinkEditTrie(symbols, childText, begin, begin + next, end))
+			return false;
+		childText.resize(currentText.size());
+	}
+
+	return true;
+}
+
+std::vector<CacheSymbol> SharedCacheMachOHeader::ReadExportSymbolTable(VirtualMemory& vm) const
```
0cyn: I don't really understand why this was named from ProcessExportTrie and ReadExportNode to ReadExportSymbolTable (it is a trie) and ProcessLinkEditTrie (nothing else in linkedit that I'm directly aware of uses Tries). Is also a divergence from the duplicated Mach-O view code. Maybe there is a reason.

view/sharedcache/core/MappedFile.h
```diff
@@ -0,0 +1,80 @@
...
+        other.hFile = nullptr;
+#else
+        fd = other.fd;
+        // Don't close the fd in the move.
+        other.fd = nullptr;
+#endif
+        other._mmap = nullptr;
+    }
+
+    // I hate C++
```
0cyn: too bad. do we need this comment. at least attribute it :sob:

view/sharedcache/core/SharedCacheController.cpp
```diff
@@ -0,0 +1,245 @@
...
+        return false;
+
+    if (image.header)
+    {
+        // Header information is applied to the view here, such as sections.
+        auto machoProcessor = SharedCacheMachOProcessor(&view, m_cache.GetVirtualMemory());
+        machoProcessor.ApplyHeader(*image.header);
+
+        // TODO: Make this optional.
+        // Load objective-c information.
```
0cyn: There is definitely a regression here at removing the option to not load Objective-C info for an image. I don't know what our options are at this point.

view/sharedcache/core/SharedCache.cpp
```diff
@@ -1,4007 +1,547 @@
...
-
-	return symbolsByImageName;
-}
-
-
-std::string SharedCache::SerializedImageHeaderForAddress(uint64_t address)
+// The next id to use when calling Cache::AddEntry
+static CacheEntryId nextId = 1;
+
+Ref<Symbol> CacheSymbol::ToBNSymbol() const
```
0cyn: ToSymbolRef? BNSymbol is a different type

view/sharedcache/core/SharedCache.cpp
```diff
@@ -1,4007 +1,547 @@
...
-		Serialize(context, symbol);
-	}
-	context.writer.EndArray();
+    // Add remaining region.
+    if (region.size > 0)
+        AddNonOverlappingRegion(std::move(region));
 }
 
-void Serialize(SerializationContext& context, const std::shared_ptr<std::vector<Ref<Symbol>>>& value)
+bool SharedCache::AddNonOverlappingRegion(CacheRegion region)
```
0cyn, line 193: Does this really need split out?

view/sharedcache/core/SharedCacheController.cpp
```diff
@@ -0,0 +1,245 @@
...
+        m_logger->LogError("Failed to read region: %s", e.what());
+        return false;
+    }
+
+    // Unique memory region name so that we don't cause collisions.
+    // TODO: Better name? I dont really think so...
+    const auto memoryRegionName = fmt::format("{}_0x{:x}", region.name, region.start);
+
+    // NOTE: Adding a data memory region will store the entire contents of the region in the BNDB.
+    // TODO: We can use the AddRemoteMemoryRegion if we want to reload on view init.
```
0cyn: Can't guarantee backing files exist

view/sharedcache/core/SharedCacheController.cpp
```diff
@@ -0,0 +1,245 @@
...
+{
+    // TODO: Check m_loadedRegions? This seems redundant...
+    // Loads the given region into the BinaryView and marks it as loaded.
+    // First check to make sure there isn't already a segment at the address in the view.
+    if (view.GetSegmentAt(region.start))
+        return false;
+
+    // Skip LINKEDIT region, it is like 300mb.
+    // TODO: Make this configurable. i.e. m_blacklistedRegions
+    if (region.name.find("LINKEDIT") != std::string::npos)
```
0cyn: Did we regress a load option here

view/sharedcache/core/SharedCacheView.cpp
```diff
@@ -661,129 +761,90 @@ bool DSCView::Init()
...
-	Ref<BinaryView> viewRef = Parse(data);
-	if (!viewRef || !viewRef->Init())
 	{
-		LogWarn("Failed to initialize view of type '%s'. Generating default load settings.", GetName().c_str());
-		viewRef = data;
+		// Add up all the cache entries using the cache processor.
+		// After this we should have all the mappings available as well.
+		CacheProcessor cacheProcessor = CacheProcessor(this);
+		auto startTime = std::chrono::high_resolution_clock::now();
+		if (!cacheProcessor.ProcessCache(sharedCache))
```
0cyn: Do we fail to deser a bndb if the files are gone? That is a huge issue if so.

view/sharedcache/core/SharedCacheView.cpp
```diff
@@ -661,129 +761,90 @@ bool DSCView::Init()
...
+		std::chrono::duration<double> elapsed = endTime - startTime;
+		logger->LogInfo("Processing %zu regions took %.3f seconds", sharedCache.GetRegions().size(), elapsed.count());
+	}
 
-bool DSCViewType::IsTypeValidForData(BinaryNinja::BinaryView* data)
-{
-	if (!data)
-		return false;
+	{
+		// TODO: Remove this before release or make it faster, this ironically is the slowest thing here.
```
0cyn: .

view/sharedcache/core/SlideInfo.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+SlideInfoProcessor::SlideInfoProcessor(uint64_t baseAddress)
+{
+	m_logger = new BinaryNinja::Logger("SlideInfoProcessor");
+	m_baseAddress = baseAddress;
+}
+
+void ApplySlideInfoV5(VirtualMemory& vm, const SlideMappingInfo& mapping)
+{
+	// TODO: This is off the entry itself.
```
0cyn: I'm unsure what this means

view/sharedcache/core/SharedCacheView.cpp
```diff
@@ -1,33 +1,154 @@
...
+
+	settings->RegisterSetting("loader.dsc.processCFStrings",
+		R"({
+		"title" : "Process CFString Metadata",
+		"type" : "boolean",
+		"default" : true,
+		"description" : "Processes CoreFoundation strings, applying string values from encoded metadata"
+		})");
+
+	settings->RegisterSetting("loader.dsc.autoLoadLibSystem",
```
0cyn: Does this do anything

view/sharedcache/core/SharedCacheView.cpp
```diff
@@ -1,33 +1,154 @@
...
+	for (const auto& override : overrides)
+	{
+		if (settings->Contains(override))
+			settings->UpdateProperty(override, "readOnly", false);
+	}
+
+	Ref<Settings> programSettings = Settings::Instance();
+	programSettings->Set("analysis.workflows.functionWorkflow", "core.function.sharedCache", viewRef);
+
+	settings->RegisterSetting("loader.dsc.processCFStrings",
```
0cyn, line 57: Does nothing

view/sharedcache/core/SharedCacheView.cpp
```diff
@@ -1,33 +1,154 @@
...
+
+	settings->RegisterSetting("loader.dsc.autoLoadLibSystem",
+		R"({
+		"title" : "Auto-Load libSystem",
+		"type" : "boolean",
+		"default" : true,
+		"description" : "Whether to automatically load libsystem_c.dylib. This image contains frequently used noreturn symbols, and not loading it will result in frequently incorrect control flows."
+		})");
+
+	settings->RegisterSetting("loader.dsc.processObjC",
```
0cyn, line 73: Does nothing

view/sharedcache/core/SharedCacheView.cpp
```diff
@@ -1,33 +1,154 @@
...
+
+	settings->RegisterSetting("loader.dsc.autoLoadStubsAndDyldData",
+		R"({
+		"title" : "Auto-Load Stub Islands",
+		"type" : "boolean",
+		"default" : true,
+		"description" : "Automatically loads stub and dylddata regions that contain just branches and pointers. These are required for resolving stub names, and performance impact is minimal. Recommended you keep this on."
+		})");
+
+	settings->RegisterSetting("loader.dsc.allowLoadingLinkeditSegments",
```
0cyn: Does nothing

view/sharedcache/core/Utility.h
```diff
@@ -0,0 +1,83 @@
...
+uint64_t readValidULEB128(const uint8_t*& current, const uint8_t* end);
+
+void ApplySymbol(BinaryNinja::Ref<BinaryNinja::BinaryView> view, BinaryNinja::Ref<BinaryNinja::TypeLibrary> typeLib, BinaryNinja::Ref<BinaryNinja::Symbol> symbol);
+
+// Returns the on disk file for a given path, this is required to support project files.
+std::string ResolveFilePath(BinaryNinja::Ref<BinaryNinja::BinaryView> dscView, const std::string& path);
+
+// Returns the "image name" for a given path.
+// /blah/foo/bar/libObjCThing.dylib -> libObjCThing.dylib
+std::string FileName(const std::string& path);
```
0cyn: basename

view/sharedcache/ui/SharedCacheUINotifications.cpp
```diff
@@ -20,124 +21,141 @@ void UINotifications::init()
...
+	auto loadImageTokenAction = [](const UIActionContext& ctx) {
+		BackgroundThread::create(ctx.context->mainWindow())
+			->thenBackground([ctx](){ loadImageAtAddr(*ctx.binaryView, ctx.token.token.value); })
+			->start();
+	};
+
+	auto isValidUnloadedRegionAction = [](const UIActionContext& ctx) {
+		uint64_t addr = ctx.token.token.value;
+		// Check if the region is already loaded in the view.
+		if (!ctx.binaryView->GetSectionsAt(addr).empty())
```
0cyn, line 99: IsValidOffset() exists

view/sharedcache/ui/symboltable.cpp
```diff
@@ -0,0 +1,129 @@
...
+	delete m_model;
+}
+
+void SymbolTableView::populateSymbols(BinaryView &view)
+{
+	if (auto controller = SharedCacheController::GetController(view)) {
+		BackgroundThread::create(this)
+			->thenBackground([this, controller]() {
+				std::vector<CacheSymbol> newSymbols = controller->GetSymbols();
+				m_symbols.swap(newSymbols);
```
0cyn: mutex? At least a fixme.

0cyn (changes requested):
I still need to test this locally, this is an initial runthrough

bdash:
> * Untangled the mess regarding Virtual Memory and File Accessors, we restrict usage of the file accessor to the entry file setup, and after only reference through the virtual memory.
> *  Virtual memory region persisted writes. Allowing us to simplify accesses to virtual memory, where writes will be persisted across file lifetime boundaries.

I'm not sure if you have profiled this at all, but these two changes add significant overhead to parsing and applying slide information. This is done on the main thread and the application is unresponsive for multiple seconds while this work is performed.

`ApplySlideInfoV5` calls `VirtualMemory::ReadUInt64` many thousands of times. This spends a full 50% of its time within `VirtualMemory::GetRegionAtAddress` repeatedly looking up the same `VirtualMemoryRegion` and locking the same `WeakFileAccessor`.  This accounts for several hundred milliseconds of work that blocks the main thread.

`ApplySlideInfoV5` also calls `VirtualMemory::WritePointer` many thousands of times. This spends 75% of its time (over 800 milliseconds) below `FileAccessorWriteLog::AddPointer` adding entries to a hash table. AFAICT this hash table is used only if the file accessor is evicted from the cache. This should be rare in typical usage given the cache size (512 entries) vs number of files in a shared cache (~70).  I'd typically look to the commit message to understand why the approach was changed from the previous approach of parsing / reapplying slide info when reopening the file. Sadly this looks to come from one of the dozens of "wip" commits which doesn't provide much clarity.

docs/dev/workflows.md
emesare: Doing this last to avoid another round off diffs

view/sharedcache/core/FileAccessorCache.cpp
emesare: Project file path resolution happens in the cache processor, yet to add back the project file resolution there, but the accessor stuff doesnt need to know about projects

view/sharedcache/core/MachO.cpp
emesare: Yea idk why i renamed it that, i am going to rename it back

view/sharedcache/core/SharedCacheController.cpp
emesare: We already discussed this internally, ill add those comments to source

view/sharedcache/core/SlideInfo.cpp
```diff
@@ -0,0 +1,304 @@
+#include "SlideInfo.h"
+
+#include "Dyld.h"
+#include "SharedCache.h"
+
+#ifdef _MSC_VER
+int count_trailing_zeros(uint64_t value) {
```
bdash: These helper functions and the other non member functions in this file should be either static or in an unnamed namespace so they're internal to this translation unit.

emesare:
> > * Untangled the mess regarding Virtual Memory and File Accessors, we restrict usage of the file accessor to the entry file setup, and after only reference through the virtual memory.
> > * Virtual memory region persisted writes. Allowing us to simplify accesses to virtual memory, where writes will be persisted across file lifetime boundaries.
> 
> I'm not sure if you have profiled this at all, but these two changes add significant overhead to parsing and applying slide information. This is done on the main thread and the application is unresponsive for multiple seconds while this work is performed.
> 
> `ApplySlideInfoV5` calls `VirtualMemory::ReadUInt64` many thousands of times. This spends a full 50% of its time within `VirtualMemory::GetRegionAtAddress` repeatedly looking up the same `VirtualMemoryRegion` and locking the same `WeakFileAccessor`. This accounts for several hundred milliseconds of work that blocks the main thread.
> 
> `ApplySlideInfoV5` also calls `VirtualMemory::WritePointer` many thousands of times. This spends 75% of its time (over 800 milliseconds) below `FileAccessorWriteLog::AddPointer` adding entries to a hash table. AFAICT this hash table is used only if the file accessor is evicted from the cache. This should be rare in typical usage given the cache size (512 entries) vs number of files in a shared cache (~70). I'd typically look to the commit message to understand why the approach was changed from the previous approach of parsing / reapplying slide info when reopening the file. Sadly this looks to come from one of the dozens of "wip" commits which doesn't provide much clarity.

The header files and comments in the source provide clarification at least from my end, we have already decided to support evicting file descriptors and re-adding them once reused. If we wanted to remove this support we could remove the write log.

For performance, the slide info stuff is currently done wholesale across all entries, however we don't need to do that unless we actually want to access them, we could do it lazily. We also can process them in parallel.

view/sharedcache/core/SharedCacheController.cpp
emesare: Not that im aware, this is a comment just that we could, a lot of my TODO's are highlighting improvements not necessarily regressions

view/sharedcache/core/SlideInfo.cpp
emesare: Sorry i forgot to remove those todos 👍

view/sharedcache/core/SharedCacheController.cpp
0cyn: To be clear, we did regress a load option here, just didn't see it for another 50 files, should've updated this comment

emesare:
For the slide info I am going to introduce some more reader/writer methods that assume the address param is already off the file accessor. This should get rid of the region lookup.

bdash:
> The header files and comments in the source provide clarification at least from my end, we have already decided to support evicting file descriptors and re-adding them once reused. If we wanted to remove this support we could remove the write log.

A write log isn't the only way to support this. The existing implementation handles this by parsing and reapplying the slide info in the rare event that a file is closed and later re-opened. I looked and did not see any comments that mention why using a write log was preferred vs that approach.

> For performance, the slide info stuff is currently done wholesale across all entries, however we don't need to do that unless we actually want to access them, we could do it lazily. We also can process them in parallel.

Laziness and/or parallelism may help some on iOS, but for arm64 macOS the shared cache is two files and a total of nine mappings. A large percentage of the data will have to be processed as soon as any image is loaded from the shared cache.

view/sharedcache/workflow/SharedCacheWorkflow.cpp
```diff
@@ -13,518 +13,373 @@
...
-					if (dest.GetSourceExpr().operation == MLIL_CONST_PTR)
+					auto targetAddr = instr.GetDestExpr<MLIL_CALL_SSA>().GetConstant();
+					if (!view->IsValidOffset(targetAddr))
 					{
-						dest = dest.GetSourceExpr();
-						if (!bv->IsValidOffset(dest.GetConstant()))
-						{
-							ProcessOffImageCall(ctx, func, mssa, dest, firstInstruction.GetSSAExprIndex());
-						}
+						LogInfo("processOffImageCall MLIL_CALL_SSA @ %llx", instr.address);
```
0cyn: spam

0cyn: I'm also getting `[Default] Loading region dyld_shared_cache_arm64e.33.dylddata::_data_4 for off-image call... type 2 @ sub func 1f4165000 symbol 1f4165000
` spam, unsure if that was removed in a later commit

0cyn (changes requested):

emesare:
Reminder to pull in: https://github.com/Vector35/binaryninja-api/commit/0212799235842096236126193623529028535727 once the workflow activity is adding the appropriate stub symbol names

view/sharedcache/api/sharedcache.cpp
emesare: Its not leaking, its an issue with assuming that the allocator is the same, i.e. allocs on api side with a free on core

objectivec/objc.h
emesare: @WeiN76LQh pinging for clarification

objectivec/objc.cpp
emesare: The shared cache reader does not emit that specific exception, we could restrict it further if desired.

emesare:
Going to start tracking issues to be closed with this release, either they:
1. Are no longer relevant
2. Have previously been fixed by one of the other commits that will be pushed alongside this one
3. Were actually fixed by the refactor

emesare:
Currently the blocking issue is with the stub function recognization (in the provided workflow). If we don't get a resolution on this today (April 1st) then this branch will be merged as is with a subsequent patch following, to allow for enough time to test the core shared cache refactor.

objectivec/objc.h
WeiN76LQh: Honestly I can't remember specifically why this was added to only skip base protocol parsing for class RO data but it might have been fixed on dev. Probably something to do with the special list types in the DSC. Not sure why this was only an issue for protocols in this scenario. This was a quick hack to avoid haaving to fix the issue so that I could get the original PR published

emesare:
@bdash I understand that the above message is not ideal, but we need to move forward. I have asked already for others on the team to review the workflow. The workflow itself is a small component of the refactor and we will have ample time to fix it for the upcoming release if it isn't fixed today. We are not shipping 5.0 with the regression.

view/sharedcache/ui/symboltable.cpp
bdash: There's a data race that leads to a use after free here.

Multiple background threads created by `populateSymbols` can be in flight at a time. The second background call overwrites `m_symbols` on the background thread while the first has moved on to reading from it on the main thread within `SymbolTableModel::updateSymbols`.

This race happens immediately when loading a shared cache since the `DSCTriageView` constructor ends up calling `populateSymbols` and then shortly after a second call is made from below `NotifyOnAfterOpenFile`.

emesare:
Pushed some fixes for the symbol table ui, this should hopefully fix the crash related to this branch and the one that existed prior: https://github.com/Vector35/binaryninja-api/commit/311380d4b7ca64d9d6258fc02ff15090f5e58eac

Let me know if that seems reasonable.

emesare:
Linux testing seems to have something in the cache processor not pickup the files in the same directory. Need to investigate that.

The new workflow from my testing has all the same properties as the original, so those regressions are seemingly gone.

I have found that the code path here: https://github.com/Vector35/binaryninja-api/blob/96c40afe7296aba26ee7fb639517c7f16e137cc6/view/sharedcache/core/Utility.cpp#L101-L128 is seemingly not getting picked up (the param type for "obj" param i mean), need to investigate that.

view/sharedcache/core/MappedFileAccessor.h
```diff
@@ -0,0 +1,103 @@
...
+
+public:
+	explicit UnmappedAccessException(uint64_t address, uint64_t fileLength) : m_address(address), m_fileLen(fileLength)
+	{}
+
+	virtual const char* what() const throw()
+	{
+		thread_local std::string message;
+		message =
+			fmt::format("Tried to access unmapped address {0:x} for file with length of {0:x}", m_address, m_fileLen);
```
plafosse: should be
```
"Tried to access unmapped address {0:x} for file with length of {01:x}"
```

view/sharedcache/core/MappedFileAccessor.cpp
```diff
@@ -0,0 +1,110 @@
...
+	if (file->Map() != MapStatus::Success)
+		return nullptr;
+	return std::make_shared<MappedFileAccessor>(std::move(*file));
+}
+
+// TODO: Will obviously not work on 32bit binaries, need to make WritePointer64 and 32 equiv.
+void MappedFileAccessor::WritePointer(size_t address, size_t pointer)
+{
+	m_dirty = true;
+	*reinterpret_cast<size_t*>(&m_file._mmap[address]) = pointer;
```
plafosse, line 20: Missing overflow check. Probably should be: 
```
void MappedFileAccessor::WritePointer(size_t address, size_t pointer)
{
	if (address + sizeof(size_t*) > Length())
		throw UnmappedAccessException(address, Length());
	m_dirty = true;
	*reinterpret_cast<size_t*>(&m_file._mmap[address]) = pointer;
}

```

view/sharedcache/core/MappedFile.cpp
```diff
@@ -0,0 +1,193 @@
...
+#else
+uint64_t AdjustFileDescriptorLimit()
+{
+	return 0x1000000;
+}
+
+MappedFile::~MappedFile()
+{
+	Unmap();
+	if (hFile)
```
plafosse, line 118: We don't consistently set hFile to null so this check everywhere we do `CloseHandle(hFile)` we should be also setting the handle to null

objectivec/objc.cpp
0cyn: I think it'd be a good TODO: to add.

objectivec/objc.cpp
```diff
@@ -0,0 +1,1572 @@
...
+	std::map<std::string, Ref<Metadata>> clsMeta;
+
+	clsMeta["loc"] = new Metadata(loc);
+	clsMeta["name"] = new Metadata(cls.name);
+	clsMeta["typeName"] = new Metadata(cls.associatedName.GetString());
+
+	std::vector<uint64_t> instanceMethods;
+	std::vector<uint64_t> classMethods;
+	instanceMethods.reserve(cls.instanceClass.methodList.size());
+	classMethods.reserve(cls.metaClass.methodList.size());
```
plafosse, line 30: Looks like no class methods are ever put into this list. What is intended here?

objectivec/objc.cpp
```diff
@@ -0,0 +1,1572 @@
...
+std::vector<QualifiedNameOrType> ObjCProcessor::ParseEncodedType(const std::string& encodedType)
+{
+	std::vector<QualifiedNameOrType> result;
+	int pointerDepth = 0;
+
+	bool readingNamedType = false;
+	std::string namedType;
+	int readingStructDepth = 0;
+	std::string structType;
+	char last;
```
plafosse, line 107: this should _probably_ be initialized

objectivec/objc.cpp
```diff
@@ -0,0 +1,1572 @@
...
+		DefineObjCSymbol(BNSymbolType::DataSymbol, m_typeNames.cls, "cls_" + name, classPtr, true);
+		DefineObjCSymbol(BNSymbolType::DataSymbol, m_typeNames.classRO, "cls_ro_" + name, classROPtr, true);
+		DefineObjCSymbol(BNSymbolType::DataSymbol, Type::ArrayType(Type::IntegerType(1, true), name.size() + 1),
+			"clsName_" + name, classRO.name, true);
+		if (classRO.baseProtocols && !m_skipClassBaseProtocols)
+		{
+			DefineObjCSymbol(BNSymbolType::DataSymbol, Type::NamedType(m_data, m_typeNames.protocolList),
+				"clsProtocols_" + name, classRO.baseProtocols, true);
+			reader->Seek(classRO.baseProtocols);
+			uint32_t count = reader->Read64();
```
plafosse, line 438: Reading a 64bit into a 32bit value... sus

objectivec/objc.cpp
```diff
@@ -0,0 +1,1572 @@
...
+			{
+				metaClsStruct.isa = ReadPointerAccountingForRelocations(reader);
+				metaClsStruct.super = reader->ReadPointer();
+				metaClsStruct.cache = reader->ReadPointer();
+				metaClsStruct.vtable = reader->ReadPointer();
+				metaClsStruct.data = ReadPointerAccountingForRelocations(reader) & ~1;
+				DefineObjCSymbol(BNSymbolType::DataSymbol, m_typeNames.cls, "metacls_" + name, clsStruct.isa, true);
+				hasValidMetaClass = true;
+			}
+			catch (...)
```
plafosse, line 461: I

objectivec/objc.cpp
emesare: https://github.com/Vector35/binaryninja-api/blame/0d228d32f6087ea95d5de5235dd25aa6eeb7f406/view/macho/objc.cpp#L37

hmm

objectivec/objc.cpp
```diff
@@ -0,0 +1,1572 @@
...
+				metaClassRO.baseMethods = ReadPointerAccountingForRelocations(reader);
+				metaClassRO.baseProtocols = ReadPointerAccountingForRelocations(reader);
+				metaClassRO.ivars = ReadPointerAccountingForRelocations(reader);
+				metaClassRO.weakIvarLayout = ReadPointerAccountingForRelocations(reader);
+				metaClassRO.baseProperties = ReadPointerAccountingForRelocations(reader);
+				DefineObjCSymbol(
+					BNSymbolType::DataSymbol, m_typeNames.classRO, "metacls_ro_" + name, metaClsStruct.data, true);
+				hasValidMetaClassRO = true;
+			}
+			catch (...)
```
plafosse, line 493: do

objectivec/objc.cpp
```diff
@@ -0,0 +1,1572 @@
...
+			}
+		}
+
+		if (classRO.baseMethods)
+		{
+			try
+			{
+				ReadMethodList(reader, cls.instanceClass, name, classRO.baseMethods);
+			}
+			catch (...)
```
plafosse, line 506: not

objectivec/objc.cpp
```diff
@@ -0,0 +1,1572 @@
...
+				m_logger->LogError("Failed to read the method list for class pointed to by 0x%llx", clsStruct.data);
+			}
+		}
+		if (hasValidMetaClassRO && metaClassRO.baseMethods)
+		{
+			try
+			{
+				ReadMethodList(reader, cls.metaClass, name, metaClassRO.baseMethods);
+			}
+			catch (...)
```
plafosse, line 517: like

objectivec/objc.cpp
```diff
@@ -0,0 +1,1572 @@
...
+			}
+		}
+
+		if (classRO.ivars)
+		{
+			try
+			{
+				ReadIvarList(reader, cls.instanceClass, name, classRO.ivars);
+			}
+			catch (...)
```
plafosse, line 529: catch alls

plafosse: We should know the type of exception we're expecting and handle them appropriately IMO

objectivec/objc.cpp
emesare: If I had to guess, class methods had a billion and was removed for performance? There is no comment so i guess @0cyn where is the next chain of custody over this code

0cyn: I am unfamiliar with most of this code at this point; My guess is that I or someone down the line just forgot to add serialization for this. This is not necessarily super impactful till we have the improved Objective-C type prop, type libraries, etc. 

Currently this serialization is not used at all internally, is not really documented, and has only really existed to give a head start to any enthusiast wanting to write a tool utilizing it. 

Best option to fix this is probably 'add an identical loop here on the contents of `cls.metaClass.methodList`', but this is not a regression and damages nothing.

PR #6553, opened 2025-04-03
[SharedCache] Fix detection of the class name of categories

bdash:
The logic for determining the class name of a category did not correctly handle classes defined in other images in the shared cache. There were two problems:
1. If the class is defined in another image that is already loaded, `ObjCProcessor` has already renamed the symbol from `_OBJC_CLASS_$_` to `cls_`. Both forms of symbol name are now handled.
2. If the class is defined in an image that is not yet loaded, no symbol name is available. The category's class is now looked up in the shared cache symbol table, and the symbol's name is parsed as if it were an import symbol.

This fixes all cases of "Failed to determine base classname for category" that I have come across.

### DIFF
--- a/objectivec/objc.cpp
+++ b/objectivec/objc.cpp
@@ -581,13 +581,26 @@ void ObjCProcessor::LoadCategories(ObjCReader* reader, Ref<Section> classPtrSect
 			categoryBaseClassName = it->second.name;
 			category.associatedName = it->second.associatedName;
 		}
-		else if (auto symbol = m_data->GetSymbolByAddress(catLocation + m_data->GetAddressSize()))
-		{
-			if (symbol->GetType() == ImportedDataSymbol || symbol->GetType() == ImportAddressSymbol)
+		else
+		{	
+			auto symbol = m_data->GetSymbolByAddress(cat.cls);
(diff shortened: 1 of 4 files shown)


emesare:
Would something like this be ok with you? https://github.com/Vector35/binaryninja-api/commit/fb1d9e15e4a1f5c0912815f7b8dd0ea5bbc9a62e

Nothing functionally changes I just think that is easier to understand from the base classes perspective, to always call `GetSymbol` rather than two separate calls which ultimately do the same thing.

If you are fine with that change I will fix it up into your commit with you being the author, or i can leave it separate, ill leave that up to you.
If you are not fine with that change I am still OK with merging without my change.

bdash:
I wasn't happy with the naming / factoring of `SymbolForUnmappedAddress` so fb1d9e15e4a1f5c0912815f7b8dd0ea5bbc9a62e is a clear improvement. Feel free to squash it if that is what works for you.

emesare:
Cherry-picked onto dev with https://github.com/Vector35/binaryninja-api/commit/254fd7b1f48c0e79889e4c786c7e507e7ac2e067

Will be on dev build `>7154`. Thank you again!

Issue #6106, opened 2024-11-08
Parsing of __init_offsets section in MachO


mostobriv:
**What is the feature you'd like to have?**
As for some time apple presented new ways to store global constructors pointers, modern compilers used to add the sections `__init_offsets` (pointers there stored as 32-bit RVA) instead of an old one `__mod_init_func` (pointers there stored as raw pointers), i think BN should be able to parse it as well and use it within its analysis.

**Is your feature request related to a problem?**
Not much, just simple enhancement

**Are any alternative solutions acceptable?**
It can be easily done with scripting yeah.

**Additional Information:**
There is dyld code, related to getting initialisers from this section:
https://github.com/apple-oss-distributions/dyld/blob/65bbeed63cec73f313b1d636e63f243964725a9d/cache-builder/OptimizerLinkedit.cpp#L370-L377

mostobriv:
Implemented in:
https://github.com/Vector35/binaryninja-api/pull/6555

PR #6555, opened 2025-04-03
[MachoView] Add support of __init_offsets section

mostobriv:
This MR is addressing #6106 

Provided code adding parsing of __init_offsets section and in addition defining 32-bit pointers (contained in this section) to init functions (pointers is treated as 32-bit relative to `__TEXT` pointers - `__ptr32 __offset(const, 0xXXXXXX)`).


0cyn:
Thank you for the PR! Merged with 4bc3153 and available in versions >= `dev 5.1.7553`

PR #6559, opened 2025-04-04
[ObjC] Parse sections containing Objective-C constants

bdash:
This adds support for the `__objc_arrayobj`, `_objc_dictobj`, `__objc_intobj`, `__objc_floatobj`, `__objc_doubleobj` and `__objc_dateobj` sections that contain Objective-C constants. These are emitted by Apple's versions of Clang for `const` literals, amongst other things.

There are a few points here where I'm not sure what the best approach is:
1. Should there be a new setting or settings to control parsing of these sections? For now I've keyed it off the setting for `CFString`, but if it stays like that the setting description will need to be updated.
2. How should the symbols for the variables in these sections be named?
3. Does it make sense to create a single `SymbolQueue` and share it between all of these functions rather than having each create its own?
4. Should this code really be doing the `BeginUndoActions` / `ForgetUndoActions` dance? I copied it from `ProcessCFStrings`. IIRC that's problematic for performance due to needing to synchronize with the main thread, and doesn't make a lot of sense for background work like this.
5. Any suggestions on how to avoid some of the duplication in this code would be appreciated. Sharing the `SymbolQueue` would simplify things slightly. 
6. This code can throw exceptions, but doesn't seem to be particularly exception safe. The `SymbolQueue` is a member variable that's a raw pointer and will leak if an exception is thrown. An exception after `BeginBulkModifySymbols` is called will leave that in an unbalanced state. These are issues are all carried over from `ProcessCFString`.

A couple of libraries that are good for testing with in the macOS shared cache:
1. ImageIO has arrays, dictionaries, floats, and doubles.
2. libXTFontStaticRegistryData.dylib has integers, dates, and data constants.

### DIFF
--- a/objectivec/objc.cpp
+++ b/objectivec/objc.cpp
@@ -1535,6 +1535,15 @@ void ObjCProcessor::ProcessObjCData()
 	m_relocationPointerRewrites.clear();
 }
 
+void ObjCProcessor::ProcessObjCLiterals()
+{
+	ProcessCFStrings();
+	ProcessNSConstantArrays();
+	ProcessNSConstantDictionaries();
+	ProcessNSConstantIntegerNumbers();
+	ProcessNSConstantFloatingPointNumbers();
+	ProcessNSConstantDatas();
+}
 
 void ObjCProcessor::ProcessCFStrings()
 {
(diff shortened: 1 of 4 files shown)


0cyn:
Merged with `d7445e23` and available in **dev versions >= 5.1.7877**

PR #6565, opened 2025-04-04
[SharedCache] Apply call type overrides to objc_msgSendSuper / objc_msgSendSuper2

bdash:
These are treated the same as `objc_msgSend` with the exception of their first argument being an `objc_super*` rather than `id`.


emesare:
image

Straightforward with good code, I am going to just go ahead and merge, seems to work well

Thanks!

PR #6576, opened 2025-04-07
[ObjC] Handle Objective-C ivars with an offset of 0

bdash:
objc-runtime-new.mm mentions an offset of 0 is used for anonymous bitfields.

There are a few libraries in the shared cache that were hitting this. Previously this would throw an exception when attempting to read from offset 0 and the types would not be applied for the ivar.


emesare:
image

image

LGTM

PR #6580, opened 2025-04-07
[SharedCache] Dont apply invalid symbols in the stub workflow

WeiN76LQh:
There was no check to ensure `symbol` actually had a value (as its a `std::optional`). This was resulting in invalid symbols being added.


emesare:
OK That should keep you as author, I just wanted to express the failure state in the return.

PR #6583, opened 2025-04-07
implement rust plugin command

### LINKED ISSUE #6362: Add PluginCommand to Rust API
This would allow us to handle interactions programmatically call plugin commands from rust, a use case can be found here:

https://github.com/Vector35/binaryninja-api/pull/6309#issuecomment-2614108163

> I know this is most likely an inappropriate place to ask, but I'm trying to understand how to run a plugin via the Rust API (BinExport) specifically. Are there any examples anywhere on how to do this? I can also ask on Slack if that's a better place?

C++ API class:

https://github.com/Vector35/binaryninja-api/blob/8767400b712f12f459556844e27d9b9f5ff037bd/binaryninjaapi.h#L14146-L14168

rbran:


rust/src/plugin_command.rs
```diff
@@ -0,0 +1,261 @@
...
+        ),
+    },
+    BNGetValidPluginCommandsForFunction, BNRegisterPluginCommandForFunction, FunctionPluginCommand, CustomFunctionPluginCommand {
+        functionCommand::functionIsValid(
+            func: *mut BNFunction: &Function = func.handle => &Function::from_raw(func),
+        ),
+    },
+    BNGetValidPluginCommandsForLowLevelILFunction, BNRegisterPluginCommandForLowLevelILFunction, LowLevelILFunctionPluginCommand, CustomLowLevelILFunctionPluginCommand {
+        lowLevelILFunctionCommand::lowLevelILFunctionIsValid(
+            // TODO I don't know what Generics should be used here
```
emesare: RegularLowLevelILFunction<CoreArchitecture>

Need to document this more... but generally use that unless you are emitting LLIL or need SSA version

rbran: `RegularLowLevelILFunction<CoreArchitecture>` with `BinaryView::default_arch`?

emesare: Looks like the LLIL function in rust needs to stop trying to track the arch at construction, i.e. do what C++ does

https://github.com/Vector35/binaryninja-api/blob/12f1e2ea803b7ebb990bef3db7c58fd10d852b1a/lowlevelil.cpp#L55-L61

IDK the implication of this, but otherwise we need to manually retrieve the underlying function object to get the architecture.

You might have ran into this with your PR https://github.com/Vector35/binaryninja-api/pull/5318

emesare: This is likely so that we get the generic A in the `arch` signature here: https://github.com/Vector35/binaryninja-api/blob/f77dbcea6074966d81a44e2ce0dd83671a056c4d/rust/src/low_level_il/function.rs#L97-L99

So we don't have another layer of indirection though the core to get at our architecture from LLIL.

emesare: Calling the default_arch obviously is not the correct move here, but you can leave it for now and add a TODO, the underlying issue is we need a way to construct the LLIL function with the function objects underlying core architecture, this is easy to "solve" with a `new` function that calls https://github.com/Vector35/binaryninja-api/blob/f77dbcea6074966d81a44e2ce0dd83671a056c4d/rust/src/low_level_il/function.rs#L153-L158 to get the function architecture

And then sets the arch handle for the LLIL function: https://github.com/Vector35/binaryninja-api/blob/f77dbcea6074966d81a44e2ce0dd83671a056c4d/rust/src/low_level_il/function.rs#L61

emesare: I think the right answer is a new `new` function that will lookup the `CoreArchitecture`. I think keeping that generic around is the right move.

EDIT: Didn't see the commit right below, ignore this message lol

PR #6621, opened 2025-04-10
implement rust language representation

### LINKED ISSUE #6484: Add language representation API to Rust
Currently the language representation API (see C++: https://github.com/Vector35/binaryninja-api/blob/dev/languagerepresentation.cpp) is only exposed in python and C++ this was done as there was a large scale refactor occurring for rust at the time. The language representation API should now be implemented in the Rust API as all of the dependent API's (text token related stuff) has been implemented and tested for the Rust render layers API.

For the rust side the render layer API offers a lot of example code that could be used as reference: https://github.com/Vector35/binaryninja-api/blob/dev/rust/src/render_layer.rs

See the C++ impl of language representation as well: https://github.com/Vector35/binaryninja-api/blob/dev/languagerepresentation.cpp

rbran:


emesare:
Merged with https://github.com/Vector35/binaryninja-api/commit/4d8feb4921cde4b70bb6db4ccd525a69205dee7f

PR #6640, opened 2025-04-14
Add thumb2 mrc il

jonathanzetier:
This PR:
* Updates an outdated README (build instructions seem to be from the era when the arch was a separate repo).
* Makes a minor change to disassembly: using the same `rX` for register names as everywhere else did in `MRC`  disassembly rather than `RX`, which makes it possible to see the source of register values when a register is highlighted.
* Adds IL to the MRC/MRRC/MCR/MCRR (and *2 variants) in thumb mode.  Nothing too fancy here, the intrinsics were already implemented for ARM mode, so this was just bringing them over to thumb mode too.

### DIFF
--- a/arch/armv7/README.md
+++ b/arch/armv7/README.md
@@ -7,8 +7,7 @@ If you're interested in contributing when you submit your first PR, you'll recei
 
 ## Building
 
-Building the architecture plugin requires `cmake` 3.9 or above. You will also need the
-[Binary Ninja API source](https://github.com/Vector35/binaryninja-api).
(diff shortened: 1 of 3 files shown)


galenbwill:
Thanks for your submission. I will review this after the 5.0 release is done.

galenbwill:
Please accept the CLA. I will merge this as soon as you have. Thanks again for the PR!

galenbwill:
Merged and should be in builds >= `5.1.7551-dev`.

PR #6657, opened 2025-04-17
[SharedCache] Always load image dependencies

WeiN76LQh:
Simple change that means an image's dependencies can be loaded via the DSC triage view whether the primary image has already been loaded or not. Sometimes I've already loaded an image and then later on decide I want to load its dependencies. Currently there is no easy way to do that in the UI.


fuzyll (approved):
This looks good, should've written it this way to begin with.

plafosse:
Merged manually 5.0.7284

PR #6700, opened 2025-04-22
[ObjC] Use relative pointer types for members of `objc_method_entry_t`

### LINKED ISSUE #6699: Use proper relative pointer in objective-c types
The objective-c workflow uses a custom data renderer for relative pointers, that predates actual core support for them. We should migrate over to using the core supported relative pointers instead of through a custom data renderer.

More info from pubslack: https://binaryninja.slack.com/archives/C1E616ZSQ/p1745278384930859

https://github.com/Vector35/workflow_objc/blob/0df5ff8bdbbbebde8a56578491a2526439cd14fe/DataRenderers.cpp#L80-L102

bdash:
Relative pointer types have been a first class citizen for some time and are rendered correctly without any additional work.

The code was previously relying on the Objective-C workflow to render the `rptr_t` typedef. 

Fixes #6699.

### DIFF
--- a/objectivec/objc.cpp
+++ b/objectivec/objc.cpp
@@ -1264,9 +1264,6 @@ void ObjCProcessor::ProcessObjCData()
 	m_symbolQueue = new SymbolQueue();
 	auto addrSize = m_data->GetAddressSize();
 
-	m_typeNames.relativePtr = defineTypedef(m_data, {"rptr_t"}, Type::IntegerType(4, true));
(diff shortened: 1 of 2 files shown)


bdash:
I initially started on this change because I had expected it'd make cross references work for functions / variables that these fields point to. Sadly it does not (#6701).

0cyn:
Thank you so much for the PR and for having it rebased! Greatly appreciated! 

Merged with `cf9a634` and will be available soon in **versions _after_ `dev 5.1.7517`**

Issue #6290, opened 2024-12-30
Use .ppc.EMB.apuinfo section in powerpc ELFs to distinguish between APUs (Altivec, SPE, etc.)


cryptwhoa:
**What is the feature you'd like to have?**
There are several different powerPC variants with overlapping opcodes, notably Altivec and SPE, among others.  Currently, ELFs just pick the default one (Altivec), and there isn't an easy way to override that if it uses SPE, for example.

However, SPE ELFs exist (I wish I had one I could share), and it would be nice to be able to auto-detect them.  The way this is done is with the `.ppc.EMB.apuinfo` section in ELFs.  PowerPC could define a platform recognizer to dispatch to the correct APU variant, but it needs more information in the `Metadata` argument passed to it.  It would be nice to add sections to the `Metadata` argument so that this would work.

Note that at the time of writing, the SPE architecture is defined in `arch_ppc`, capstone isn't routed to correctly use it (the call to `cs_open` fails; try `Architecture["ppc_spe"].get_instruction_text(b"\x10\x00\x00\x00", 0x1000)` in the binaryninja console and it segfaults binaryninja on my box -- note that this isn't a "binaryninja isn't using capstone correctly" issue (other than not checking the call to `cs_open`), but a bug in capstone itself)...don't get too hung up on this, I'm hoping to push a PR with a replacement to capstone in a few weeks.

**Is your feature request related to a problem?**
A PowerPC SPE elf is basically unanalyzable by binaryninja right now.

**Are any alternative solutions acceptable?**
I'm sure there's a way to override the architecture of an ELF, but it would be nice not to have to.

**Additional Information:**
https://www.nxp.com/docs/en/engineering-bulletin/EB622.pdf

PR #6740, opened 2025-04-28
Add support for PowerPC VLE instruction set

cryptwhoa:
This adds support for PowerPC VLE to the homegrown PPC (ie non-capstone) branch.

Note that some instructions in the sample ELFs (see below) seem undecoded; this is due to issue #6290, because we don't support figuring out what type of vector instructions are supported yet, so the ELF recognizer doesn't know to decode them as SPE instructions.  I expect to add support for this in a future PR, but this is the VLE instruction set by itself.  Note that at the time I started writing this (maybe it's been fixed since then? probably not though if #6290 is still alive), it's a limitation of PPC binaries in general.

There was some grossness in getting the auto-VLE detection to work in ELFs.  The mechanism used to distinguish VLE instructions is through page bits in the MMU.  ELFs denote VLE sections the section's flag value.  Discussions on slack indicated that the best way forward was to pass data from the ELF view to a platform recognizer via `Metadata`.  Unfortunately, this seemed to require duplicating the section parsing logic.

Note that binaryninja's ELF loader doesn't seem to have a clean way to set architectures of different sections.  In theory, an ELF could have both VLE and non-VLE code sections in it, though I don't know how often these appear in practice.  The auto-detection logic only sets the architecture to VLE if all of the executable sections have the VLE flag.

Currently, the PR focuses on big-endian 32-bit VLE.  There are likely other architectures: VLE instruction decoding is always big endian, but the docs don't say anything about data endianness, so a little-endian 32-bit VLE is probably possible.  There can also be 64-bit VLE, of big/little endian.  When variants are added (SPE, Altivec, etc.), these architectures will start to balloon, for the whole matrix of:
* VLE vs. non-VLE
* 32-bit vs 64-bit
* big vs little endian
* SPE vs Altivec vs ???

so I'm punting on that complexity for #6290.

Datasheets:
[VLEPEM.pdf](https://github.com/user-attachments/files/19933726/VLEPEM.pdf) (the main document for VLE)
[VLE_addendum.pdf](https://github.com/user-attachments/files/19933729/VLE_addendum.pdf) (documentation for a few more instructions)

Sample ELFs can be generated from a container here: https://github.com/AutomotiveDevOps/nxp-devkit-mpc57xx-docker.  For convenience, here's a zipped up package of all the ELFs: 
[ppcvle_elfs.zip](https://github.com/user-attachments/files/19933725/ppcvle_elfs.zip)


plafosse:
Closing this PR as it was manually merged.

galenbwill:
Merged in versions >= 5.1.7754-dev

PR #6757, opened 2025-05-01
prefer fp, lr for x29, x30

### LINKED ISSUE #6512: Use special ARM64 register names in disassembly
The ARM64 architecture reserves `x29` and `x30` for special purposes: the frame pointer `fp` and the link register `lr`, respectively.

I think using these special names in the disassembly view would improve the experience.

For instance, here is the current disassembly of a classic function prologue:

```aarch64
sub     sp, sp, #0x30
stp     x29, x30, [sp, #0x20] {__saved_x29} {__saved_x30}
add     x29, sp, #0x20 {__saved_x29}
```

And here is what it would look like using special register names:

```aarch64
sub     sp, sp, #0x30
stp     fp, lr, [sp, #0x20] {__saved_fp} {__saved_lr}
add     fp, sp, #0x20 {__saved_fp}
```

yrp604:
- update tests
- drive by windows build improvement (docs here: https://learn.microsoft.com/en-us/windows/win32/msi/installation-context see ProgramFiles64Folder
- drive by readme path update

Closes #6512


galenbwill:
Thanks for the PR. Per our DM conversation on slack, I will need to update our internal test oracles in order to merge this, which I will do asap.

PR #6760, opened 2025-05-01
Implement idb_import dependency prompts.

rbran:
This also includes:
* A Til import command.
* Til import is run on the background with a progress message.


emesare:
This will be more approachable after https://github.com/Vector35/binaryninja-api/pull/7960, will need to circle back around to do a TIL to BNTL specifically, keeping this open until then

PR #6771, opened 2025-05-02
[SharedCache] Fix parsing of Objective-C metadata

bdash:
8267ab475cb2ac82fd3724c009f21e8e30143f6b introduced a logic error in `VirtualMemoryReader::ReadPointer` that caused `VirtualMemoryReader` to attempt to read pointers from the wrong address, breaking parsing of Objective-C metadata.

@emesare


emesare:
Oh wonderful! 😮‍💨 Thanks for the PR.

PR #6777, opened 2025-05-03
[SharedCache] Avoid reading header fields outside of the header for iOS 10 and older

### LINKED ISSUE #6073: DSCView::Init accesses potentially inexisting field when parsing platform
- Binary Ninja Version: 4.2.6335-dev Personal (1828c7e4)
- OS: macOS
- OS Version: 14.5 (23F79)
- CPU Architecture: arm64

**Bug Description:**

`DSCView::Init` reads `uint32_t platform` from hardcoded offset 0xd8 of the dsc header.  However, `struct dyld_cache_header` for older shared caches (say, iOS 10) is smaller than 0xd8 and does not include a platform field.  When loading such a dyld shared cache, Binary Ninja fails to initialize DSCViewAlpha and logs a misleading error message showing a garbage platform value.

**Steps To Reproduce:**

1. Download ipsw for some old iOS version, say, iOS 10.3
2. Extract the dyld shared cache from the system volume dmg in the ipsw
3. Open the dyld shared cache in Binary Ninja
4. See that DSCViewAlpha failed to initialize with "unknown platform" and a garbage value read from outside of the header

**Expected Behavior:**

Ideally, load the shared cache with DSCView fully functional (happy to assist with testing).

At a minimum though, `DSCView::Init` should not read fields that do not exist in the header, and fail with a more self-explanatory error message such as "dyld shared cache format unsupported" or some such.

There is a trick to figure out the size of the header struct:  Based on the assumption that the first mapping will immediately follow the header, we can look at `mappingOffset` and never read any header fields that extend beyond that offset.

```
[Default] Unknown platform: 24576
[Default] BinaryView of type 'DSCViewAlpha' failed to initialize!
```

**Binary:**

See repro.

**Additional Information:**

droe:
Use mappingOffset as an upper bound for the header size, and avoid reading any header fields from beyond that offset.

- Avoids printing garbage platform values for old (iOS 10 and older) DSCs.
- Avoids picking the wrong platform in case the bytes at the offset outside of the header just happen to have a value that is also a valid platform value, which is unlikely, but not impossible.

Fixes the remaining aspects of #6073.

Tested against 5.1.7363-dev (acd6c39c)

### DIFF
--- a/view/sharedcache/core/SharedCacheView.cpp
+++ b/view/sharedcache/core/SharedCacheView.cpp
@@ -175,9 +175,6 @@ bool SharedCacheView::Init()
 
 	m_logger = new Logger("SharedCache.View", GetFile()->GetSessionId());
 
-	uint32_t platform;
-	// NOTE: This entry only exists on ios 11 and later, older versions will just assume iOS.
-	GetParentView()->Read(&platform, 0xd8, 4);
 	char magic[17];
 	GetParentView()->Read(&magic, 0, 16);
 	magic[16] = 0;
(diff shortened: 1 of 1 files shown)


emesare:
LGTM

emesare:
Cherry picked on dev with https://github.com/Vector35/binaryninja-api/commit/c9253993ee4085880468cf9345519f45adfa77a3 to cleanup some of the old TODO comments and return false on unknown platform. Thanks for the PR!

PR #6785, opened 2025-05-04
[Rust] Avoid leaking a reference within {Medium,High}LevelILFunction::ssa_form

bdash:
Return a `Ref<_>` so that the underlying core object will have its reference count decremented when the caller drops the object.


emesare:
LGTM

PR #6789, opened 2025-05-04
Rust cleanup

### LINKED ISSUE #6778: Python bindings do not account for invalid UTF-8 strings
- Binary Ninja Version: 5.1.7320-dev (aa12054b)
- OS: macOS 15.4.1
- CPU Architecture: arm64

**Bug Description:**
`BnString::as_str` panics if the underlying string is not valid UTF-8. Many of these strings come from core which does not make any guarantees about their UTF-8ness. Symbol names, for instance, may not be valid UTF-8.

`BnString` is described as being analogous to `CString`. The similar methods on `CString` are failable:

```
pub fn into_string(self) -> Result<String, IntoStringError>
pub fn to_str(&self) -> Result<&str, Utf8Error>
```

IMO `BnString::as_str` should return `Result` and let the caller `unwrap()` if that's what they want.

It might be worth renaming `as_str` to `to_str` to match the naming on `CStr`.

### LINKED ISSUE #6784: [Rust] BnString::as_bytes_with_null does not include a nul byte
- Binary Ninja Version: 5.1.7320-dev (aa12054b)
- OS: macOS 15.4.1
- CPU Architecture: arm64

**Bug Description:**
`BnString::as_bytes_with_null` has no documentation, but based on its name I assumed it would include a trailing NUL. It doesn't. It returns a byte array of the string with the trailing NUL byte removed.

If it intends to include the NUL it should call through to `CStr::to_bytes_with_nul` rather than `CStr::to_bytes`.

It wouldn't hurt to update the method name to be consistent with the Rust standard library either (`as` -> `to` and `null` -> `nul`).

### LINKED ISSUE #6817: [Rust] Support SSA form of LLIL
Support for the working with the SSA form of LLIL via the Rust API.

rust/src/low_level_il has some references to the various SSA instructions and expression types which made me think it worked, but it doesn't handle them correctly. That's not as big a problem as it seems though as there doesn't appear to be any way to get the SSA form of a function.

There's a lot of analysis that's challenging to implement without the SSA form.

emesare:
1. Remove architecture trait bound on LowLevelILFunctions (this then impacts other types)
2. Remove infallible BnString conversions
3. Misc cleanup

For 1 I think this is pretty easy migration for users, they will just need remove the extra bound, build errors are descriptive enough.

For 2 this just means that users are more responsible for handling invalid UTF-8 which is the correct call in the cases where we cant already just make that assumption, or that a lossy conversion is acceptable.

I need to move on to some other tasks so this PR will likely not grow, I am going to mark this ready for review.

CC: @bdash @rbran 

Related: https://github.com/Vector35/binaryninja-api/issues/6784 https://github.com/Vector35/binaryninja-api/issues/6778


rust/src/string.rs
```diff
@@ -69,11 +69,26 @@ impl BnString {
         }
     }
 
+    /// Take an owned core string and convert it to [`String`].
+    ///
+    /// This expects the passed raw string to be owned, as in, freed by us.
```
bdash, line 78: This should probably document the behavior when the input contains non-UTF-8 characters.

rust/src/string.rs
```diff
@@ -69,11 +69,26 @@ impl BnString {
...
+    pub unsafe fn into_string(raw: *mut c_char) -> String {
+        Self::from_raw(raw).to_string()
```
bdash: This confused me for a bit. Is there a reason this calls `to_string` rather than directly calling `to_string_lossy`? IMO the function name should also indicate that this is a lossy conversion.

It looks like `to_string` is available on `BnString` due to the blanket impl of `ToString` for any type implementing `Display`.  `BnString`'s `Display` impl delegates to `to_string_lossy`.

Having `to_string` be available but effectively aliased to `to_string_lossy` seems surprising. It might be better to follow `CString`'s example and not implement `Display` or `ToString`.

rust/src/string.rs
```diff
@@ -87,34 +102,11 @@ impl BnString {
         mem::forget(value);
         res
     }
-
-    pub fn as_str(&self) -> &str {
```
bdash, line 91: I guess the idea is that these methods are available via deref to `CStr` so there's no need to wrap them?

rust/src/lib.rs
```diff
@@ -299,31 +299,33 @@ pub fn set_bundled_plugin_directory(new_dir: impl AsRef<Path>) {
...
 }
 
-pub fn settings_file_name() -> PathBuf {
+pub fn settings_file_path() -> PathBuf {
     let settings_file_name_ptr: *mut c_char = unsafe { BNGetSettingsFileName() };
     assert!(!settings_file_name_ptr.is_null());
-    let bn_settings_file_name = unsafe { BnString::from_raw(settings_file_name_ptr) };
-    PathBuf::from(bn_settings_file_name.to_string())
+    let settings_file_path_str = unsafe { BnString::into_string(settings_file_name_ptr) };
+    PathBuf::from(settings_file_path_str)
```
bdash, line 319: File system paths are not required to be UTF-8 on all platforms. I don't know whether it matters for these sorts of paths, but it could be handled by going from a `BnString` to an `OsStr` then to a `PathBuf` rather than via `String`.

rust/src/section.rs
```diff
@@ -90,12 +90,12 @@ impl Section {
         SectionBuilder::new(name, range)
     }
 
-    pub fn name(&self) -> BnString {
-        unsafe { BnString::from_raw(BNSectionGetName(self.handle)) }
+    pub fn name(&self) -> String {
```
bdash: If sections are named with the sections taken from the loaded binary, I don't think there's any guarantee they'll be UTF-8 names.

Doing a lossy conversion here will make it so that `view.section_by_name(section.name()))` fails. That doesn't seem great.

rust/src/headless.rs
```diff
@@ -71,7 +71,7 @@ pub fn init() -> Result<(), InitializationError> {
 ///
 /// ⚠️ Important! Must be called at the end of scripts. ⚠️
 pub fn shutdown() {
-    match crate::product().as_str() {
+    match crate::product().to_string().as_str() {
```
bdash: Isn't `crate::product()` already a `String`?

rust/src/string.rs
emesare: Not having the display impl is kinda annoying, maybe we could get away with demoting it to a debug impl that is lossy?

bdash: The `Debug` impl is sufficient for logging with `{:?}`. There's a lot of types that require that rather than the `{}` that `Display` allows. It could be a little annoying to adjust to, but it still seems better than silently doing a lossy conversion.

emesare: Ok I replaced the Display impl with Debug impl with https://github.com/Vector35/binaryninja-api/pull/6789/commits/90062c77233d2e9650a93b0cbc5985c84ade1c90

emesare:
Threw [this](https://github.com/Vector35/binaryninja-api/pull/6789/commits/b3aff171eec118c1517cc150837c7dbe7d001441) on the branch, still need to run more tests, I wouldn't spend time looking at it right now.

emesare:
CC @mkrasnitski I added you as a co-author to https://github.com/Vector35/binaryninja-api/pull/6789/commits/e66b8785c211f69e384978d1c9b5435a5e6dcbb1 seeing as it is effectively a continuation of your PR. Let me know if that is alright, and also if you have any input into this PR.

mkrasnitski:
Why not just merge #5897? There are no merge conflicts and then you can rebase this PR on top without the string commit.

I did a local rebase to test things, and it also let me compare your string commit with mine. They do essentially the same thing except you add a couple lines to the docstring for `BnString`. ~~However, you also left all the leftover `as *const c_char` casts which are unnecessary because `CStr::as_ptr` already returns `*const c_char`. Whereas, my PR removes them which improves formatting in a lot of places by getting rid of line breaks.~~ EDIT: I didn't see that you had that as a separate commit, my bad.

rust/src/architecture.rs
```diff
@@ -1404,7 +1404,8 @@ impl CoreArchitecture {
     }
 
     pub fn by_name(name: &str) -> Option<Self> {
-        let handle = unsafe { BNGetArchitectureByName(name.to_cstr().as_ptr() as *mut _) };
+        let name = name.to_cstr();
```
emesare, line 1407: Premature drop possible

rust/src/string.rs
```diff
@@ -186,110 +183,124 @@ unsafe impl CoreArrayProviderInner for BnString {
...
 }
 
-pub unsafe trait BnStrCompatible {
-    type Result: AsRef<[u8]>;
+pub unsafe trait AsCStr {
+    type Result: Deref<Target = CStr>;
 
-    fn into_bytes_with_nul(self) -> Self::Result;
+    fn to_cstr(self) -> Self::Result;
 }
```
mkrasnitski: Having `to_cstr` take ownership can lead to some footguns around pointers to temporaries the same way `BnStrCompatible` did. In my PR, I have `as_cstr` borrow `&self` so that writing `as_cstr().as_ptr()` inline isn't  a problem except if we're dealing with iterators of string-like objects where a local collection of `&CStr` needs to be created first, and then another collection of pointers. That also lets me return a concrete `Cow<'_, CStr>` given that I have access to the lifetime `&self`, instead of requiring an associated type.

rust/src/string.rs
mkrasnitski: I think this is an antipattern in this case. The `Deref` implementation for `BnString` requires a linear scan, whereas for things like `String/&str` it's a constant-time operation. Deref should be reserved for smart pointers or constant-time pointer operations, and other uses are heavily warned against: https://rust-unofficial.github.io/patterns/anti_patterns/deref.html

rust/src/string.rs
emesare: The Cow lifetime does not affect how long the owned memory lives:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7924a5caad39f442aab62c6eba1f7ad1

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e6271f359b299aa003268326766603c1

`bad` is still bad with your PR, it will free the underlying memory even if you tie the lifetime of &self to Cow.

PR #6790, opened 2025-05-04
[ObjC] Improve naming of arguments to Objective-C methods

bdash:
Detect and remove common prefixes used on selectors so that their arguments are more naturally named. For instance, a selector of `initWithURL:withStagedURL`: now ends up with arguments named `URL` and `stagedURL`, rather than `initWithURL` and `withStagedURL`.

### DIFF
--- a/objectivec/objc.cpp
+++ b/objectivec/objc.cpp
@@ -6,27 +6,57 @@ using namespace BinaryNinja;
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


0cyn:
Thank you! Merged with 425a1be and expected to be in **builds >= `dev 5.1.7537`**

Issue #6779, opened 2025-05-03
It should be easier to add additional activities to Objective-C and shared cache workflows


bdash:
**Version and Platform (required):**
- Binary Ninja Version: 5.1.7363-dev (acd6c39c)
- OS: macOS 15.4.1
- CPU Architecture: arm64

**Bug Description:**
There is currently no good way for a plug-in to add function activities to run in all Mach-O binaries (without Objective-C, with Objective-C, or that are loaded from the shared cache).

The `core.function.objectiveC` and `core.function.sharedCache` workflows both clone from `core.function.baseAnalysis` so they do not pick up any modifications a plug-in may make to `core.function.metaAnalysis`.

**Additional Information:**
There was some discussion [on Slack](https://binaryninja.slack.com/archives/C1E616ZSQ/p1745532210477839) about this. It sounds like the Objective-C and shared cache plug-ins could themselves modify `core.function.metaAnalysis` and make use of view type predicates to ensure their activities run only in their views.

Alternatively the workflows could instead clone `core.function.metaAnalysis`. It seems like there'd be an ordering issue there? That is, plug-ins would somehow need to ensure they're loaded and have modified `core.function.metaAnalysis` prior to the Objective-C or shared cache plug-ins creating their clones. I'm not sure there's any way to ensure that happens.

---

For now I work around this by explicitly registering activities in all three workflows. This requires that plug-ins declare an optional dependency on the `workflow_objc` and `sharedcache` plug-ins so that they are loaded and create their workflows before my plug-in is initialized. 

Sadly registering the activity multiple times while using automatic eligibility results in some log spew about the setting name:
```
[Default error] Setting: bdash.arm64e-pac already registered!
[Default error] Setting: bdash.arm64e-pac already registered!
```

bpotchik:
As far as I understand, all of these related workflows are using the meta workflow now, which means they should all be available without needing to grab a separate named workflow.

bdash:
The shared cache was updated in b0965fa to modify `core.function.metaAnalysis` rather than adding a new workflow. The Objective-C workflow has not yet been updated:

https://github.com/Vector35/binaryninja-api/blob/e96a7a078a9909fc1212861c9669e6af0b034e46/plugins/workflow_objc/Workflow.cpp#L292-L300

IIRC the challenge was around having it respect the `loader.macho.processObjectiveC` setting:

[...]

PR #6791, opened 2025-05-05
[SharedCache] Add activities to core.function.metaAnalysis rather than a new named workflow

bdash:
Activities are registered with a `viewType` predicate to ensure they only run in shared cache views.

Adding activities to `core.function.metaAnalysis` lets activities registered by plug-ins run in the shared cache without the plug-ins having to be aware of the shared cache.

Work towards #6779.


emesare:
LGTM, cherry picked with https://github.com/Vector35/binaryninja-api/commit/b0965fa2a5c703214f60a6c5af01f6ddf21e6f42

Issue #742, opened 2017-07-18
Function level lifting support


plafosse:
Some instruction sets don't easily lift one instruction at a time. For instance wasm, .Net, Java, I'm sure others. One possible solution to this is to allow lifting of the entire function at once rather than one instruction at a time as the current architecture plugins support.

alexrp:
> This is a blocker for Qualcomm Hexagon support.

Would you mind elaborating on why Hexagon needs this? I'm not familiar with the architecture and now I'm curious. 🤔

plafosse:
It is heavily pipelined. You need to be able to lift multiple instructions at once in order to appropriately translate the semantics.

rssor:
The AnalyzeBasicBlocks portion of this is implemented and on dev now. This means that for architectures that need control over the recursive descent/basic block recovery portion of the analysis (e.g., architectures with function headers) or similar should be possible to implement now. Hexagon should _also_ be implementable with these changes.

Although the lifting changes aren't implemented yet, for architectures that can cope with being invoked once per basic block (especially if the arch plugin already identified the basic blocks) and don't need to peer at data _except for_ that basic block, they can simply consume entire basic blocks in the lifting callback.

Work that remains:
- Maintaining context between basic block recovery and lifting and text rendering
- Allowing text tokens from architectures to contribute newlines
- Expand the lifter to allow it to look at data outside of the current basic block

What _can't_ be implemented until other work is completed:
- If lifting IL for a block requires data from a **different** block that is still unrepresentable

zznop:
New APIs have been added for performing custom control-flow recovery, lifting entire functions in one swoop, and sharing custom context between control-flow recovery, lifting, and disassembly stages. In architecture plugins you can now override `analyze_basic_blocks`, `lift_function`, and `get_instruction_text_with_context` to take advantage of this functionality. A Python example can be found here: https://github.com/zznop/bn-brainfuck/blob/shared_context/__init__.py

These same APIs are used by our default implementation of function lifting and basic block analysis, which can be found here: https://github.com/Vector35/binaryninja-api/blob/dev/defaultarch.cpp

Issue #551, opened 2016-11-17
architecture doesn't have reference to the BinaryView


lucasduffey:
This will make certain arches possible - like dex smali.

I need to be able to access bv.session_data.string_table from the Architecture class.

psifertex:
I suspect there's another way around this -- we make Thumb2/Arm work without it, can you clarify a bit more about why you need session data from architecture?

lucasduffey:
In smali the instructions reference strings and methods using an table index

My current idea: save the table in SessionData, and access it that way.

whitequark:
I would also like to see this addressed; the primary motivation is so that Python-based plugins could use settings defined for a Resource to change their disassembly format.

ccarpenter04:
I also have a use case for this, smali is a powerful IL that's used for jvm bytecode transformations and I've been working on jvm bytecode support for some time (on and off, work is crazy). It's quite simple to convert Dalvik and "Java bytecode" to smali and with this change (and a preprocessing step to perform the initial conversion), I believe both Dalvik classes and "Java" classes would be able to leverage the benefits of the different ILs offered (and it would be much more reasonable to do now that projects support has been added)

Cussrro:
Hopefully this issue can be improved, as it greatly limits the flexibility of developing decompilation plugins for unknown instruction sets.

Martyx00:
For a C166 architecture, this will allow proper decoding of sring references which are addressed in a specific bank context.

bdemick:
@xusheng6 suggested I weigh in here: I'd like to perform some limited instruction decoding (basically a limited linear sweep with no further analysis) to extract data from executable code in a flat binary file that references data needed to inform subsequent loading of the file.

Cussrro:
After so long, the most basic problem still hasn't been solved?

![Image](https://github.com/user-attachments/assets/2c2816d2-2472-4ecf-88a4-73103a57132a)

psifertex:
It might seem like a most basic issue but it requires some significant core changes to implement. We intentionally made some design decisions early on that made multithreaded analysis better/easier for some architectures but made it more difficult for architectures like this.

That said, I have good news, this issue is currently being worked on and we hope to have it resolved by the 5.1 release (still several months out though so no promises...)

zznop:
[...]

PR #6792, opened 2025-05-05
Allow architecture plugins to perform basic block analysis

zznop:
Moves `Function::AnalyzeBasicBlocks` from the core to the API and allows it to be overridden in architecture plugins for custom implementations, if desired. This PR addresses #742 and #551 by allowing architecture plugins to override `Architecture::AnalyzeBasicBlocks` to perform control flow analysis at the function level.


plafosse:
This was merged manually.

PR #6797, opened 2025-05-06
Correct typos in Function.add_tag docs. Fixes #6796

### LINKED ISSUE #6796: Documentation for Function.add_tag in Python is Incorrect
The documentation for the Function.add_tag API in Python is incorrect. It looks like someone cut and pasted the BinaryView.add_tag documentation, but the parameters are slightly different.

Also, the link to the other documentation has a typo, so it does not get rendered as an actual link.

Image

https://api.binary.ninja/binaryninja.function-module.html#binaryninja.function.Function.add_tag

utkonos:


psifertex:
looks good, much appreciated!

PR #6807, opened 2025-05-08
Implement a Pseudo Objective-C language representation

bdash:
This is implemented in the Pseudo C plug-in as it shares 99% of the logic. The Objective-C support is implemented in a subclass of `PseudoCFunction`. The handling of instruction types that the Objective-C representation needs to customize is extracted into virtual functions that the Objective-C subclass overrides.

This currently supports:
 * Rewriting `objc_msgSend` / `objc_msgSendSuper2` with constant selectors to `[receiver message]` notation.
 * Rewriting calls to `objc_alloc` / `objc_alloc_init` / `objc_new` to the equivalent message send notation.
 * Rewriting `objc_retain` / `objc_release` and friends to the equivalent message send notation.
 * Displaying Objective-C class references as their class names rather than `_OBJC_CLASS_$_` symbol names.
 * Displaying Objective-C string literals as `@"..."`.

This works best when used in conjunction with [bdash/bn-objc-extras](https://github.com/bdash/bn-objc-extras) as it supports eliminating the reference counting runtime calls that add so, so much clutter.

| Before | After (w refcounting) | After (w/o refcounting) |
|--------|-------|-------|
| objc-before  | objc-after-with-refcounting | objc-after |

### DIFF
--- a/lang/c/plugin.cpp
+++ b/lang/c/plugin.cpp
@@ -1,0 +1,30 @@
+#include "binaryninjaapi.h"
+#include "pseudoc.h"
+#include "pseudoobjc.h"
+
+using namespace BinaryNinja;
+
+extern "C"
+{
+	BN_DECLARE_CORE_ABI_VERSION
+
+#ifndef DEMO_EDITION
+	BINARYNINJAPLUGIN void CorePluginDependencies()
+	{
+	}
+#endif
+
+#ifdef DEMO_EDITION
+	bool PseudoCPluginInit()
+#else
+	BINARYNINJAPLUGIN bool CorePluginInit()
+#endif
+	{
+		LanguageRepresentationFunctionType* type = new PseudoCFunctionType();
+		LanguageRepresentationFunctionType::Register(type);
+
+		type = new PseudoObjCFunctionType();
+		LanguageRepresentationFunctionType::Register(type);
+		return true;
+	}
+}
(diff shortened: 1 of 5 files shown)


plafosse:
This is very cool. We will look into merging this but I have some questions we need to discuss internally before. I think we probably would have preferred this to be in its own set of files rather than mixed in with the pseudo-c. Its understandable why as they share a lot of code but maybe it would have been best to copy the code. 
This also makes me think that we should add feature that would allow automatic selection of preferred language representation. i.e. if we detect that it's obj-c then the default language representation is pseudo-objc if we detect rust then pseudo-rust 🤔

bdash:
I'm not clear why you'd prefer that this copy the ~2,500 lines of code from the pseudo-c plug-in, rather than sharing the common code. Objective-C is very much a superset of the C language so sharing the code seems like the best solution.

Based on what I've seen with other cases in binaryninja-api where code has been copied rather than shared (Objective-C metadata and Mach-O parsing), the "shared" code will very quickly get out of sync as people will either not know or not want to take the time to make the same change in multiple places.

plafosse:
Yeah you're right after looking at your changes a bit more what you did makes more sense than copying the existing code.

0cyn:
Thank you _so_ much for this PR, this has been a desired feature for a very long time! We've merged this PR with `40a395b` and expect it to ship in **dev builds >= `5.1.7527`**

PR #6808, opened 2025-05-08
[ObjC] Retrieve category's class name when class is an external symbol

bdash:
This also updates the symbol names that are generated when the name cannot be found to include the address in hex rather than decimal so it's easier to navigate to.


emesare:
LGTM

PR #6814, opened 2025-05-08
[ObjC] Add an explicit reference to a method impl from its selector

bdash:
This makes it easier to see the possible message send targets when looking at a call to `objc_msgSend`.


0cyn:
Thanks for the PR! Merged with 8ef932b and will be available with **builds _after_ `5.1.7554`**

Issue #6522, opened 2025-03-24
Improve 32bit vtable identification for Itanium RTTI


emesare:
After recent changes to the Itanium RTTI processing we do now get most of the info however it seems that vtables are still being picked up only sometimes, this is likely due to the offset not being attributed to some base class in the class information.

<img width="949" alt="Image" src="https://github.com/user-attachments/assets/d6a6f80e-d62a-4dd4-b993-920e815b8cbc" />

In the image you can see the VMI class types vtables not being picked up.

emesare:
A binary with 32bit Itanium RTTI can be found here: https://github.com/Vector35/binaryninja-api/issues/6525#issue-2943892882

PR #6819, opened 2025-05-09
[RTTI] Handle cxxabi vtables being referenced via RELOC_COPY

bdash:
32-bit ELF binaries that are dynamically linked to the C++ runtime may use a copy relocation for the vtable. The vtable itself will be defined in the `.bss` section, and the copy relocation will cause the dynamic linker to populate it at load time from the C++ runtime library.

Detect this by looking for a symbol pointing to the start of the vtable data, two pointers before the vtable address.


emesare:
I would imagine this would resolve https://github.com/Vector35/binaryninja-api/issues/6522 or atleast partially resolve, IIRC there was another issue with the VMI class type overflowing into other data.

PR #6823, opened 2025-05-10
cmake: Update minimum CMake version to 3.10

ekilmer:
This silences the following deprecation warning when running with CMake 3.31+

```
CMake Deprecation Warning at examples/bin-info/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.
```

Also remove the `FATAL_ERROR` keyword because it's ignored by CMake 2.6 and higher.

https://cmake.org/cmake/help/v3.31/command/cmake_minimum_required.html

The following script was used to make the changes

```bash
#!/bin/bash

# Find all files tracked by git
git_files=$(git ls-files)

# Loop through each file and perform the replacement
for file in $git_files; do
  # Check if the file contains the pattern we're looking for
  if grep -q "cmake_minimum_required(VERSION 3.9" "$file" 2>/dev/null; then
    echo "Updating $file"
    # Use sed to replace the entire line containing the pattern
    sed -i 's/cmake_minimum_required(VERSION 3.9.*/cmake_minimum_required(VERSION 3.10)/' "$file"
  fi
done

echo "Finished updating CMake minimum version requirements"
```

### DIFF
--- a/arch/arm64/CMakeLists.txt
+++ b/arch/arm64/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.9 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.10)
 
 project(arch_arm64 CXX C)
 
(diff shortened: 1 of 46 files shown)


ekilmer:
Fixed by 1408212717861485772985940474034078282829

PR #6838, opened 2025-05-16
[ObjC] Fix another case where a category's class name was not being resolved

bdash:
Handle the case where the `class` field of the category is a direct reference to an external symbol via a bind fixup. In that case, the pointer's value when read from the view is 0. To determine the class name it is necessary to look up the relocation for the pointer's address and parse the class name from the symbol's name.

### DIFF
--- a/objectivec/objc.cpp
+++ b/objectivec/objc.cpp
@@ -1,7 +1,32 @@
 #include "objc.h"
 #include "inttypes.h"
+#include <optional>
 
 using namespace BinaryNinja;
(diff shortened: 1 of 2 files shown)


0cyn:
Thank you for the PR! Merged with [fa8ecc1](https://github.com/Vector35/binaryninja-api/commit/fa8ecc11840135ae2050b8af16373d3e3ffeebbc), this will be available in versions >= `dev 5.1.7535`

PR #6839, opened 2025-05-16
[Merged] [MachO] Don't add symbols when processing bind fixups

### LINKED ISSUE #5351: Macho loader treating symbol bindings wrong
- Binary Ninja Version: 4.1.5218-dev Personal (423fe6c2)
- OS: MacOS

**Bug Description:**
I think there is some misconception of how the macho bindings works, as instead of putting the value that can be located by the external symbol to specified address, current BNs macho loader defining the symbol on address instead.

**Steps To Reproduce:**
1. Open attached binary
2. Go to `0x100008310`

**Expected Behaviour:**
As i've said in the description, there should be not defined symbol (`void* SYMBOL: ...`), but the external value should be stored on address (`void* data_XXX: SYMBOL`).

[test_binary.zip](https://github.com/Vector35/binaryninja-api/files/15190364/test_binary.zip)

bdash:
The target of the bind opcode has its symbol applied when the dynamic symbol table is parsed.

This was previously adding a symbol at the fixed-up address with the name of the symbol being bound. This was resulting in dozens of `_OBJC_CLASS_$_NSObject`, `_OBJC_METACLASS_$_NSObject`, and `__objc_empty_cache` symbols being created: one for each `objc_class` that referenced those symbols.

Fixes #5351.

### DIFF
--- a/view/macho/machoview.cpp
+++ b/view/macho/machoview.cpp
@@ -2900,8 +2900,6 @@ void MachoView::ParseDynamicTable(BinaryReader& reader, MachOHeader& header, BNS
 					if (name == NULL)
 						throw MachoFormatException();
 
-					DefineMachoSymbol(symtype, string(name), address, binding, true);
-
 					memset(&externReloc, 0, sizeof(externReloc));
 					externReloc.nativeType = BINARYNINJA_MANUAL_RELOCATION;
 					externReloc.address = address;
@@ -2915,8 +2913,6 @@ void MachoView::ParseDynamicTable(BinaryReader& reader, MachOHeader& header, BNS
 				case BindOpcodeDoBindAddAddressULEB:
 					if (name == NULL)
 						throw MachoFormatException();
-
-					DefineMachoSymbol(symtype, string(name), address, binding, true);
 
 					memset(&externReloc, 0, sizeof(externReloc));
 					externReloc.nativeType = BINARYNINJA_MANUAL_RELOCATION;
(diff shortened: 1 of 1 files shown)


0cyn:
I am largely in favor of this landing, however we _do_ lose some symbols in IL because of this. The current code is a long-standing holdover for this reason.

Screenshot 2025-06-04 at 10 34 35 AM

Occasionally it is able to resolve the pointers although that is analysis based. I think a good solution here may be to instead define a `'__ptr_'+name` symbol at this location? That way we can avoid IL quality degrading while making things more accurate.

bdash:
The current IL is wrong so IMO the change shown in the screenshot is an improvement. The argument to `dispatch_once` isn't `__NSConcreteGlobalBlock`, it's a pointer to a static struct whose first member is `__NSConcreteGlobalBlock`.

0cyn:
I picked a bad example 😄 but your point stands. 

Still diffing ILs on several binaries for this in different conditions, but assuming that's the only case of this I can find, this should be able to land. Most other cases (e.g. `dyld_stub_binder`) seem to resolve in HLIL which is fine.

0cyn:
Merged with [3e7c1b6](https://github.com/Vector35/binaryninja-api/commit/3e7c1b6a41c666f227237d5751b29d3538aa68b3). 

I missed the `git push --force` while rebasing this for a few changes that got pushed to dev, so github is marking it as closed/unmerged, however it has landed. 

Available in versions >= `dev 5.1.7565`

PR #6887, opened 2025-05-27
[Rust] Allow mapping between SSA and non-SSA forms of LLIL instructions / expressions

bdash:
For `LowLeveLIL{Expression,Instruction}<M, NonSSA>` this adds:

```rust
pub fn ssa_form<'a>(&self, ssa: &'func LowLevelILFunction<M, SSA>) -> …
```

For `LowLeveLIL{Expression,Instruction}<M, SSA>` this adds:

```rust
pub fn non_ssa_form(&self, non_ssa: &'func LowLevelILFunction<M, NonSSA>) -> …
```

One thing I wasn't sure about is having these validate that `self.function` and the passed-in function represent the same underlying function object. Is there an easy way to verify that?

### DIFF
--- a/rust/src/low_level_il/expression.rs
+++ b/rust/src/low_level_il/expression.rs
@@ -104,6 +104,36 @@ where
     }
 }
 
+impl<M, R> LowLevelILExpression<'_, M, SSA, R>
+where
+    M: FunctionMutability,
+    R: ExpressionResultType,
+{
+    pub fn non_ssa_form<'func>(
+        &self,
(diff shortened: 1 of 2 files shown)


bdash:
The latest push fixed the clippy warnings the earlier iteration had introduced.

emesare:
Cherry picked on dev with https://github.com/Vector35/binaryninja-api/commit/82a211d4312a7ce1dd3e3800b15648620c65f48b. Thank you!

PR #6890, opened 2025-05-27
[Rust] Support retrieving use / def of SSA registers in LLIL

bdash:
This adds:

```rust
impl<M: FunctionMutability> Ref<LowLevelILFunction<M, SSA>> {
    pub fn get_ssa_register_uses<R: ArchReg>(
        &self,
        reg: LowLevelILSSARegisterKind<R>,
    ) -> Vec<LowLevelILInstruction<M, SSA>> { … }

   pub fn get_ssa_register_definition<R: ArchReg>(
        &self,
        reg: &LowLevelILSSARegisterKind<R>,
    ) -> Option<LowLevelILInstruction<M, SSA>> { … }
}
```

---

I'm not clear whether this handling of partial registers is correct. There's not a lot of information on their semantics and I couldn't immediately find how Python or C++ handle them.

### DIFF
--- a/rust/src/low_level_il/function.rs
+++ b/rust/src/low_level_il/function.rs
@@ -209,6 +209,57 @@ impl Ref<LowLevelILFunction<Mutable, NonSSA>> {
     }
 }
 
+impl<M: FunctionMutability> Ref<LowLevelILFunction<M, SSA>> {
+    /// Return a vector of all instructions that use the given SSA register.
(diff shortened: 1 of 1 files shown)


emesare:
Cherry picked on dev with https://github.com/Vector35/binaryninja-api/commit/abc355d98467bc1fce4368d4a7e1aff0e12cc3c3. Thank you!

PR #6922, opened 2025-06-04
Fix incorrect paths in EFI Resolver, and let it run in headless mode

mkrasnitski:
EFI Resolver was failing to apply protocol definitions for me, and I traced it down to incorrect calculation of the path to `types/efi.c`. On linux, the calculated path was `/opt/binaryninja/plugins../types/efi.c`, which obviously wouldn't get interpreted correctly. Adding the missing leading separator should fix the problem. The same problem likely appears on Windows, though I haven't confirmed this.

Another bug I found was that running in headless mode seems to not trigger the workflow, and this I traced down to `identifyModuleType` calling `bv->GetCurrentView()` instead of `bv->GetFile()->GetViewOfType("...")` when checking for a DXE/PEI file.

### DIFF
--- a/plugins/efi_resolver/include/ModuleType.h
+++ b/plugins/efi_resolver/include/ModuleType.h
@@ -13,10 +13,10 @@ enum EFIModuleType
 
 static inline EFIModuleType identifyModuleType(BinaryView* bv)
 {
-	std::string viewType = bv->GetCurrentView();
-	if (viewType == "Linear:PE")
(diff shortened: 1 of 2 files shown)


zznop:
Good catch. The new implementation of EFI resolver must not have been thoroughly tested on Windows or Linux. This morning I tested on MacOS, Windows, and Linux with these changes and everything looks good.

I'm letting a test build run through CI now to make sure unit tests pass and then I'll merge this.

PR #6923, opened 2025-06-04
Factor out branch metadata fields in BNInstructionInfo

mkrasnitski:
Introduces a new `BNBranchInfo` type that contains relevant info pertaining to a given instruction branch. This keeps related data together and simplifies handling instruction branches in other languages.


mkrasnitski:
Rebased this, should be fine for review now @emesare

emesare:
The reason (i believe) the core def is defined like that is to shrink the size of the structure, factoring out the branch info like that will increase the size of the structure from 72 to 96 bytes, because an API wraps this and its not exposed to the end user I am going to close this.

PR #6964, opened 2025-06-16
Add support for string wrapping in formatter

spoonmilk:
This adds support for String wrapping in the formatter up to the length specified in "Maximum String annotation size," allowing for the display of longer inline strings.

Currently, the scheme for wrapping is as follows:
- Wrap along spaces and punctuation when possible, but in cases of repeated punctuation/whitespace, treat it as one "atom" (smallest segment that can be wrapped)
- Treat format specifiers and newlines as singular atoms as to not break them up on line wrapping


spoonmilk:
Delaying until user testing

PR #7008, opened 2025-06-30
Add GetPathInProject() to core API

spoonmilk:
Adds GetPathInProject() function to core API, returning the path of a file in a project relative to the project rather than the actual location on disk.

### DIFF
--- a/python/project.py
+++ b/python/project.py
@@ -190,6 +190,22 @@ class ProjectFile:
 		"""
 		return core.BNProjectFileExport(self._handle, str(dest))
 
+	def get_path_on_disk(self) -> Optional[str]:
+		"""
+		Get this file's path on disk
+		
+		:return: The path on disk of the file or None
+		"""
+		return core.BNProjectFileGetPathOnDisk(self._handle)
+
+	def get_path_in_project(self) -> Optional[str]:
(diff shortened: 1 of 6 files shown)


spoonmilk:
Newest commit added `get_file_by_path_on_disk` and `get_file_by_path_in_project` to the python API bindings, along with fixing some logic.

python/project.py
```diff
@@ -650,6 +666,43 @@ def get_file_by_id(self, id: str) -> Optional[ProjectFile]:
...
+		:param path: Path of the file on the disk
+		:return: File with the requested path or None
+		"""
+		handle = core.BNProjectGetFileByPathOnDisk(self._handle, path)
+		if handle is None:
+			return None
+		file = ProjectFile(handle)
+		return file
+
+	def get_file_by_path_in_project(self, path: str) -> Optional[List[ProjectFile]]:
```
negasora: The return type should just be `List[ProjectFile]`

negasora (changes requested):
Also `GetFileByPathInProject` and equivalent should be pluralized (`GetFilesByPathInProject`) since they return a vector/list

negasora (approved):

PR #7036, opened 2025-07-03
[Rust] Fix clippy + rustfmt complaints

bdash:
The clippy complaint was a real issue: the doc comment intended for the commented-out `unresolved_indirect_branches` was being interpreted as the start of the doc comment for the following function.


emesare:
LGTM, merge whenever

Issue #6723, opened 2025-04-24
Loading ELF with no program headers and first section at non-zero address has segments/sections at wrong address


negasora:
In the attached binary, there are no program headers and the first section has address `0x13000` and size `0x13cc0`.
When we load this file we create this section at whatever load base was requested (`0x400000` by default) instead of the load base plus `0x13000`

This is the section table entry:
```
0000b270      [0x1] = 
0000b270      {
0000b270          uint32_t name = 0x1
0000b274          enum sh_type type = SHT_NOBITS
0000b278          enum sh_flags flags = SHF_ALLOC | SHF_EXECINSTR
0000b280          uint64_t address = 0x13000
0000b288          uint64_t offset = 0x0
0000b290          uint64_t size = 0x13cc0
0000b298          uint32_t link = 0x0
0000b29c          uint32_t info = 0x0
0000b2a0          uint64_t align = 0x1000
0000b2a8          uint64_t entry_size = 0x0
0000b2b0      }
```

And the resulting section/segments:

![Image](https://github.com/user-attachments/assets/cc6ef4b1-f94c-439e-854c-f1450570494a)

We should create the section/segment at `0x413000` instead of `0x400000`

[wrong_load_base.zip](https://github.com/user-attachments/files/19898762/wrong_load_base.zip)

negasora:
This might be as simple as switching calls from `BinaryView::GetStart()` to `BinaryView::GetImageBase()`

hongyu-dev:
This is what I currently have for this issue:

<img width="657" alt="Image" src="https://github.com/user-attachments/assets/470e3405-7917-4161-a25f-b60a0fc8b766" />

<img width="639" alt="Image" src="https://github.com/user-attachments/assets/65fc622b-b983-4771-8aab-0ef7e349dc9b" />

None of those functions have disassembly information and there are many `[Analysis error] Attempting to add function not backed by file: <some address>` Any feedback or any guidance for next steps for this issue would be appreciated.

hongyu-dev:
This fix has been merged check PR:https://github.com/Vector35/binaryninja-api/pull/7052 and [commit](https://github.com/Vector35/binaryninja-api/commit/c2a48b1ff770c092a9745eaa28092b89159beb4f#diff-d1146fa5b4c12afd61009d753c792960630a0639dbf7f87a8534d8796c2d9ff3)

PR #7052, opened 2025-07-07
Fix shared object start address without program header

hongyu-dev:
This PR is for issue: https://github.com/Vector35/binaryninja-api/issues/6723

Symbols before fix:
Screenshot 2025-07-07 at 2 16 22 PM

Symbols after fix:
Screenshot 2025-07-07 at 2 16 58 PM

Memory map before fix:
Screenshot 2025-07-07 at 2 17 15 PM

Memory map after fix:
Screenshot 2025-07-07 at 2 17 48 PM

The code before fix didn't include the situation for shared object file without program header case and section start address got overwritten by `segmentStart` which start with value 0.

Current fix creates segments based on sections start address.

### DIFF
--- a/view/elf/elfview.cpp
+++ b/view/elf/elfview.cpp
@@ -474,7 +474,7 @@ bool ElfView::Init()
 	{
 		if (settings->Contains("loader.imageBase"))
 			preferredImageBase = settings->Get<uint64_t>("loader.imageBase", this);
-
+			
 		if (settings->Contains("loader.platform"))
 		{
 			BNSettingsScope scope = SettingsAutoScope;
@@ -658,11 +658,10 @@ bool ElfView::Init()
 			semantics = ReadOnlyDataSectionSemantics;
 		else if ((m_elfSections[i].flags & ELF_SHF_WRITE) || In(sectionNames[i], readWriteDataSectionNames))
 			semantics = ReadWriteDataSectionSemantics;
-
 		if (m_elfSections[i].size != 0)
 		{
 			if (m_programHeaders.size() == 0)
-			{
+			{	
 				// We have an object file so we'll just create segments for the sections
 				uint32_t flags = 0;
 				if (semantics == ReadOnlyCodeSectionSemantics)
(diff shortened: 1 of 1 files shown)


fuzyll (changes requested):
I am not sure _what_ changes this needs, but I can't imagine this looks the way it's supposed to.

The Raw view shows a whole bunch of data that I assume is code:

image

But, the ELF view shows a bunch of bare symbols that map to 0's:

image

@negasora: Since you provided the original binary, are you able to give us any additional context on how you would like this to look after it's been loaded? I can't imagine we're supposed to have 0 functions and no actual code, given how the Raw view looks, but willing to change my mind if this is somehow expected.

EDIT: I _think_ the problem is that, while we've adjusted where the segments should be loaded, we are not also adjusting where we're looking for functions in this case? Like, the offset in the file needs to be adjusted _back_ for the amount we're 'off' due to loading at a later spot?

hongyu-dev:
I was very surprised with this binary too. `.text` section type is NOBITS that's why its all 0s in those functions.


Screenshot 2025-07-07 at 6 50 59 PM


My guess is this is CTF challenge that intentionally used malformed binary the `.text` section was marked as `NOBITS` instead of `PROGBITS`. This caused `objdump -d` shows no disassembly while `readelf -S` reveals section structure.  The actual code is embedded in the binary but hidden.  

https://github.com/Vector35/binaryninja-api/blob/768f7c78465fb93936e5ca50a0ca712664fe54e7/view/elf/elfview.cpp#L675

fuzyll:
With @negasora's help, I've tracked down where this came from, and it's an intentionally 'anonymized' binary from an Enterprise customer. Working with them to see if we can get something that will let us confirm this is fixed or to see if they would be willing to run a test build with these changes to confirm things load properly for them.

Sorry for the longer wait to get this merged, but this is on 'my' plate for the time being.

fuzyll:
Also, I confirmed that the sample, as given to us, is also messed up under `objdump`.

fuzyll:
Customer hasn't gotten back to me, pinged them again today and explained our upcoming code freeze deadline.

fuzyll:
Still uncomfortable shipping this without a real test case (which also means we can't add a unit test). I'm 95+% sure this is good to go, but I would still really rather have a customer with a real file tell us that definitively. As a result, we're going to hold off on merging this till after the release since the customer hasn't responded to us yet.

negasora:
This has been confirmed to fix the issue

fuzyll (approved):
As @negasora said in his comment, the customer got back to us and, upon using the test build we gave them, confirmed that this fixed their issue.

Still don't have a unit test for this, which isn't great, but since this passes ours and fixes the customer's problem, I am perfectly okay with shipping it.

Apologies for the wait and the earlier confusion over why it appeared to be failing.

PR #7057, opened 2025-07-08
[ObjC] Avoid leaking SymbolQueue and MachoObjCProcessor

bdash:
`SymbolQueue`'s lifetime was being managed via `new` / `delete`. This was error-prone in the face of code that can a) throw exceptions, or b) return early. Typically the solution would be to move to `std::unique_ptr` and call it a day, but nothing is ever easy.

The `SymbolQueue`s created by `ObjCProcessor` are intended to live for the duration of `ProcessObjCData` and `ProcessCFStrings` methods. Since they are used multiple levels down the call tree, the active `SymbolQueue` is stored as a member variable on `ObjCProcessor`. Switching to `std::unique_ptr` would ensure that the `SymbolQueue` is destroyed, but would leave a dangling pointer in the member variable.

To address this I'm introducing a `ScopedSingleton` class that provides a thread-local singleton whose lifetime is controlled by a guard object. `ProcessObjCData` / `ProcessCFStrings` call `ScopedSymbolQueue::Make` and store the returned guard object in a local. Code that accesses the symbol queue uses `ScopedSymbolQueue::Get` to retrieve the current instance. The guard object ensures the `SymbolQueue` is deleted and the `current` pointer is cleared no matter how the scope is exited.

A better longer-term design is to introduce a class for processing the Objective-C runtime metadata and a class for processing constants such as `CFString`s. These could directly own the symbol queue so it would be accessible to any member functions that define symbols, while inherently having the correct lifetime. This is a more involved refactoring than I have time for right now.

---

I'm not in love with this approach, but it was cleaner than the alternatives I experimented with:

1. Remove `m_symbolQueue` in favor of local `SymbolQueue` variables in `ProcessObjCData` / `ProcessCFStrings`. This works, but requires threading the object through virtually every function in the class so is quite unpleasant.
2. Make `m_symbolQueue` a `std::unique_ptr` and use a `scope_exit` type within `ProcessObjCData` / `ProcessCFStrings`. This works but it feels odd to have it be a member that's only set for the duration of a single method , which is how I ended up at `ScopedSymbolQueue`.

I'd probably have gone with 2 if there was an existing `scope_exit` type to use, but if I'm adding my own type for this purpose then it's nice to avoid the weirdness.

### DIFF
--- a/objectivec/objc.cpp
+++ b/objectivec/objc.cpp
@@ -2,9 +2,59 @@
 #include "inttypes.h"
 #include <optional>
 
+#define RELEASE_ASSERT(condition) ((condition) ? (void)0 : (std::abort(), (void)0))
+
 using namespace BinaryNinja;
 
 namespace {
+
+	// ScopedSingleton is a thread-local singleton that allows for scoped
+	// instantiation and destruction of an object. It is useful for managing
+	// resources that should only exist during a specific scope, but where it
+	// would be inconvenient to pass the object around explicitly.
+	//
+	// Calling `Make` initializes the thread-local singleton and returns a `Guard`
+	// object. When the `Guard` object goes out of scope, the singleton is destroyed.
+	template <typename T>
+	class ScopedSingleton
+	{
+		static thread_local T* current;
+
+	public:
+		class Guard
+		{
+			friend class ScopedSingleton;
+			Guard() = default;
+
+		public:
+			~Guard()
+			{
+				delete current;
+				current = nullptr;
+			}
+			Guard(Guard&&) = default;
+			Guard(const Guard&) = delete;
+			Guard& operator=(const Guard&) = delete;
+		};
+
+		static T& Get()
+		{
+			RELEASE_ASSERT(current);
+			return *current;
+		}
+
+		static Guard Make()
+		{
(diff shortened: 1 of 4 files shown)


emesare:
I am fine with this approach, the symbol queue construct itself is not pretty and IMO the entire idea should go away, I don't think constructing a lambda for every symbol object is very sane (although I would need to see some more profiling before i can say delete it!), but that is besides the point.

Looking around at some of the usage of the processor I see we do not properly free it in the macho view:
https://github.com/Vector35/binaryninja-api/blob/68672b4438a962e837067c719188995844746200/view/macho/machoview.cpp#L2370-L2395

more so we keep the "processor" around in a member variable of the view, that hardly seems appropriate.

bdash:
Yeah, it looks like `m_objcProcessor` is a member variable for the same reason `m_symbolQueue` was... some functions that are called a few levels deep need to poke at it. It's not used in anywhere near as many places so threading it through via function parameters isn't too bad. I'll push a new commit to take care of that.

`MachoView` also has its own `m_symbolQueue` member that should be cleaned up too. It can't currently leak so I'm leaving it as-is.

PR #7071, opened 2025-07-10
Fix the Triage Summary view's exports list to only update when visible

bdash:
The `ExportsTreeView` now asks the model to pause / resume updates when its visibility changes. When paused, all notifications from the view are ignored. When resumed, notifications set a flag to indicate that an update is needed and resume the update timer if it is not already active. The timer is stopped after an update is processed.

There's some extra complexity here to avoid emitting a signal for every `BinaryView` notification that is processed. These notifications are typically generated on background threads. The overhead of emitting a signal and it being routed to the main thread adds up given the number of notifications involved when loading a large binary. This manifests on macOS as time being spent inside `CFRunLoopWakeUp`.

### DIFF
--- a/examples/triage/exports.cpp
+++ b/examples/triage/exports.cpp
@@ -25,13 +25,10 @@ GenericExportsModel::GenericExportsModel(QWidget* parent, BinaryViewRef data): Q
 	}
 
 	m_updateTimer = new QTimer(this);
-	m_updateTimer->setSingleShot(true);
 	m_updateTimer->setInterval(500);
 	connect(m_updateTimer, &QTimer::timeout, this, &GenericExportsModel::updateModel);
-	connect(this, &GenericExportsModel::modelUpdate, this, [=, this]() {
-		if (m_updateTimer->isActive())
-			return;
-		m_updateTimer->start();
+	connect(this, &GenericExportsModel::updateTimerOnUIThread, this, [=, this]() {
+		updateTimer(m_needsUpdate);
 	});
 
 	m_data->RegisterNotification(this);
@@ -49,6 +46,10 @@ GenericExportsModel::~GenericExportsModel()
 
 void GenericExportsModel::updateModel()
 {
+	if (!m_needsUpdate)
+		return;
+
+	setNeedsUpdate(false);
 	beginResetModel();
 	m_allEntries.clear();
 	for (auto& sym : m_data->GetSymbolsOfType(FunctionSymbol))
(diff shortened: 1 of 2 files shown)


examples/triage/exports.cpp
```diff
@@ -237,40 +238,80 @@ void GenericExportsModel::setFilter(const std::string& filterText)
...
+	// given how frequently we receive notifications, so we only
+	// emit a signal if we didn't already need an update.
+	if (!m_needsUpdate.exchange(true))
+		emit updateTimerOnUIThread();
+}
 
 void GenericExportsModel::OnAnalysisFunctionAdded(BinaryNinja::BinaryView* view, BinaryNinja::Function* func)
 {
-	emit modelUpdate();
+	onBinaryViewNotification();
```
plafosse: As this is for the Exports section is there a reason we're even catching this notification? I think this could be removed. Along with FunctionRemoved and FunctionUpdated

examples/triage/exports.cpp
```diff
@@ -237,40 +238,80 @@ void GenericExportsModel::setFilter(const std::string& filterText)
...
 }
 
 
 void GenericExportsModel::OnSymbolAdded(BinaryNinja::BinaryView* view, BinaryNinja::Symbol* sym)
 {
-	emit modelUpdate();
+	onBinaryViewNotification();
```
plafosse: This would probably be even better
```
if ((sym->GetBinding() == GlobalBinding) || (sym->GetBinding() == WeakBinding))
   onBinaryViewNotification();
```

bpotchik:
Implementation looks good.

Additional context (informational only):

`SidebarWidgets` and `Views` include a quiescing interface (`enableRefreshTimer`) that provides a base timer and control signals for when widgets are shown/hidden in the UI. The current approach using showEvent/hideEvent works well for this scenario since the widget resets the model rather than collecting data from each `BinaryDataNotification`.

Additionally, it's possible to completely unregister from `BinaryViewNotifications` when the widget is hidden, then re-register and rebuild the model when it becomes visible again. I only mention this because, occasionally when opening a file and view navigation fails, the triage view is gratuitously created but never shown. It would then be an idle consumer of all notifications it registered for. There are some nuances to implementing this, so just providing it as information.

plafosse (approved):
I'm approving but I left a couple of comments on some potential improvements. Address those comments if you agree with them and then merge.

examples/triage/exports.cpp
bdash: Yeah, I'm not sure why these notifications were being observed. Since this is only concerned with symbols it seems like those should be sufficient. I'll remove these.

examples/triage/exports.cpp
bdash: Will do.

PR #7120, opened 2025-07-18
[PPC] Make assembler.cpp compile over 30 times faster

bdash:
Clang was taking around 30 seconds to compile assembler.cpp on my machine, with most of its time spent in code generation for the global initializer for `lookup`.

Changing the map key from `std::string` to `std::string_view` drops compile time to under a second. This is because the compiler no longer has to code gen the allocation, initialization, and moves of 2,100 `std::string`s. `std::string_view` is effectively free to initialize in comparison.

While I was here I made the map `static const`. It's not required for the build time improvement, but it is more correct and helped me check that no-one is mutating the map.

### DIFF
--- a/arch/powerpc/assembler.cpp
+++ b/arch/powerpc/assembler.cpp
@@ -43,7 +43,7 @@ struct info {
 	uint32_t mask; /* which bits to mutate */
 };
 
-map<string, info> lookup = {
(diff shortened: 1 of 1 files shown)


plafosse (approved):
Approved to push after we've released 5.1 stable

PR #7123, opened 2025-07-18
[Rust] Make fields of LookupTableEntry public

ex0dus-0x:
Allows a user of the UIDF API to access the `LookupTableEntry` after value-set analysis.


emesare:
LGTM unfortunately we are releasing a stable right now so I cannot merge this onto dev until our stable branch is split off from dev.

PR #7125, opened 2025-07-18
Update support matrix for kernel/shared cache

emesare:
I believe we have previously tested both kernel cache and shared cache for both macOS and iOS beta builds, I figure it be good to update the support matrix.


0cyn:
Have confirmed these worked

plafosse (approved):

PR #7142, opened 2025-07-21
Misc WARP fixes for 5.1 stable

emesare:
These were issues caught during testing, all of which should be included in the stable.


emesare:
Merged with https://github.com/Vector35/binaryninja-api/commit/df0d8aaeeed4be4d3aa9c8545088646e6c95e2ef

PR #7144, opened 2025-07-21
Fix crash on R_MIPS_HI16 relocs caused by bad cast

zznop:
Fix for https://github.com/Vector35/binaryninja/issues/1131


plafosse (approved):

plafosse:
Merged

PR #7145, opened 2025-07-21
Fix read one past the end of token list in the line formatter

D0ntPanic:
This showed as a heap read overflow while running tests under ASAN.

This is unlikely to actually crash, but it is definitely possible depending on heap implementation.


fuzyll:
Merged in Vector35/binaryninja@2b41abba988ddadbfe520bcbf5c06344037a2f71.

PR #7153, opened 2025-07-22
Add on_action_executed_immutable to python API

spoonmilk:
Adds on_action_executed_immutable, for a python-based version of OnActionExecuted that omits the problematic std::function& parameter.


plafosse:
Added in  5.1.7999

PR #7174, opened 2025-07-23
Generate WARP function guids for mach-o objective-c binaries

### LINKED ISSUE #7172: Warp function guids not created in objective-c binaries
When you open an objective-c binary we change the function workflow off of the meta workflow and to another objectivec specific one, one which does not have the WARP registered function activities. This culminates in the following error:

```
[WARP] Trying to match with an older database, please reanalyze the database
```

Related: https://github.com/Vector35/binaryninja-api/issues/6779

emesare:
Previously we only added the warp function activities to the meta workflow, however objective-c plugin currently registers its own workflow that we must add the activities to as well.

Fixes https://github.com/Vector35/binaryninja-api/issues/7172

### DIFF
--- a/plugins/warp/src/plugin.rs
+++ b/plugins/warp/src/plugin.rs
@@ -12,9 +12,9 @@ use binaryninja::background_task::BackgroundTask;
 use binaryninja::command::{
     register_command, register_command_for_function, register_command_for_project,
 };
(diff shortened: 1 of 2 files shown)


plafosse:
Fixed in 5.1.8005

PR #7200, opened 2025-07-29
[CMake] Specify CONFIGURE_DEPENDS for all file(GLOB ...)

bdash:
This ensures that CMake detects when files that match the glob are added or removed.


CouleeApps (approved):
Looks good in this repo. We should do the same in the various other repos (core, debugger, etc)

bdash:
Yeah, I have corresponding changes prepared for the other repositories.

PR #7201, opened 2025-07-29
Add dependency tracking to the Python code generation

bdash:
This ensures that the Python source files are only generated and copied into the output directory if inputs have changed, rather than being done unconditionally.


CouleeApps (approved):
LGTM. Similar comment as to the last PR about including the debugger repo.

PR #7209, opened 2025-07-30
[ARM64] Fix lifting of tbz / tbnz for bits >= 32

### LINKED ISSUE #7205: tbnz condition generates incorrect mask
- Binary Ninja Version: 5.2.8032-dev (4177d06d)
- Edition: Commercial
- OS: macOS
- OS Version: 26.0 Beta (25A5306g)
- CPU Architecture: M1

**Bug Description:**
Binary Ninja lifts `tbnz    x0, #0x3f, 0x18008e130` into `if ((x0 & 0xffffffff80000000) != 0) then 6 @ 0x18008e130` which is incorrect. The code should check the top (63rd) bit; the correct mask is 0x8000000000000000.

**Steps To Reproduce:**
1. Open `object_getClass` from libobjc.dylib in the dyld shared cache

**Expected Behavior:**
The check should test if the pointer is negative.

Image

**Binary:**
```
$ cat | clang -x c -shared -
#include <stdint.h>
#include <stdbool.h>
#include <stdlib.h>

void foo(int64_t i) {
	if (i < 0) {
		abort();
	}
}
```

[a.out.zip](https://github.com/user-attachments/files/21505865/a.out.zip)

**Additional Information:**

bdash:
Fixes https://github.com/Vector35/binaryninja-api/issues/7205.

Since the integer literal `1` fits in an int, that is its type. Shifting it by values larger than 31 give unexpected behavior in this context. I'm not entirely sure how we get `0xffffffff80000000` as the result. That doesn't match my reading of https://en.cppreference.com/w/cpp/language/operator_arithmetic.html#Built-in_bitwise_shift_operators. Prior to C++20 this was undefined behavior territory, but with C++20 it should be well defined.

The fix is to ensure that the value we are shifting is a `uint64_t`.

### DIFF
--- a/arch/arm64/il.cpp
+++ b/arch/arm64/il.cpp
@@ -1159,13 +1159,13 @@ static ExprId ExtractBits(
     LowLevelILFunction& il, InstructionOperand& reg, size_t nbits, size_t rightMostBit)
 {
 // Get N set bits at offset O
-#define BITMASK(N, O) (((1LL << nbits) - 1) << O)
+#define BITMASK(N, O) (((UINT64_C(1) << nbits) - 1) << O)
 	return il.And(REGSZ_O(reg), ILREG_O(reg), il.Const(REGSZ_O(reg), BITMASK(nbits, rightMostBit)));
 }
 
 static ExprId ExtractBit(LowLevelILFunction& il, InstructionOperand& reg, size_t bit)
 {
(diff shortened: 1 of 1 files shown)


plafosse (approved):
I looked at this this morning and identified this as the problem too. 👍

PR #7215, opened 2025-07-30
[ELF] Fix a memory leak when parsing .gnu_debugdata

bdash:
It is necessary to call `FileMetadata::Close` to break the reference cycle between the `BinaryView` and `FileMetadata`.

Fixes https://github.com/Vector35/binaryninja/issues/1060.


emesare:
Straightforward fix, LGTM!

plafosse (approved):

PR #7216, opened 2025-07-31
[Rust] Don't leak the callbacks within DownloadInstance::perform_custom_request

bdash:
I'm not entirely clear why this function copies the callbacks to the heap given that `BNPerformCustomRequest` is synchronous. Rather than changing that, I added an explicit `Box::from_raw` to ensure it is deallocated. This matches what is done within `perform_request`.

### DIFF
--- a/rust/src/download_provider.rs
+++ b/rust/src/download_provider.rs
@@ -253,6 +253,7 @@ impl DownloadInstance {
             )
         };
 
+        unsafe { drop(Box::from_raw(callbacks)) };
         if result < 0 {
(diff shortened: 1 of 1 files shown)


emesare:
I think we probably should just avoid boxing it altogether and pass it by ref into the function: https://github.com/Vector35/binaryninja-api/commit/966097bd3254231643ef36119a7e16359b9fd47c

bdash:
@emesare is going to fix the leak via https://github.com/Vector35/binaryninja-api/commit/966097bd3254231643ef36119a7e16359b9fd47c in the near future.

PR #7236, opened 2025-08-05
update idb_import idb-rs to 0.1.12

rbran:
Unfortunately I'm unable to test this on the latest dev release of Binary Ninja, it seems that something on the BinaryViewReader changed (likely `BinaryView::offset_valid`) and it broke all the versions of idb_import.

I'm investigating.

### DIFF
--- a/plugins/idb_import/src/lib.rs
+++ b/plugins/idb_import/src/lib.rs
@@ -1,5 +1,10 @@
 mod types;
-use idb_rs::{IDAKind, IDAUsize};
+use std::borrow::Cow;
+use std::io::{BufRead, Cursor, Seek};
+
+use idb_rs::id1::ID1Section;
+use idb_rs::id2::{ID2Section, ID2SectionVariants};
+use idb_rs::{IDAKind, IDAUsize, IDBFormat};
 use types::*;
 mod addr_info;
 use addr_info::*;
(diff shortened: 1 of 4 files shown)


rbran:
Seems like `self.bv.offset_valid(offset)` was returning true if `offset = [the file size]` previously, but now it returns false.

emesare:
We added a reader / writer impl recently in the rust bindings, maybe those work? I recall you asking for them @rbran 

https://github.com/Vector35/binaryninja-api/blob/cde1c38689188efbfd47cd9509beb348284d486b/rust/src/binary_view/reader.rs#L101-L138

https://github.com/Vector35/binaryninja-api/blob/dev/rust/tests/binary_reader.rs

rbran:
Humm...

I forgot about that, that wold had worked too.

plugins/idb_import/src/lib.rs
```diff
@@ -86,10 +86,9 @@ struct BinaryViewReader<'a> {
 impl std::io::Read for BinaryViewReader<'_> {
```
emesare, line 86: https://github.com/Vector35/binaryninja-api/blob/cde1c38689188efbfd47cd9509beb348284d486b/rust/src/binary_view/reader.rs#L101-L138

As per the discussion we had in this PR swap BinaryViewReader out for the one in bindings

rbran:
I'm unable to use `BinaryReader`, I get a `io error: Read out of bounds`

emesare:
Merged with https://github.com/Vector35/binaryninja-api/commit/8d4fcf6df521c56ed8332540fc0c03d64e342f96 apologies for the delay!

PR #7254, opened 2025-08-11
[Rust] Type-safe workflow and activity configuration

### LINKED ISSUE #6014: Improve workflow API ergonomics in Rust API
Specifically want this for the _rust api_.

Right now workflows are in two states: registered (immutable) and unregistered (mutable).

The distinction itself is hidden in the type system making it very easy for a user to mistake an API meant for unregistered workflows for one they can use on a registered workflow.

Adding a `WorkflowBuilder` or something that acts as a mutable workflow would allow us to use the type system to enforce correct behavior.

bdash:
This introduces `binaryninja::workflow::activity::Config` and related types that can be used for describing the configuration of workflow activity. This makes it possible to see what configuration options are available and eliminates errors due to typos in JSON string literals. Subsequent commits update the SVD and Warp plug-ins to use the new types.

I'm open to feedback on naming and exactly where this should live. The nested `activity` module isn't great, but it makes clear that this is for configuring activities and not the workflow more generally.

### DIFF
--- a/rust/src/workflow.rs
+++ b/rust/src/workflow.rs
@@ -9,9 +9,12 @@ use crate::low_level_il::{LowLevelILMutableFunction, LowLevelILRegularFunction};
 use crate::medium_level_il::MediumLevelILFunction;
 use crate::rc::{Array, CoreArrayProvider, CoreArrayProviderInner, Guard, Ref, RefCountable};
 use crate::string::{BnString, IntoCStr};
-use std::ffi::{c_char, c_void};
+use std::ffi::c_char;
 use std::ptr;
 use std::ptr::NonNull;
+
+pub mod activity;
+pub use activity::Activity;
 
 #[repr(transparent)]
 /// The AnalysisContext struct is used to represent the current state of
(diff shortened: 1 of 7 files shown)


rust/src/workflow.rs
```diff
@@ -579,3 +580,384 @@ unsafe impl CoreArrayProviderInner for Workflow {
...
+        }
+
+        #[must_use]
+        pub fn run_once(mut self, value: bool) -> Self {
+            self.run_once = Some(value);
+            self
+        }
+
+        #[must_use]
+        pub fn run_once_per_session(mut self, value: bool) -> Self {
```
emesare: A comment referring to the file session would be ideal, as "session" is pretty generic.

Ref: https://dev-rust.binary.ninja/binaryninja/file_metadata/struct.FileMetadata.html#method.session_id

rust/src/workflow.rs
```diff
@@ -579,3 +580,384 @@ unsafe impl CoreArrayProviderInner for Workflow {
...
+        Listener,
+        Selector,
+        Sequence,
+        Subflow,
+        Task,
+    }
+
+    #[derive(Deserialize, Serialize, Debug)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Eligibility {
```
emesare: Pulling the docs out into comments here would be helpful: https://docs.binary.ninja/dev/workflows.html#activity-eligibility

rust/src/workflow.rs
```diff
@@ -579,3 +580,384 @@ unsafe impl CoreArrayProviderInner for Workflow {
...
+    impl AsConfig for Config {
+        fn as_config(&self) -> CString {
+            (&self).as_config()
+        }
+    }
+
+    #[derive(Deserialize, Serialize, Debug)]
+    #[serde(rename_all = "camelCase")]
+    #[derive(Default)]
+    pub enum Role {
```
emesare: Also here: https://docs.binary.ninja/dev/workflows.html#activity-roles_1

rust/src/workflow.rs
```diff
@@ -579,3 +580,384 @@ unsafe impl CoreArrayProviderInner for Workflow {
         )
     }
 }
+
+pub mod activity {
```
emesare: Move this out into `/src/workflow/activity.rs` perhaps? Alongside the activity itself.

Then probably add a re-export at the top of this file.

emesare:
Other than the comments I left (all of which are not blocking, pick what you want) LGTM

rust/src/workflow.rs
```diff
@@ -579,3 +584,384 @@ unsafe impl CoreArrayProviderInner for Workflow {
...
+    }
+
+    impl AsConfig for &str {
+        fn as_config(&self) -> std::ffi::CString {
+            self.to_cstr()
+        }
+    }
+
+    #[derive(Deserialize, Serialize, Debug)]
+    pub struct Config {
```
emesare: Generic name, if we expect people to import this we likely want to rename to `ActivityConfig`

bdash: They can import the `activity` module and qualify things that way if they'd like. Jamming an `Activity` prefix on the front of every type sorta defeats the point of using modules for namespacing.

rust/src/workflow.rs
bdash: Done!

bdash:
Updated to:
1. Add a lot of documentation to the new activity configuration types.
2. Rework `Workflow` so that a `WorkflowBuilder` type is used for all of its mutating methods.

bdash:
Fixed the formatting issues.

emesare (approved):

PR #7257, opened 2025-08-12
Fix disallowBranchToString in default abb.

Alkalem:
The setting should halt analysis of branch targets that fall within string references. Previous implementation looked up string references at branch source.


zznop:
Merged - good catch. This change will be available in `5.2.8140-dev`

PR #7265, opened 2025-08-14
[DSC] Add support for DriverKit shared caches

bdash:
Shared cache files use the same DriverKit platform value for both Mac and iOS. I opted to map them both to a Mac platform for now. It seems to work fine in my testing with both Mac and iOS caches. Is this likely to cause any problems?


0cyn:
lgtm! 

iOS/mac platform differentiation primarily matters for type libraries; if the DriverKit libraries are importing public frameworks and the function signatures may differ between platforms then it could cause decomp issues. Otherwise this is ok for now.

bdash:
The image paths are all prefixed with /System/DriverKit so I think neither Mac nor iOS type libraries will match at the moment.

0cyn (approved):

bdash:
This is available in 5.2.8119 and newer.

Issue #7280, opened 2025-08-15
Support Always/Never Branch patching for TBNZ/TBZ/CBNZ/CBZ on AArch64


galenbwill:
**What is the feature you'd like to have?**
Enable Always or Never Branch patching for TBNZ/TBZ/CBNZ/CBZ on AArch64

**Is your feature request related to a problem?**
On AArch64, per @ElykDeer, Invert Branch patching is supported for TBNZ/TBZ/CBNZ/CBZ instructions, but not Always or Never Branch patching.

**Are any alternative solutions acceptable?**
"Never Branch" is equivalent to "Convert to NOP", but that is not obviously equivalent to the user.

ElykDeer:
Resolved in builds >5.2.8250-dev. Thank you!

PR #7281, opened 2025-08-15
[aarch64] Enable Always and Never Branch patching for TBZ/TBNZ and CBZ/CBNZ

galenbwill:
Addresses https://github.com/Vector35/binaryninja-api/issues/7280

This PR implements the fix, but someone should sign off on this with respect to the following concern:

> while i can fix always branch (by forcing it to be a TBZ/CBZ and changing the tested register to the zero XZR/WZR so the branch always happens), its’s not clear to me that it’s “ok”, in the sense that it changes the data dependency of the instruction, from the tested register to the zero register.

### DIFF
--- a/arch/arm64/arch_arm64.cpp
+++ b/arch/arm64/arch_arm64.cpp
@@ -1168,7 +1168,7 @@ class Arm64Architecture : public Architecture
 		Instruction instr;
 		if (!Disassemble(data, addr, len, instr))
 			return false;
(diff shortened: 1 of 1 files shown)


ElykDeer:
I'm okay with the data-flow caveat; x86 will completely `nop` things out for some branches, and destroys register references in others as well. These changes look fine to me....but would there be any benefit to using a `nop` in some cases? Unless someone can think of a reason I think it's fine as-is.

PR #7299, opened 2025-08-22
[ObjC] Rewrite Obj-C workflow in Rust

### LINKED ISSUE #6779: It should be easier to add additional activities to Objective-C and shared cache workflows

bdash:
This is functionally equivalent to the previous workflow_objc, with the following changes:
1. It mutates the `core.function.metaAnalysis` workflow rather than registering a new named workflow. The activities now all check for the presence of the Objective-C metadata added by `ObjCProcessor` to determine whether they should do work, rather than relying on `MachoView` to override the function workflow when Objective-C metadata is present. This fixes https://github.com/Vector35/binaryninja-api/issues/6779.
2. The auto-inlining of `objc_msgSend` selector stub functions is performed in a separate activity from the processing of `objc_msgSend` call sites. The selector stub inlining activity is configured so that it does not run in `DSCView` as the shared cache needs different behavior for stub functions more generally that `SharedCacheWorkflow` already provides.
3. The way that types like `id` and `SEL` are referenced is fixed so that they show up as `id` rather than `objc_struct*`.

This also replaces the Objective-C portion of the shared cache's workflow, and incorporates several bug fixes that had been applied to it but not the standalone Objective-C workflow.


bdash:
Having multiple different types of analysis of `objc_msgSend` call sites be in a single activity doesn't feel great, but the process of determining finding the individual call sites is complex enough that it'd be kinda gross to have to do it in multiple activities.

Perhaps there's some way I'm missing to communicate data between different activities that would allow me to have an "identify call sites" activity that call type adjustment and direct call rewriting activities could depend on so they could reuse data that it computes.

plugins/workflow_objc/src/metadata/global_state.rs
```diff
@@ -0,0 +1,191 @@
...
+    NotLoaded,
+    Loaded(Option<SelectorImplementations>),
+}
+
+struct SelectorImplementations {
+    sel_ref_to_impl: HashMap<u64, Vec<u64>>,
+    sel_to_impl: HashMap<u64, Vec<u64>>,
+}
+
+// TODO: How do we ever clean up after these?
```
emesare: https://dev-rust.binary.ninja/binaryninja/trait.ObjectDestructor.html

bdash: Thanks! The C++ workflow had just been letting its equivalent maps grow indefinitely.

Issue #7196, opened 2025-07-28
MachoView should define imported symbols even if they are not presented in SYMTAB


mostobriv:
**Version and Platform (required):**
- Binary Ninja Version: 5.2.8012-dev Personal (aec518eb)

**Bug Description:**
Current implementation of MachoView relies on symtab when defining the symbols in the macho. However it is possible to bind imported (external) symbols even if they are not in the symtab, as binding from DYLD_INFO uses their own inlined strings. So for the cases when symbol is used in the binding, but not listed in the symtab, leads to the error with applying relocation during macho analysis.

**Steps To Reproduce:**
Please provide all steps required to reproduce the behavior:
1. Open attached binary
2. Go to `0x100004000`
3. There is null instead of external symbol

**Expected Behavior:**
I think MachoView should define symbols when they referenced in the dyld binding as well.

**Screenshots/Video Recording:**

<img width="1577" height="1056" alt="Image" src="https://github.com/user-attachments/assets/aa662333-f724-4a2c-bcbd-29d740b56293" />

<img width="759" height="361" alt="Image" src="https://github.com/user-attachments/assets/c4128e26-1506-424e-aec0-cbf1d76e82b4" />

**Binary:**
[bin.zip](https://github.com/user-attachments/files/21467470/bin.zip)

**Additional Information:**
Definition of the imported macho symbol is nested in the loop over symtab, so if there is no symbols (`symtab.nsyms == 0`) then no symbols will be defined, it's incorrect.

https://github.com/Vector35/binaryninja-api/blob/2f1d09c6b714a9f2664eab0a9dbef79e3cbd8919/view/macho/machoview.cpp#L3121

mostobriv:
Found out, that MachoView even won't apply any relocations due to failed search of symbol, so `bv.relocations_at(addr) == []`

mostobriv:
Here is a bit more interesting example. In this binary dynsymtab and symtab is empty, however there is plenty of bindings in DYLD_INFO, even weak bindings into self

[libloader.zip](https://github.com/user-attachments/files/21471003/libloader.zip)

mostobriv:
@0cyn are you planning to fix it in current milestone? I can try to fix this issue.

PR #7302, opened 2025-08-23
[MachoView] Fixed missing symbols in binding parsing

mostobriv:
This MR addressing #7196

### DIFF
--- a/view/macho/machoview.cpp
+++ b/view/macho/machoview.cpp
@@ -2146,8 +2146,15 @@ bool MachoView::InitializeHeader(MachOHeader& header, bool isMainHeader, uint64_
 
 		default:
 			if (ordinal > 0)
-			{
-				if (auto symbol = GetSymbolByRawName(name, GetExternalNameSpace()); symbol)
+			{ 
+				auto symbol = GetSymbolByRawName(name, GetExternalNameSpace());
+				if (!symbol)
+				{
+					// As macho can bind symbols, which is not listed in the symtab
+					symbol = DefineMachoSymbol(ExternalSymbol, name, 0, GlobalBinding, false);
+				}
+
+				if (symbol)
 				{
 					DefineRelocation(m_arch, relocation, symbol, relocation.address);
 					handled = true;


mostobriv:
@0cyn @bdash what do you think about bindings into Self (`BindSpecialDylibSelf`), should MachoView handle them as before - if can't find, then ignore and log an error, or should define them as external in case if can't find (this behaviour can add some extra info to the user)?

0cyn:
Hi, I apologize for the delay on this issue. The binary provided in #7196 ended up having exposing a sweeping set of issues with how we process binding information, handle stubs sections, etc. 

I've created a draft PR containing the in-progress fixes for this branch here: https://github.com/Vector35/binaryninja-api/pull/7303 

Further testing needs done on this branch to ensure it doesn't break or drastically slow down any existing Mach-O processing. 

I'm going to go ahead and close this PR in favor of that one, as it includes the fix contained in this PR. 

We are hoping to merge this very soon after we're able to validate the changes. Thank you so much for your PR!

mostobriv:
Sure, no problem

mostobriv:
However, what about this?

> what do you think about bindings into Self (`BindSpecialDylibSelf`), should MachoView handle them as before - if can't find, then ignore and log an error, or should define them as external in case if can't find (this behaviour can add some extra info to the user)?

Can you cover this as well in your MR?

0cyn:
I decided to handle this as before in the PR, logging an error if it cannot be found, as they afaik should not be bindable at runtime. 

Proving an error with the symbol information and address it attempted to bind should be sufficient enough info for the user in identifying something strange occurring, without incorrectly responding to these ordinals by trying to bind something to a symbol that is not there. 

I'm very hesitant to try and infer something the binding information disagrees with.

---

I am seeing these in this libloader binary: (was this dumped from memory out of curiosity?)

Screenshot 2025-08-23 at 10 02 00 AM

In this case the intern bind itself will fail, but the la_symbol_ptr entry already encodes the value of the target function.

mostobriv:
> as they afaik should not be bindable at runtime.

they are a weak bindings, so they should actually, i mean there is no-error in the context of linking them in runtime, however i guess logging such symbols is ok as well

> was this dumped from memory out of curiosity?

nope, this is enterprise protection (RASP) of an iOS apps library, this is how its shipped in IPAs

UPD: oh, wait, ye, this one is dumped actually, but not in form of you ment i guess, its dumped via old version of mremap_encrypted, so technically it wasnt loaded (?)

mostobriv:
I can try to create such binary statically for you, if it helps, guess it will only require bit of llvm magic

bdash:
I didn't see anything in the binary that looked like a self bind.

I think an oversight here is that bindings in the weak bindings table [cannot explicitly set a library ordinal](https://github.com/apple-oss-distributions/dyld/blob/93bd81f9d7fcf004fcebcb66ec78983882b41e71/common/MachOLayout.cpp#L1008-L1012). The code as it stands treats those weak bindings as having a library ordinal of 0 (`BIND_SPECIAL_DYLIB_SELF`), when it [should treat all weak bindings as implicitly having a library ordinal of -3 (`BIND_SPECIAL_DYLIB_WEAK_LOOKUP`)](https://github.com/apple-oss-distributions/dyld/blob/93bd81f9d7fcf004fcebcb66ec78983882b41e71/common/MachOLayout.cpp#L991).

Something like:

```diff
-               uint64_t ordinal = 0;
+               uint64_t ordinal = binding == WeakBinding ? BindSpecialDylibWeakLookup : 0;
```

PR #7296, opened 2025-08-20
[ARM64] fix ldrsw lift to sign extend in certain encodings

yrp604:
fixes #7295

Issue #2612, opened 2021-09-01
Workflow for lifter checking


CouleeApps:
Make a PluginLifterDebugWorkflow with extra lifting debug messages (tags??) that may be too slow to run on normal workflows. Would be a great resource for lifter plugin devs

Many good checks in Liftcheck: https://gitlab.com/rsnyder/binja-riscv/-/blob/master/src/liftcheck.rs
Could throw in other checks too if there are good ideas

yrp604:
Two relevant existing issues:

#1908 
#1931

nshp:
Maybe #1629 as well

yrp604:
Unknown registers #2307

yrp604:
Flag when lifting `il.{Zero,Sign}Extend(N, il.Load(N, ...))`

Or more general, trying to do any extension to any value where extension size <= existing size.

PR #7307, opened 2025-08-24
[ARM64] Only extend if src size is less than target

yrp604:
Shink expr tree sizes with equivalent semantics. Developed on top of #7296 for my convenience.

This changes 3ish instructions:
- ldr variants -- conditionally emit the zero/sign extension when needed.
- fmov variants -- conditionally emit the zero extension when needed. I don't think there is existing test coverage for the `ENC_FMOV_64VX_FLOAT2INT` case, but the comment in `decode_scratchpatd` indicates it's only used with 64 bit registers.
- ubfx/ubfiz -- there are no encodings where the second argument is a different size than the first.

This removes about 700 total extensions from an arm64 ntdll.

Also added this check to the liftcheck wishlist in #2612


rssor:
There's a possibility that this winds up influencing types of temp registers due to spills in a way that is actually useful so I need to check on that a bit once I push some other stuff before accepting

Issue #7311, opened 2025-08-26
Proper MacOS/iOS arm64 syscall calling convention


mostobriv:
**What is the feature you'd like to have?**
For some reason, there is for quite some time BN use cdecl CC for syscalls on macos/ios aarch64 platform,  even there is macos-syscall

https://github.com/Vector35/binaryninja-api/blob/eb09898c0ea423328d929dec4d1730030f99abe4/arch/arm64/arch_arm64.cpp#L2687-L2700

And, for some reason, platform `MacArm64Platform` (as well as `IOSArm64Platform`) doesn't set any systemcall CC:

https://github.com/Vector35/binaryninja-api/blob/eb09898c0ea423328d929dec4d1730030f99abe4/platform/mac/platform_mac.cpp#L134-L146

PR #7313, opened 2025-08-27
[Arm64, Apple] Apple arm64 syscall calling convention

mostobriv:
This MR addressing #7311. A bit confused why hasn't it been done earlier.

As "macos-syscall" was never registered and used by any of platform types, it is safe to rename it to "apple-syscall" (as it applied to both macos and ios).

Also changed arguments for this cc from X0-X5 to X0-X8 as syscalls on ios/macos can handle up to 9 arguments


mostobriv:
I didn't change anything in mac_kernel platform, as i'm not much familiar with osx kernel stuff, however i have a feeling that syscalls there behave the exact same way, so if i should add cc there lmk

bdash:
Thank you very much for the change!

PR #7351, opened 2025-09-05
Supports pointer base types and offsets in Rust API

lukbukkit:
This PR implements Rust API functions to set and read the pointer base type and its accompanying offset in [TypeBuilder](https://github.com/Vector35/binaryninja-api/blob/dev/rust/src/types.rs#L57) and to read the respective properties from a [Type](https://github.com/Vector35/binaryninja-api/blob/dev/rust/src/types.rs#L446).


emesare:
LGTM, not sure why the rust CI stuff did not run (unrelated).

Seeing as we still have a single monolithic type builder in the rust api this is fine. In the future we will be trying to split it up into each "kind" of type so that people don't try and set these for an integer, again, unrelated to this PR just trying to add context where I can.

PR #7356, opened 2025-09-06
[MIPS] Mark `jalr[.hb] $zero, $ra` as a return instruction

### LINKED ISSUE #7355: MIPS64R6 `jalr[.hb] $zero, $ra` should be treated as a return variant
As per [*MIPS&reg; Architecture For Programmers Volume II-A: The MIPS64&reg; Instruction Set Reference Manual*](https://archive.org/details/mips-doc), Revision 6.06, pp. 272:

> Release 6 maps JR and JR.HB to JALR and JALR.HB with rd = 0:
> Pre-Release 6, JR and JALR were distinct instructions, both with primary opcode SPECIAL, but with distinct function codes.
> Release 6: JR is defined to be JALR with the destination register specifier rd set to 0. The primary opcode and function field are the same for JR and JALR. The pre-Release 6 instruction encoding for JR is removed in Release 6.
> Release 6 assemblers should accept the JR and JR.HB mnemonics, mapping them to the Release 6 instruction encodings.

`jalr[.hb] $zero, $ra` is architecturally synonymous with `jr[.hb] $ra`, and both should be treated as return instructions. However, the current lifter fails to do so:

https://github.com/Vector35/binaryninja-api/blob/9e83ea144de09b78a696a0a9997ec28e3bc55327/arch/mips/arch_mips.cpp#L336-L339

https://github.com/Vector35/binaryninja-api/blob/9e83ea144de09b78a696a0a9997ec28e3bc55327/arch/mips/arch_mips.cpp#L383-L390

CSharperMantle:
This PR marks MIPS `jalr[.hb] $zero, $ra` as a return instruction.

Resolves #7355.

## Effects

Before:

before

After:

after

### DIFF
--- a/arch/mips/arch_mips.cpp
+++ b/arch/mips/arch_mips.cpp
@@ -332,10 +332,12 @@ protected:
 			result.AddBranch(CallDestination, instr.operands[0].immediate, nullptr, hasBranchDelay);
 			break;
 
-		//Jmp to register register value is unknown
 		case MIPS_JALR:
 		case MIPS_JALR_HB:
-			result.delaySlots = 1;
+			if (instr.operands[0].reg == REG_ZERO && instr.operands[1].reg == REG_RA)
+				result.AddBranch(FunctionReturn, 0, nullptr, hasBranchDelay);
+			else
+				result.AddBranch(UnresolvedBranch, 0, nullptr, hasBranchDelay);
 			break;
 
 		case MIPS_BGEZAL:
--- a/arch/mips/il.cpp
+++ b/arch/mips/il.cpp
@@ -1430,7 +1430,10 @@ bool GetLowLevelILForInstruction(Architecture* arch, uint64_t addr, LowLevelILFu
 		{
 			operand = 2;
 		}
-		il.AddInstruction(il.Call(ReadILOperand(il, instr, operand, registerSize(instr.operands[operand]), addrSize, true)));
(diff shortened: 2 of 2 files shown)


CSharperMantle:
Hi @psifertex and @zznop, I've rebased the branch onto the latest `dev` branch. Any progress on getting this reviewed? :)

zznop:
Yes, I'll get this tested and merged today or tomorrow.

zznop:
Looks like this code change affects jalr branches that don't use the zero operand too, and it causes regressions

image

image

Should be a simple fix. Looking into it now.

zznop:
I'm going to go with:

```
		case MIPS_JALR_HB:
			if (instr.operands[0].reg == REG_ZERO && instr.operands[1].reg == REG_RA)
				result.AddBranch(FunctionReturn, 0, nullptr, hasBranchDelay);
			else
				result.delaySlots = 1;
			break;
```

I see that you were probably following the pattern for `MIPS_JR`/`MIPS_JR_HB`, but I think we want to continue to not do anything in cases where it's not a `jalr $zero, $ra` and let Binja figure out the branch type when lifting.

zznop:
Merged your commit with the changes discussed above. This will be available in 5.3.8765-dev and later

CSharperMantle:
> Merged your commit with the changes discussed above. This will be available in 5.3.8765-dev and later

Thanks!

PR #7399, opened 2025-09-14
[DSC] Fix support for iOS 15 shared caches

### LINKED ISSUE #7393: [DSC] Problems when loading an iOS 15 shared cache
- Binary Ninja Version: 5.2.8258-dev Ultimate, 42dff25b
- OS: macos
- OS Version: 26.0
- CPU Architecture: arm64

**Bug Description:**
When loading the shared cache from `19D50__iPhone10,3_6`, several problems are visible:

1. Objective-C method names are garbage.

   Older iOS versions stored the relative selector base offset in libobjc's `__objc_opt_ro` section, rather than at a location referenced by the shared cache's header. We don't look there.

2. `dyld_shared_cache_arm64.symbols::0_0x0` is mapped into the view at address 0.

    This causes all null pointers to show up as pointers to the string `dyld_v1   arm64`. This doesn't happen with newer shared caches and isn't intended.

3. Attempting to save a database throws an exception:
    > Error while saving database snapshot: stackTop_ + sizeof(T) * count <= stackEnd_

    This appears to be related to the `.symbols` file. If I tweak `SharedCacheBuilder::AddFile` to skip it then I'm able to save the database successfully.

4. Cross-image calls show as bare addresses, rather than names.

   The stub functions reside in `__stubs` sections within library images, rather than in stub islands or other region types. We automatically load stub island regions to allow stub symbols to be resolved, but intentionally avoid loading images.

    I don't see any obvious pattern as to which images the stubs are in. `objc_msgSend` is found in SwiftUI, `__os_log_pack_fill` is in AVFCore, `___error` is in `AppSupport`.

    The context menu correctly shows which library the stub function resides in and loading it applies the correct symbol name, but has the obvious consequence of analyzing that entire image.

**Steps To Reproduce:**
1. Open 19D50__iPhone10,3_6/dyld_shared_cache_arm64
2. Load CFNetwork

bdash:
This does the following:
* Adds support for relative direct selectors in older shared cache versions.

    Prior to macOS 13 / iOS 16, the base offset to use for relative direct selector references within Objective-C message lists was stored within the `__TEXT,__objc_opt_ro` section of /usr/lib/libobjc.A.dylib.

* Reworks handling of .symbols files to be compatible with iOS 15

   In some iOS 15 caches, the .symbols file's mapping has an address of 0. This would cause it to be returned by `SharedCache::GetEntryContaining` and loaded into the view. The .symbols file contains the local symbol tables for images in the shared cache. It is not intended to be mapped into the same address space as the rest of the shared cache.
    
    `SharedCache` now tracks the symbols cache entry separately from other entries, and `SharedCacheMachOProcessor` now directly accesses the local symbols cache entry rather than needing to search for it, and uses the separate VM object for reading data from it.

Fixes #7393.

### DIFF
--- a/view/sharedcache/core/SharedCache.cpp
+++ b/view/sharedcache/core/SharedCache.cpp
@@ -91,10 +91,8 @@ CacheEntry CacheEntry::FromFile(const std::string& filePath, const std::string& 
 	{
 		// We found a single symbols cache entry file. Mark it as such!
 		type = CacheEntryType::Symbols;
-		// Adjust the mapping for the symbol file, they seem to be only for the header.
-		// If we do not adjust the mapping than we will not be able to read the symbol table through the virtual memory.
-		mappings[0].fileOffset = 0;
-		mappings[0].size = file->Length();
+		// Symbol files are not mapped into the address space.
+		mappings.clear();
 	}
 	else if (mappings.size() == 1 && header.imagesCountOld == 0 && header.imagesCount == 0
 		&& header.imagesTextOffset == 0)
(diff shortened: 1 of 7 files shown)


view/sharedcache/core/SharedCache.cpp
```diff
@@ -231,6 +229,17 @@ void SharedCache::AddEntry(CacheEntry entry)
 	// Get the file accessor to associate with the virtual memory region.
 	auto fileAccessor = FileAccessorCache::Global().Open(entry.GetFilePath());
 
+	if (entry.GetType() == CacheEntryType::Symbols)
+	{
+		m_localSymbolsEntry = std::move(entry);
+		// Map the entire file into its own virtual memory space.
+		// This is necessary due to code that processes symbols being written in terms of a `VirtualMemory`
```
emesare, line 236: You can use the `fileAccessor` in place of the `m_localSymbolsVM`, since it is just that file mapped at zero and it does not exist inside the regular virtual memory.

Just a suggestion, nothing blocking

bdash: As the comment mentions, this uses a `VirtualMemory` as `SharedCacheMachOHeader::ReadSymbolTable` requires one. Since it is also used to process symbol tables inside images, such as in macOS shared caches, it cannot be updated to work with a `FileAccessor`. Some additional refactoring would be required to remove that use of `VirtualMemory`.

PR #7417, opened 2025-09-20
[AArch64] Update disassembler based on 2025-06 ARM ISA data

### LINKED ISSUE #7402: Unhandled iPhone Air 26.0 kernel instructions
- Binary Ninja Version: 5.2.8218-dev, 60b14381
- OS: macos
- OS Version: 15.7
- CPU Architecture: arm64

**Bug Description:**
the new iPhone Air/17 introduced some new features which depend on new ARM instructions

**Steps To Reproduce:**
Load in the kernelcache.release.iPhone18,4 from the 26.0 IPSW, browse to 0xFFFFFE000842D1EC (which should be `arm_vm_init`). Notice the missing instructions following the mcr operations:

```
fffffe000842d460  481018d5   msr     cpacr_el1, x8
fffffe000842d464  df3f03d5   isb     
fffffe000842d468  e8018052   mov     w8, #0xf
fffffe000842d46c  0800a872   movk    w8, #0x4000, lsl #0x10  {0x4000000f}
fffffe000842d470  c81218d5   msr     smcr_el1, x8
🚫fffffe000842d474  2858bf04   ??
```

**Expected Behavior:**
Full perfect disassembly

**Screenshots:**

**Binary:**
https://ipsw.me/install/iPhone18,4/23A341

**Additional Information:**

bdash:
Alongside the update I also added support for decoding:

* LDR / STR (table)
* PMULLB / PMULLT
* ABS / CNT / CTZ
* SMIN / SMAX / UMIN / UMAX
* RPRFM
* PSEL

Note that while these instructions will now be disassembled, they are not yet lifted to LLIL.

Additionally, I fixed a number of errors in the decoding of some less commonly occurring instructions.

Fixes #7402.


plafosse:
Just marking down my observations:
Removing sysreg.h - I assume this is due to how we're handling these as enums now, and this is no longer needed?
You fixed FPOne and FPTwo - Surprised that code wasn't throwing a warning....
New register `REG_ZT0`?
Looks like `sysregs_gen.h` has not been updated was this intentional?
Looks like there are ~500 more instruction in `operations.h` were we really missing that many?

bdash:
The large number of new operations is due to many features being picked up for the first time. And many of the new instructions have multiple variants.
* `FEAT_LSFE`, `FEAT_LSE128`, and `FEAT_LSUI ` add new atomic operations with multiple variants for the different combinations of acquire / release semantics. These account for between half and two thirds of the new instructions.
* `FEAT_MOPS` adds 40 different memory set / memory copy instructions that each have 3 variants.
* `FEAT_CMPBR` adds 36 different comparison instructions (6 base instructions with 6 different condition codes).
* `FEAT_THE` adds 12 instructions related to TLB handling that each have 4 variants.
* `FEAT_CSSC` adds 22 new instructions.

The code related to sysregs was changed to be generated in b095215397ca8fe7e9808f312403ec4494de76f6 but the old version of the code was not removed.

`ZT0` is added as part of `FEAT_SME2`.

bdash:
In the latest change I restored the handling of `UDF_perm_undef` instructions. They pseudocode in the latest ARM data had gained an explicit `EndOfDecode(Decode_UNDEF)` that was causing them to be decoded to `error` rather than `udf #imm` like they were previously.

PR #6721, opened 2025-04-24
implement rust custom data renderer

rbran:

PR #7438, opened 2025-09-28
Rust API: Add Data Renderer API

### LINKED ISSUE #6485: Add data renderer API to Rust
Pretty straightforward now that we have better text token API in Rust. For some more context also see the code referenced in this ticket: https://github.com/Vector35/binaryninja-api/issues/6484

Relevant data renderer C++ API:
https://github.com/Vector35/binaryninja-api/blob/1675348020724286554260961199460841497619/datarenderer.cpp

lukbukkit:
This pull request adds an implementation for the data renderer API currently available for Python and C++ plugins. My work is based on #6721. This PR fixes #6485. I've added an example plugin which renders UUIDs of the Mach-O load command `LD_UUID`. 

The tests for this feature are currently in progress. What kind of tests would you expect for this feature, should they be similar to `render_layer.rs`?


emesare:
For tests it would be similar to the render layer one, however seeing as you may be writing them blind (not sure if you have a headless supporting license) whenever this PR is ready for review I can write the tests for the API.

For rust we typically have the `examples` directory be actual binary examples, I think we may want a seperate directory for plugin examples to limit confusion (`examples` directory is typically reserved for `cargo run --examples`). If you would like you can make a new directory called `plugin_examples` I think that would be an improvement, obviously if you feel differently then we can have a discussion here.

lukbukkit:
> For tests it would be similar to the render layer one, however seeing as you may be writing them blind (not sure if you have a headless supporting license) whenever this PR is ready for review I can write the tests for the API.

That would be awesome because my personal license does not support headless operation, and I'm not too familiar with writing tests for Binary Ninja.

> For rust we typically have the examples directory be actual binary examples, I think we may want a seperate directory for plugin examples to limit confusion (examples directory is typically reserved for cargo run --examples). If you would like you can make a new directory called plugin_examples I think that would be an improvement, obviously if you feel differently then we can have a discussion here.

Yes, that sounds good!

emesare:
Thank you for helping push this through, I made some changes to the API that I think better reflect how its intended to be used and some other comments that I think help clarify things that are not as obvious.

https://github.com/Vector35/binaryninja-api/commit/e3a3ac2e055117495981172fd75d33d5a65360ac

Please let me know if this looks acceptable and i will squash all the changes up and merge (with you and @rbran as author)

lukbukkit:
Thanks for improving the API even further. The final result looks good to me, and it also works for my custom data renderer plugin.

emesare:
Added with https://github.com/Vector35/binaryninja-api/commit/64633f61f7b9e03be9437b5f4896bbe122f7a7a2 thank you!

PR #7440, opened 2025-09-30
[ObjC] Add support for removing runtime calls that perform retain count operations

bdash:
A new activity is added that will remove calls to `objc_retain`, `objc_release`, `objc_autorelease`, and related functions. It is disabled by default due to the fact it changes the semantics of the code. It can be enabled on a per-function basis via the Function Settings context menu or command palette entry, or it can be enabled in Open with Options or in user settings if a user would prefer it be on for an entire file or for all files they open.

For now the activity is only eligible within arm64 binaries. Supporting x86_64 will require matching some slightly different IL patterns.

### DIFF
--- a/plugins/workflow_objc/src/activities/remove_memory_management.rs
+++ b/plugins/workflow_objc/src/activities/remove_memory_management.rs
@@ -1,0 +1,192 @@
+use binaryninja::{
+    architecture::{Architecture as _, CoreRegister, Register as _, RegisterInfo as _},
+    binary_view::{BinaryView, BinaryViewExt as _},
+    low_level_il::{
+        expression::{ExpressionHandler, LowLevelILExpressionKind},
+        function::{LowLevelILFunction, Mutable, NonSSA},
+        instruction::{
+            InstructionHandler, LowLevelILInstruction, LowLevelILInstructionKind,
+            LowLevelInstructionIndex,
+        },
+        lifting::LowLevelILLabel,
+        LowLevelILRegisterKind,
+    },
+    workflow::AnalysisContext,
+};
+
+use crate::{error::ILLevel, metadata::GlobalState, Error};
+
+// TODO: We should also handle `objc_retain_x` / `objc_release_x` variants
+// that use a custom calling convention.
+const IGNORABLE_MEMORY_MANAGEMENT_FUNCTIONS: &[&[u8]] = &[
+    b"_objc_autorelease",
+    b"_objc_autoreleaseReturnValue",
+    b"_objc_release",
+    b"_objc_retain",
+    b"_objc_retainAutorelease",
+    b"_objc_retainAutoreleaseReturnValue",
+    b"_objc_retainAutoreleasedReturnValue",
+    b"_objc_retainBlock",
+    b"_objc_unsafeClaimAutoreleasedReturnValue",
+];
+
+fn is_call_to_ignorable_memory_management_function<'func>(
+    view: &binaryninja::binary_view::BinaryView,
+    instr: &'func LowLevelILInstruction<'func, Mutable, NonSSA>,
+) -> bool {
+    let target = match instr.kind() {
+        LowLevelILInstructionKind::Call(call) | LowLevelILInstructionKind::TailCall(call) => {
+            let LowLevelILExpressionKind::ConstPtr(address) = call.target().kind() else {
+                return false;
+            };
+            address.value()
+        }
+        LowLevelILInstructionKind::Goto(target) => target.address(),
+        _ => return false,
+    };
+    let Some(symbol) = view.symbol_by_address(target) else {
+        return false;
+    };
+
+    let symbol_name = symbol.full_name();
+    let symbol_name = symbol_name.to_bytes();
+
+    // Remove any j_ prefix that the shared cache workflow adds to stub functions.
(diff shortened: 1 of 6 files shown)


plugins/workflow_objc/src/activities/remove_memory_management.rs
```diff
@@ -0,0 +1,196 @@
...
+            );
+        },
+        LowLevelILInstructionKind::Goto(_) => {
+            // The shared cache workflow inlines calls to stub functions, which causes them
+            // to show up as a `lr = <next instruction>; goto <stub function instruction>;`
+            // sequence. We need to remove the load of `lr`  and update the `goto` to jump
+            // to the next instruction.
+
+            let Some(prev) =
+                llil.instruction_from_index(LowLevelInstructionIndex(insn.index.0 - 1_usize))
```
VisualEhrmanntraut:
```suggestion
                llil.instruction_from_index(LowLevelInstructionIndex(insn.index.0 - 1))
```

isn't it more readable that way?

plugins/workflow_objc/src/activities/remove_memory_management.rs
```diff
@@ -0,0 +1,196 @@
...
+    view: &binaryninja::binary_view::BinaryView,
+    instr: &'func LowLevelILInstruction<'func, Mutable, NonSSA>,
+) -> bool {
+    let target = match instr.kind() {
+        LowLevelILInstructionKind::Call(call) | LowLevelILInstructionKind::TailCall(call) => {
+            if let LowLevelILExpressionKind::ConstPtr(address) = call.target().kind() {
+                address.value()
+            } else {
+                return false;
+            }
```
VisualEhrmanntraut:
```suggestion
            let LowLevelILExpressionKind::ConstPtr(address) = call.target().kind() else {
                return false;
            };
            address.value()
```

plugins/workflow_objc/src/activities/remove_memory_management.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    let symbol_name = symbol.full_name();
+    let mut symbol_name = symbol_name.to_bytes();
+
+    // Remove any j_ prefix that the shared cache workflow adds to stub functions.
+    if symbol_name.starts_with_str("j_") {
+        symbol_name = &symbol_name[2..];
+    }
+
+    IGNORABLE_MEMORY_MANAGEMENT_FUNCTIONS.contains(&symbol_name)
```
VisualEhrmanntraut:
```suggestion
    let symbol_name = symbol_name.to_bytes();

    // Remove any j_ prefix that the shared cache workflow adds to stub functions.
    IGNORABLE_MEMORY_MANAGEMENT_FUNCTIONS.contains(symbol_name.strip_prefix(b"j_").unwrap_or(symbol_name))
```

plugins/workflow_objc/src/activities/remove_memory_management.rs
```diff
@@ -0,0 +1,196 @@
...
+            func_start: func.start(),
+        });
+    };
+
+    let mut function_changed = false;
+    for block in llil.basic_blocks().iter() {
+        for insn in block.iter() {
+            match process_instruction(&view, &llil, &insn, link_register, link_register_size) {
+                Ok(true) => function_changed = true,
+                Ok(_) => {}
```
VisualEhrmanntraut, line 174:
```suggestion
                Ok(v) => function_changed |= v,
```

VisualEhrmanntraut:
sorry, didn't mean to start separate reviews, just provide suggestions 😅

plugins/workflow_objc/src/activities/remove_memory_management.rs
bdash: Good point, this is a big readability improvement.

plugins/workflow_objc/src/activities/remove_memory_management.rs
bdash: Shorter isn't always better.

VisualEhrmanntraut: I made a sample to see which is better

image

This is before the proposed change:
[SampleBefore.txt](https://github.com/user-attachments/files/22613549/SampleBefore.txt)

And this is after, much shorter generated assembly:
[SampleAfter.txt](https://github.com/user-attachments/files/22613550/SampleAfter.txt)

edit: in other words, it allows the compiler to optimise better

bdash: Faster is not what I meant by better. Simpler is.

Optimization is irrelevant to this code fragment. Updating a flag to indicate that we've modified the function is going to be utterly dwarfed by the actual work of iterating, matching, and replacing instructions.

That said, it's important to read the assembly rather than judging it based on the number of lines. In the "before" case the compiler has unrolled the loop so it can use SSE to test multiple values at once. In the "after" case it is processing a single item at a time. Mostly that just shows that microbenchmarks are bad unless you're careful.

VisualEhrmanntraut: less code means the CPU is more likely to keep it cached, and more code in branches would make the code run slower when the branch prediction fails

bdash: None of that is relevant here. Please stop.

Issue #7247, opened 2025-08-08
[SharedCache] The SharedCache plugin is not identifying all images in the DYLD Shared Cache


WeiN76LQh:
**Version and Platform (required):**
- Binary Ninja Version: 5.2.8080-dev (dc62d33d)
- Edition: Commercial
- OS: macOS
- OS Version: 15.5
- CPU Architecture: M1

**Bug Description:**
For some reason the SharedCache plugin is not identifying all images within a copy of the DYLD Shared Cache. An example of this is `/System/Library/PrivateFrameworks/ChronoUIServices.framework/ChronoUIServices` in the DYLD Shared Cache for iOS 18.1.1 for an iPhone 16 Pro Max (iPhone17,2_18.1.1_22B91 downloadable [here](https://ipsw.me/download/iPhone17,2/22B91)). The load address for that image in that specific instance of the DYLD Shared Cache is `0x1bb7c3000`.

There are a total of 3776 images within the DSC for that ipsw but the SharedCache plugin for Binary Ninja is only able to identify 3711.

**Additional Information:**
Using the `ipsw` tool you can dump information about an instance of DYLD Shared Cache. Running the following command on the copy of the DSC found in the ipsw mentioned above, results in the following output:

```bash
> ipsw dyld info dyld_shared_cache_arm64e -d | head -n10
Header
======
Magic          = "dyld_v1  arm64e"
Platform       = iOS
OS Version     = 18.1
Max Slide      = 0x20000000 (ASLR entropy: 15-bits, 512MB)
Num Images     = 3776
Num SubCaches  = 62

Shared Region:                 4GB, address: 0x180000000 -> 0x280CB0000
```

bdash:
I'm not quite sure what's going on here, but I am not able to reproduce this. I've tried both the version this was reported against (5.2.8080-dev) and current dev (5.2.8211-dev) get the same result in both:
```
>>> bv.file
<FileMetadata: /Users/mrowe/22B91__iPhone17,2/dyld_shared_cache_arm64e>
>>> shared_cache
<SharedCacheController: 3776 images, 22087 regions>
>>> shared_cache.get_image_with_name('/System/Library/PrivateFrameworks/ChronoUIServices.framework/ChronoUIServices')
<CacheImage '/System/Library/PrivateFrameworks/ChronoUIServices.framework/ChronoUIServices': 0x1bb7c3000>
```

Are you still able to reproduce this?

WeiN76LQh:
[...]

PR #7477, opened 2025-10-10
[SharedCache] Fix issue #7247

WeiN76LQh:
See https://github.com/Vector35/binaryninja-api/issues/7247#issuecomment-3392192797 for details.


WeiN76LQh:
I've tested this and it does fix it, however I'm not sure if there are other implications to the changes. Maybe there is a reason the code wasn't written this way in the first place that causes issues further on down the line? Or maybe it was just a simple mistake.

emesare:
I think this is OK and the correct thing to do, I cannot recall any reason as to why you would want to add the on-disk directory when opening a project file.

Issue #7480, opened 2025-10-13
Support loading DWARF information with relocations


psifertex:
Customer reported the following crash:

```
curl -o kernel-dbg.txz https://download.freebsd.org/releases/amd64/14.3-RELEASE/kernel-dbg.txz
curl -o kernel.txz https://download.freebsd.org/releases/amd64/14.3-RELEASE/kernel.txz
tar -xJf kernel.txz -O boot/kernel/plasma_saver.ko >plasma_saver.ko
tar -xJf kernel-dbg.txz -O usr/lib/debug/boot/kernel/plasma_saver.ko.debug >plasma_saver.ko.debug

RUST_BACKTRACE=1 python -c 'import binaryninja as bn; bv = bn.load("plasma_saver.ko"); [print(f) for f in bv.functions if not f.is_thunk]'
[10024803453866859793:10490689671908210455 BinaryView.ElfView  warn] The program header size reported by e_phentsize (0x0) is different from the size of Elf64_Phdr (0x38). The parsing proceeds with the size of Elf64_Phdr.
[10024803453866859793:10490689671908210455 BinaryView.ElfView  warn] The program header size reported by e_phentsize (0x0) is different from the size of Elf64_Phdr (0x38). The parsing proceeds with the size of Elf64_Phdr.
[110084809857316975:3098629949048523757 BinaryView.ElfView  warn] The program header size reported by e_phentsize (0x0) is different from the size of Elf64_Phdr (0x38). The parsing proceeds with the size of Elf64_Phdr.
[110084809857316975:3098629949048523757 BinaryView.ElfView  warn] The program header size reported by e_phentsize (0x0) is different from the size of Elf64_Phdr (0x38). The parsing proceeds with the size of Elf64_Phdr.
[110084809857316975:3098629949048523757 Analysis error] Attempting to add function not backed by file: 0x400180
[110084809857316975:3098629949048523757 Analysis error] Attempting to add function not backed by file: 0x4001b0
[110084809857316975:3098629949048523757 Analysis error] Attempting to add function not backed by file: 0x400470
[110084809857316975:3098629949048523757 Analysis error] Attempting to add function not backed by file: 0x400480
[110084809857316975:3098629949048523757 Analysis error] Attempting to add function not backed by file: 0x400000
[110084809857316975:3098629949048523757 Analysis error] Attempting to add function not backed by file: 0x400080
[...]

PR #7487, opened 2025-10-16
Apply relocations manually in DWARF parsing.

fuzyll:
After some DMs with @negasora, this should hopefully address #7480.

The problem is that we're not applying relocations. So, the data we're giving to `gimli` is incorrect. We need to apply those relocations.

Unfortunately, since we're in a raw view, that's not trivial to do. An earlier version of this code tried to apply the relocations ourselves, but it turns out we can't even get the correct relocation ranges out of the BV at this point.

We had the idea of taking on `object` as a dependency to tell us where the relocations are and  help us apply them manually before we hand stuff off to `gimli`. This feels bad, though, so I tried to walk Claude Code through the problem in the hopes that it would help us solve it without needing `object`.

After about 30 minutes, it decided to give up and take the dependency itself despite my having told it not to. So, I've thrown in the towel, and here's a PR with a cleaned-up version of the solution.

Feel free to reject and do something else if you can, but this code _does_ work for the crashing sample. I have it going through CI right now, so we'll see if it survives other platforms and unit tests, too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -747,6 +747,7 @@ dependencies = [
  "binaryninja",
  "binaryninjacore-sys",
  "gimli",
+ "object 0.36.7",
  "thiserror 2.0.12",
  "zstd",
 ]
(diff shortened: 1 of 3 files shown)


negasora:
Closing since https://github.com/Vector35/binaryninja-api/commit/b6740d03abd22441e2e79e1d6a13c474e3fcba90 got merged

Issue #7510, opened 2025-10-21
Incorrect GPR Encoding for PPCVLE Architecture


chedahub:
**Version and Platform (required):**
- Binary Ninja Version: 5.2.8428-dev Ultimate, 8ddb6332
- OS: windows
- OS Version: 11
- CPU Architecture: x86_64

**Bug Description:**
The current implementation of the ppcvle architecture incorrectly decodes 4-bit General Purpose Register (GPR) fields in Variable Length Encoding (VLE) instructions. According to the PowerPC VLE specification, a 4-bit field is used to specify a GPR within the ranges R0-R7 or R24-R31. The issue arises when the Most Significant Bit (MSB) of this 4-bit field is set to 1; the disassembler fails to add the required offset of 24 to the register index, leading to incorrect disassembly and analysis.

**Steps To Reproduce:**
1. Load a PPCVLE binary into Binary Ninja.
2. Navigate to a function that utilizes a 2-byte VLE instruction referencing a GPR in the R24-R31 range.
3. Observe the disassembled code.
4. Note that the GPR operand is incorrectly identified as a register in the R8-R15 range instead of the correct R24-R31 range.

**Expected Behavior:**
As documented in the PowerISA v2.03 specification (page 644), when decoding a 4-bit GPR field in a VLE instruction:
If the MSB is 0, the register should be interpreted as R0-R7.
If the MSB is 1, an offset of 24 should be added to the value, correctly mapping it to the R24-R31 range.
For example, a 4-bit field of 1000 should be decoded as GPR 24 (i.e., r24), not GPR 8.

**Screenshots:**
<img width="1228" height="364" alt="Image" src="https://github.com/user-attachments/assets/bb006320-d71c-43cd-a6cd-f4618430d49f" />

**Additional Information:**
The official specification confirming this behavior can be found in the PowerISA_V2.03_Final_Public.pdf on page 644, which states that the 4-bit GPR field is a "Field used to specify a General Purpose Register in ranges R0:R7 or R24:R31 to be used as a source or as a destination for load/store data." This confirms the intended split-range encoding.

plafosse:
Fixed in >= 5.2.8448

PR #7515, opened 2025-10-22
Fix: Incorrect GPR Encoding for PPCVLE Architecture (issue #7510)

chedahub:
### Description
This PR fixes an issue in the PowerPC VLE instruction decoder where 4-bit General Purpose Register (GPR) fields were not being decoded correctly.

### Motivation and Context
According to the PowerISA v2.03 specification (page 644), 4-bit GPR fields in VLE instructions utilize a split-range encoding. This field designates registers in the range R0-R7 or R24-R31.

The previous implementation only masked the 4-bit value, correctly handling the R0-R7 range but failing to map values to the upper R24-R31 range. This occurred when the Most Significant Bit (MSB) of the 4-bit field was set, leading to incorrect disassembly and flawed control-flow analysis. For example, a register field of 1000b (8) was incorrectly decoded as r8 instead of the correct r24.

### Changes
The Get16Rx, Get16Ry, and Get16Rz (if applicable) helper functions have been updated to align with the specification. The new logic checks the MSB of the 4-bit register field (& 0x8). If the bit is set, an offset of 16 (0x10) is added to the value. This correctly maps the register indices 8-15 to the GPRs R24-R31.

### Visual Comparison

To clearly demonstrate the impact of this fix, here is a side-by-side comparison of the disassembly before and after the changes.

- Before (Incorrect Disassembly)
As shown below, the instruction `se_lbz r6, 2(r31)` was incorrectly disassembled as `se_lbz r6, 2(r15)`.
image

-  After (Correct Disassembly)
With the updated logic, the 4-bit register field is now correctly interpreted.
image

### DIFF
--- a/arch/powerpc/decode/vle16.c
+++ b/arch/powerpc/decode/vle16.c
@@ -335,17 +335,20 @@ static InstructionId Decode16Vle(uint16_t word16, uint32_t decodeFlags)
 }
 static uint16_t Get16Rx(uint16_t word16)
 {
-	return word16 & 0xf;
+	uint16_t rx = word16 & 0xf;
+	return rx + ((rx & 0x8) >> 3) * 16;
 }
 
 static uint16_t Get16Ry(uint16_t word16)
 {
-	return (word16 >> 4) & 0xf;
+	uint16_t ry = (word16 >> 4) & 0xf;
+	return ry + ((ry & 0x8) >> 3) * 16;
 }
 
 static uint16_t Get16Rz(uint16_t word16)
 {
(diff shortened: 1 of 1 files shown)


d0now:
This PR resolves https://github.com/Vector35/binaryninja-api/issues/7510

plafosse:
Thanks for the PR!

plafosse:
I've verified the fix as correct and will be pushing the fix shortly.

plafosse:
Merged in >= 5.2.8448 Thanks again for the commit.

PR #7517, opened 2025-10-22
[PPC-VLE] Fix(ppc/il): Treat r0 as a register in e_addi/e_add16i IL

chedahub:
### Description
This PR updates the Intermediate Language (IL) generation for the e_addi and e_add16i instructions on the PowerPC VLE architecture to ensure correct data-flow analysis.

### Motivation and Context
For `e_addi` and `e_add16i` instructions, the IL was incorrectly interpreting the `r0` register as a constant `0` instead of a register source. This was caused by the `OTI_GPR0_ZERO` flag.

### Changes Made
The `OTI_GPR0_ZERO` flag has been removed.
As a result, `r0` is now correctly treated as a general-purpose register in the IL, enabling accurate data-flow analysis.
- Before: `operToIL(il, oper1, OTI_GPR0_ZERO, ...)`
- After: `operToIL(il, oper1)`

This ensures that r0 is consistently treated as a register operand in the IL for these instruction.

### Visual Comparison
- Before
image
image

- After
image

### DIFF
--- a/arch/powerpc/il.cpp
+++ b/arch/powerpc/il.cpp
@@ -611,7 +611,7 @@ bool GetLowLevelILForPPCInstruction(Architecture *arch, LowLevelILFunction &il,
 				ei0 = il.Const(addressSize_l, oper2->simm);
 			ei0 = il.Add(
 				addressSize_l,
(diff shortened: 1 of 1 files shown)


plafosse:
Thanks for the PR! This should be fixed in 5.2.8512

PR #7538, opened 2025-10-25
[CMake] Ignore compiler warnings from API headers

ekilmer:
Before this change, when external plugin projects use add_subdirectory on the API repository, any warnings they enable in their CMake project will affect the API headers and potentially show warnings that shouldn't matter to the plugin source code they're trying to develop.

Now, we use CMake's SYSTEM option for API include directories when the API is not top-level, which silences compiler warnings originating from the API header files.

Full disclosure, this implementation is copied from this (very helpful) project https://github.com/friendlyanon/cmake-init-shared-static in [`cmake/variables.cmake`](https://github.com/friendlyanon/cmake-init-shared-static/blob/16c35490df8326a10f3e98657081f73460fc3739/cmake/variables.cmake#L24-L41), which is generated from https://github.com/friendlyanon/cmake-init

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,30 @@ if(NOT CMAKE_SIZEOF_VOID_P EQUAL 8)
     endif()
 endif()
 
+# PROJECT_IS_TOP_LEVEL is set by project() in CMake 3.21+
+if(CMAKE_VERSION VERSION_LESS "3.21")
+    string(COMPARE EQUAL "${CMAKE_SOURCE_DIR}" "${PROJECT_SOURCE_DIR}" PROJECT_IS_TOP_LEVEL)
+endif()
+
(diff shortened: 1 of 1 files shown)


ekilmer:
Closing this because I just learned about the `SYSTEM` option in CMake 3.25 when using `add_subdirectory` https://cmake.org/cmake/help/latest/command/add_subdirectory.html

So, I can get the same result as this PR by doing the following with CMake 3.25+

```cmake
add_subdirectory(binaryninja-api SYSTEM)
```

**EDIT:** Plus, this PR missed making the vendored `fmt` headers included with `-isystem` using their option `FMT_SYSTEM_HEADERS`, i.e.,

```cmake
set(FMT_SYSTEM_HEADERS ON)
add_subdirectory(vendor/fmt EXCLUDE_FROM_ALL)
```

PR #7547, opened 2025-10-29
[CMake] Allow flexible user plugin install location

ekilmer:
Create a cache variable of the same name as the previous normal variable that users can set to a specific path for user plugin installation destination.

This is primarily to help with CI builds to make it easier to find where the final plugin is located before uploading as a release.

**NOTE:** I wasn't sure the best way to handle the duplicate variables, so I just modified both locations in the same way. The effect of setting the `BN_USER_PLUGINS_DIR` cache variable multiple times shouldn't be an issue, but I'm happy to change it to something better if you have some suggestions. In fact, with `BN_USER_PLUGINS_DIR` as a cache variable, we could remove the duplication in the `bn_install_plugin` function.


CouleeApps:
Simple and works. LGTM.

PR #7557, opened 2025-11-02
Add functions_by_name to rust implementation of BinaryViewExt

james-a-johnson:
Python has the get_functions_by_name method on binary views that is useful for finding a function by name. This adds that same method to the rust implementation for binary views.

The implementation is essentially the same as the python version apart from the ordering of the symbols. Python will order functions defined in the binary before functions defined outside of the binary. The rust implementation of `symbols_by_name` does not take an ordering for the symbols so that is currently left out here.


emesare:
LGTM

PR #7560, opened 2025-11-03
[PPCVLE] Fix VLE ANDIx translation to always store result in destination register

chedahub:
### Issue
VLE instructions translated to ANDIx (e.g., `e_andi`, `e_and2i`) did not store 
results in the destination register when the rc flag was not set.

### Fix
Changed `SetRegister` to always store the result, with conditional flag writing 
based on `instruction->flags.rc`.

### Impact
VLE ANDIx-family instructions now correctly store computation results regardless 
of rc flag state.


plafosse (approved):

plafosse:
Thanks for your contribution! Fixed in > 5.2.8545

PR #7565, opened 2025-11-04
Enums from macro definitions for ios sdk

mostobriv:
This is the enums for an iOS generated from macro definitions from `iOS18.6.sdk`, collected automatically via clang python api + names for them generated via lowtier LLM (as i has no access to the frontier LLMs atm) and partially analysed and brushed manually (only collisions within a single file resolved + few splits/merges done by hand). 
I don't know if i will complete this, as i've brushed ~50 files and bored as hell of this routine, however i hope it may help you.

current todos:
- regenerate enum names with something better than qwen-30b
- comment/remove conditional macros as they are just flooding the namespace
- brush the most of the files...*

\* as the macros gathered automatically, main heuristic for their relation is continuity of the definitions + i've added some dumb heuristics for the common prefixes and ascending values, however macros sometimes isn't defined continuously but blasted within the file, so to increase the quality of the generated types it should be verified by hand 

IMPORTANT NOTICE: some of the macro values may be  incorrect because of multiple reasons:
- some of the headers is private, intended to be used via chained including from other headers and it may refer to the "parent" namespace, but i crawled them independently, (for example `netinet6/in6.h` is considered to be used as part of `netinet/in.h`, so `IPV6PROTO_MAXID` is defined referencing `IPPROTO_PIM` macro from `netinet/in.h`
- some of macros body contains function-like macros, but without preprocessing we can't use em, and as my crawler is pretty dumb this may lead to the identifiers simply being removed, so macro would be instantiated incorrectly, for example `err_sub_hypervisor` from `arm64/hv/hv_kern_types.h` it defined as `err_sub(0xba5)`, in my current implementation of crawler when trying to evaluate this expression `err_sub` is removed, thus only `(0xba5)` is left, which is incorrect


mostobriv:
What's hand-checked (names still may be needed to change):
- `include/AppleArchive/*`
- `include/malloc/*`
- `include/arm/*`

mostobriv:
i've decided to close this PR, as generated enums is pretty low-quality, so i'd better rewrite my crawler and throw bunch of checks to increase quality

PR #7566, opened 2025-11-04
Fix UAF when passing TypeBuilder objects into some APIs

plafosse:
This is a fundamental rewrite of how TypeBuilder objects are handled in the python API.

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -5432,8 +5432,8 @@ class BinaryView:
 
 		if isinstance(var_type, str):
 			(var_type, _) = self.parse_type_string(var_type)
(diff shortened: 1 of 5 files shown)


CouleeApps:
Good changes, a bunch of UAFs fixed here. This exposes a bunch of places where we do this pattern:

	if isinstance(var_type, str):
		(var_type, _) = self.parse_type_string(var_type)
	var_type = var_type.immutable_copy()

... to handle var_type being one of Type, TypeBuilder, or str. We may want to factor out this common pattern in a future change, since it is so common, and so that we don't miss any future calls to `immutable_copy`.

PR #7611, opened 2025-11-10
Fix crash for psuedo views (C/Obj-C/Rust) with unknown array access e…

### LINKED ISSUE #7603: Crashing for specific function in Psuedo-C/Rust
- Binary Ninja Version: 5.1.8104 Personal, 6eaaf348
- OS: windows
- OS Version: 11
- CPU Architecture: x86_64

**Bug Description:**
When performing a search in Binary Ninja ("Find type: Text (Pseudo C)") in a specific binary, Binary Ninja crashes.

**Steps To Reproduce:**
- Open fvevol.sys in Windows 11 25H2 (26200).
- Open find window, select "Text (Pseudo C)" as Find type
- Search for " + 0x334" 
- Crash at roughly 43% of the search

**Expected Behavior:**
Finishing the search without a crash.

**Screenshots:**
-

**Binary:**
https://msdl.microsoft.com/download/symbols/fvevol.sys/37057C75eb000/fvevol.sys

**Additional Information:**

emesare:
…xpr type

Fixes https://github.com/Vector35/binaryninja-api/issues/7603

### DIFF
--- a/lang/c/pseudoc.cpp
+++ b/lang/c/pseudoc.cpp
@@ -592,16 +592,27 @@ void PseudoCFunction::AppendFieldTextTokens(const HighLevelILInstruction& instr,
 		if ((!settings || settings->IsOptionSet(ShowTypeCasts)) && srcExpr.operation == HLIL_ARRAY_INDEX)
 		{
 			auto arrayIndexExpr = srcExpr.GetSourceExpr<HLIL_ARRAY_INDEX>();
-			if (arrayIndexExpr.operation == HLIL_VAR
-				&& arrayIndexExpr.GetType()->GetChildType()->GetWidth() < instr.size)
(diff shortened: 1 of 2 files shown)


D0ntPanic:
The other paths that use the variable type also check to make sure the variable type isn't null. Probably should check that as well to guard against missing variable info.

D0ntPanic:
Definitely need to check the variable type for null, as the following code is present on the code to fetch the variable type from the function:

```c++
	//Too noisy, disabling until we believe we've solved the obvious cases.
	//LogWarn("Unable to retrieve variable type for variable with id 0x%" PRIx64 " in function 0x%" PRIx64 ". Please report this issue.", var.ToIdentifier(), m_start);
	return nullptr;
```

The core code is not confident that variable types will not be null.

D0ntPanic (approved):

PR #7614, opened 2025-11-11
handle None cases when empty strings are returned from the core

psifertex:
Prior to this commit (and the broken "fix" before it in [ac909374541f32697a065c43e7e12c5d6916f937](https://github.com/Vector35/binaryninja-api/commit/ac909374541f32697a065c43e7e12c5d6916f937#diff-6782036b11fde2c417fa4d886777898de32ff7edff2d3d0e6171614516127f3aR167-R168)), the following API failed:

`CoreVersionInfo("4.2.0")` 

This resolves the issue by checking if a string is empty before trying to decode it and failing.

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -164,7 +164,7 @@ class CoreVersionInfo:
 			self.major = core_version_info.major
 			self.minor = core_version_info.minor
 			self.build = core_version_info.build
-			if core_version_info._channel is not None:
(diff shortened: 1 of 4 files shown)


fuzyll:
I've added some things on top of this, including unit tests. Unfortunately, that means we can't handle the PR at the API level anymore. So, I'm closing this one in favor of the new PR I've created that also includes this API branch.

PR #7616, opened 2025-11-11
[KernelCache] Remove 'Load Selected Image and Dependencies'

### LINKED ISSUE #7615: [KernelCache] "Load Selected Image and Dependencies" crashes
- Binary Ninja Version: 5.2.8605-dev Ultimate, 20561124
- OS: macos
- OS Version: 26.0
- CPU Architecture: arm64

**Bug Description:**
Using `Load Selected Image and Dependencies` from the context menu of the image list in the KCTriage view results in a crash.

**Steps To Reproduce:**
1. Load any kernel cache.
2. Right-click on an image in the image list.
3. Click "Load Selected Image and Dependencies".

**Actual Behavior:**
Null pointer dereference.

**Additional Information:**
This is a regression from stable.

The underlying issue here is `KernelCacheController::GetImageDependencies` failing to check the return value of `BNKernelCacheControllerGetImageDependencies` and using an uninitialized variable if `BNKernelCacheControllerGetImageDependencies` returns null. This appears to have existed since the code was written, but `count` was harmlessly being initialized to 0. A change in the build configuration around 5.2.8258 means that `count` is now getting a random non-zero value, causing a crash.

I'm not even clear what "Load Selected Image and Dependencies" is expected to do in the kernel cache view. It appears to use code copied from the shared cache view and looks for `LC_LOAD_DYLIB` to determine dependencies, but kernel caches do not contain `LC_LOAD_DYLIB` commands.

bdash:
Fixes https://github.com/Vector35/binaryninja-api/issues/7615.

The dependency loading for kernel cache images was copied from the shared cache code, and looks for `LC_LOAD_DYLIB` load commands to determine the direct dependencies of an image. Images in kernel caches don't use `LC_LOAD_DYLIB` load commands so no dependencies were ever loaded. It could, however, result in a crash.

### DIFF
--- a/view/kernelcache/ui/kctriage.cpp
+++ b/view/kernelcache/ui/kctriage.cpp
@@ -208,7 +208,6 @@ QWidget* KCTriageView::initImageTable()
 
 		QAction noSelectionAction("No Images Selected", m_imageTable);
 		QAction loadImagesAction("", m_imageTable);
-		QAction loadImagesWithDepsAction("", m_imageTable);
 		if (selectedCount == 0)
 		{
 			noSelectionAction.setEnabled(false);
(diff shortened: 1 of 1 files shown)


plafosse:
Merged this manually

PR #7715, opened 2025-12-01
STABLE: Allow loading supplementary dwarf debug info in .gnu_debugaltlink

emesare:
PR for stable

0. Must not be testing on windows, see 5
1. Extract [testsupl.zip](https://github.com/user-attachments/files/23862761/testsupl.zip)
2. Open `touch`
3. You should see a bunch of warnings like `[DWARF] Failed to fetch DIE. Supplementary debug information may be incomplete.`
5. Copy `coreutils.debug` to `/tmp/thisisatestfolderfordebuginfooo/coreutils.debug` (obviously only works on unix)
6. Reopen `touch`
7. You should see types, data variables, and functions applied from debug info and no warnings/errors


emesare:
Verified that commit fixes the issue

Issue #6737, opened 2025-04-26
Call type adjustments should not apply to all calls to the same stub function within a function


bdash:
**Version and Platform (required):**
- Binary Ninja Version: 5.1.7313-dev (7e9d5c7a)
- OS: macOS 15.4.1
- CPU Architecture: arm64

**Bug Description:**
Call type adjustments created by `Function::SetAutoCallTypeAdjustment` / `Function::SetUserCallTypeAdjustment` appear to be tracked by the combination of the containing function and the address of the call instruction. This causes incorrect behavior if a call whose type is being adjusted was inlined from another function that was marked as being inlined during analysis. Since the address of the call is the address within the original function, applying a call type adjustment to a single call has the effect of applying the adjustment to every copy of the call that was inlined into the same function.

This could be fixed by tracking call type adjustments by IL instruction index instead.

**Steps To Reproduce:**
1. Open /usr/libexec/syspolicyd on macOS 15.4.x
2. Go to the definition of `-[XPCActivityJob .cxx_destruct]`
3. Right-click on one of the three calls to `_objc_storeStrong` and select `Override Call Type…`
4. Update the type of second argument to something that will display differently (`bool` or `int32_t` work).
5. Note that it updates only a single call.
6. Right-click on one of the three calls to `_objc_storeStrong` and select `Edit Function Properties…`.
7. Select "Inline during analysis" and Apply.
8. Try overriding the call type again.

**Expected Behavior:**
It should only override the types for the specific call, but instead affects all three calls within the function.

**Additional Information:**
The steps to reproduce involve the UI, but I first saw this in a workflow that programmatically applies call type adjustments. It worked fine on normal Mach-O binaries, but triggered this behavior when run on a dyld shared cache. This happened because the shared cache's workflow calls `SetAutoInlinedDuringAnalysis` on stub functions when the library containing the destination of the stub call is loaded.

WeiN76LQh:
Can confirm this is quite a problematic limitation of call type adjustments. I have a bug in a plugin which was doing call type adjustments and I didn't realise until now that the issue is due to the limitations of the API plus the fact that the DSC plugin makes heavy use of function inlining due to stubs. Essentially I was adjusting the call type of a call in a function that was inlined in another function multiple times. Each call site had different types but because they all have the same address due to the inlining I end up observing all instances having the same call type, which was incorrect in this case.

WeiN76LQh:
[...]

PR #7720, opened 2025-12-02
Allow overriding the IL source location that is used by *ILInstruction::CopyTo

bdash:
This support in `LowLevelILInstruction::CopyTo` will be used in the future to provide the option for automatic inlining during analysis to create the inlined instructions with the address of the call instruction, rather than the address of the instruction in the original function.

The `MediumLevelILInstruction` and `HighLevelILInstruction` versions are updated for consistency, with no specific use case in mind.

### DIFF
--- a/highlevelilinstruction.cpp
+++ b/highlevelilinstruction.cpp
@@ -1370,162 +1370,165 @@ void HighLevelILInstruction::VisitExprs(const std::function<bool(const HighLevel
 }
 
 
-ExprId HighLevelILInstruction::CopyTo(HighLevelILFunction* dest) const
+ExprId HighLevelILInstruction::CopyTo(HighLevelILFunction* dest, const ILSourceLocation& sourceLocation) const
 {
-	return CopyTo(dest, [&](const HighLevelILInstruction& subExpr) { return subExpr.CopyTo(dest); });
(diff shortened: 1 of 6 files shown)


plafosse:
Did you discuss this with @D0ntPanic? I know when he originally implemented it he discarded the idea of using the call site address because of unforeseen side effects that I don't remember.

rssor:
It's definitely not the general purpose solution for inlining in general, but it's needed for a very particular case where the side effect is absolutely intended.

rssor:
The only alternative would be an enormous, and expensive lift, where places we key markup on `ArchAndAddr` would have to _all_ be moved to basically `vector<ArchAndAddr>` to allow references to program locations as essentially a path through a tree, with nodes being the actual terminal location or the point at which inlining occurred, which at an API level would be almost unusable and in the core would be quite expensive because `map<ArchAndAddr, X>` is used all over the place.

bdash:
As Ryan mentions, this is intended to be used as part of a minimally invasive approach to fixing https://github.com/Vector35/binaryninja-api/issues/6737 by changing the addresses of instructions inlined from stub functions only. The default behavior when inlining instructions will still be to use the instruction's original address.

PR #7729, opened 2025-12-03
Pass arch to finalized LowLevelILFunction

trumank:
Currently it's possible to create a LowLevelILFunction with no backing function, but the finalized function does not have an arch associated with it so it's nearly unusable. For example:

```rs
let llil = LowLevelILMutableFunction::new(self.arch.clone(), None); // <- no owning function
let (instr_len, _) = self.arch.instruction_llil(&bytes, addr, &llil).unwrap();
let llil = llil.finalized();

let inst = llil.instruction_from_index(LowLevelInstructionIndex(0)).unwrap();
if let SetReg(operation) = inst.kind() {
    if let Reg(operation) = operation.source_expr().kind() {
        operation.source_reg(); // PANICS as unwrap in LowLevelILFunction::arch
    }
}
```


emesare:
Seems reasonable, I will add a unit test for this as well once merged, thank you for the illustrative example!

PR #7745, opened 2025-12-05
[Rust] Add `LowLevelILSSARegister` type

emesare:
The partial register id is the "main" id so we return that, whereas the full register id is not.

@bdash did you have any more thoughts surrounding the usage of this API?

### DIFF
--- a/rust/src/low_level_il.rs
+++ b/rust/src/low_level_il.rs
@@ -14,7 +14,7 @@
 
 use std::borrow::Cow;
 use std::fmt;
-use std::fmt::{Display, Formatter};
+use std::fmt::{Debug, Display};
 // TODO : provide some way to forbid emitting register reads for certain registers
 // also writing for certain registers (e.g. zero register must prohibit il.set_reg and il.reg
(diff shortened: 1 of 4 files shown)


bdash:
I'm not sure about this API. It does not seem obvious to a client what `id` will return for a partial register. Maybe a doc comment fixes that, or perhaps there's a clearer name.

I'm not even 100% sure whether `get_ssa_register_uses` and friends are doing the right thing with how they handle partial registers at the moment.

emesare:
> I'm not even 100% sure whether `get_ssa_register_uses` and friends are doing the right thing with how they handle partial registers at the moment.

Yea it seems that it was never intended to be queried with a partial register, I fixed up the `LowLevelILSSARegisterKind` to make it so that partial register will not be able to be passed to any of those api's. I also added some more doc comments to hopefully try and explain what `Full` and `Partial` ssa register kinds are.

https://github.com/Vector35/binaryninja-api/pull/7745/commits/11cc1662ab4bbd364e46b29e7afcfdadacb702ea

Issue #6655, opened 2025-04-16
Crash: Create Folder in Project with Description `None`: `bv.project.create_file(b'\x00', None, 'foo', None)`


utkonos:
**Version and Platform (required):**
- Binary Ninja Version: 5.0.7252-dev
- OS: macOS
- OS Version: 15.4
- CPU Architecture: x64

**Bug Description:**
This crashes Binja:

```python
bv.project.create_file(b'\x00', None, 'foo', None)
```

**Steps To Reproduce:**
1. Open binary
2. Type this in Python console: `bv.project.create_file(b'\x00', None, 'foo', None)`
3. Observe crash.

**Expected Behavior:**
No crash.

utkonos:
Additional issue: `bv.store_metadata(None, 'foo')`

utkonos:
All related to: https://github.com/Vector35/binaryninja-api/issues/6653

Issue #6653, opened 2025-04-16
Crash: Crash on Write None to Globals `bv.file.database.write_global('foo', None)`


utkonos:
**Version and Platform (required):**
- Binary Ninja Version: 5.0.7248-dev
- OS: macOS
- OS Version: 15.4
- CPU Architecture: x64

**Bug Description:**
Crashes when this line of code is run in Python console: `bv.file.database.write_global('foo', None)`

**Steps To Reproduce:**
1. Open binary
2. Type this is Python console: `bv.file.database.write_global('foo', 'bar')`
3. Type this is Python console: `bv.file.database.write_global('foo', None)`
4. Observe crash
**Expected Behavior:**
No crash

**Screenshots/Video Recording:**

![Image](https://github.com/user-attachments/assets/d58da478-2676-4dfd-8c96-2f3ce60731ed)

utkonos:
Report: red moon writes happily

utkonos:
Crash happens in project: wise river climbs gracefully

utkonos:
Aha. Here is how to do it anywhere:

```python
bv.file.database.write_global('foo', 'bar')
bv.file.database.write_global('foo', None)
```

CouleeApps:
Yep that straight up dereferences a null pointer, fix is just `assert value is not None` in the api.

utkonos:
@CouleeApps I am trying to delete that global, but I don't see an API for that, so I tried this and found the crash.

CouleeApps:
There... actually there is no API for deleting a global. Just straight up not implemented, even in the core. Seems like we've never had a need for that.

utkonos:
My use case is storing persistent data for a plugin. Globals appear to be the place to do that. Is there another place?

CouleeApps:
Database globals are a bit of a weird place to use, considering they exist outside of the snapshot / undo system. I would recommend trying BinaryView Metadata, which gets saved in snapshots and can be undone if desired.

utkonos:
I noticed that when I add a database global, the indicator that the database has changes and should be saved does not appear.

CouleeApps:
Yeah, in the future we should probably add a big warning notice to the Database family of APIs saying "this is probably not what you expect, try BinaryView metadata instead."

utkonos:
I kindof felt like I was in a bad part of town.

utkonos:
[...]

PR #7748, opened 2025-12-05
Fix show_message_box crashing with `None` description in Python

### LINKED ISSUE #7747: Null Pointer Dereference via Interaction Module in Python API
- Binary Ninja Version: 5.3.8733-dev Ultimate, f55f20f8
- OS: macos
- OS Version: 26.1.0
- CPU Architecture: arm64

**Bug Description:**
The `show_message_box` function in the interaction module in the Python API has a possible null pointer dereference ala issue #6653. If triggered Binary Ninja crashes.

**Steps To Reproduce:**
Run either of these in the Python console:

```python
show_message_box(None, 'Bar', MessageBoxButtonSet.YesNoButtonSet)
```

```python
show_message_box('Foo', None, MessageBoxButtonSet.YesNoButtonSet)
```

**Expected Behavior:**
No crash.

utkonos:


mostobriv:
As `str(None) == "None"`, current fix may lead to cases when title or text would be just "None" instead of nothing

utkonos:
@mostobriv, that's not correct. This fix leads to the title or text being "None" instead of crashing Binary Ninja.

I opened an issue about this here:
https://github.com/Vector35/binaryninja-api/issues/7747

I found a number of locations where this type of crash occurs back in April. There are tickets for those here:
https://github.com/Vector35/binaryninja-api/issues/6653
and here:
https://github.com/Vector35/binaryninja-api/issues/6655

The fix for those was performed by people at Vector35, but I figured that I'd go ahead and submit a PR with the same way of fixing this one rather than waiting. Here is the commit for those fixes:
https://github.com/Vector35/binaryninja-api/commit/0bf8a9632ea3ee570b4d56dd269d61fde28ca910

PR #7757, opened 2025-12-09
Specify fixed underlying types for enums exposed by core

bdash:
This allows a few widely-used enums to be shrunk from 4 bytes to 1 byte, improving packing when they're used as struct members.

To remain compatible with C, we follow CoreFoundation's approach and use a macro when defining the enum:

```
#if defined(__cplusplus) || __has_extension(c_fixed_enum)
	#define BN_ENUM(type, name) enum name : type
#else
	#define BN_ENUM(type, name) typedef type name; enum 
#endif

BN_ENUM(uint8_t, SomeEnum)
{
   ... 
}
```

In C++ and C23 this will expand to an enum with a fixed underlying type. In older C language versions, this will result in the enum type being a typedef of the underlying type, with an unnamed enum providing the enum values.

Minor changes were needed within the Python bindings to update places that made assumptions about the underlying type of the enums.

### DIFF
--- a/binaryninjacore.h
+++ b/binaryninjacore.h
@@ -37,14 +37,14 @@
 // Current ABI version for linking to the core. This is incremented any time
 // there are changes to the API that affect linking, including new functions,
 // new types, or modifications to existing functions or types.
-#define BN_CURRENT_CORE_ABI_VERSION 151
+#define BN_CURRENT_CORE_ABI_VERSION 152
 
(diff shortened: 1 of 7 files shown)


plafosse:
I've read through all the code and I generally am in favor. I guess my one thought that I haven't fully thought through is default enum signedness on clang vs msvc. I would suppose your code has resolved any cross compiler differences now that they're all explicitly unsigned. I'm just trying to think if there are any negative side effects that haven't been considered.

PR #7758, opened 2025-12-09
Use constexpr arrays for metadata about IL instruction operands

bdash:
This avoids constructing numerous unordered maps and vectors at library load time in every plug-in that uses the C++ API. Additionally, the arrays allow for more efficient look-ups.

### DIFF
--- a/highlevelilinstruction.cpp
+++ b/highlevelilinstruction.cpp
@@ -18,7 +18,12 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 // IN THE SOFTWARE.
 
-#include <string.h>
+#include <algorithm>
+#include <array>
+#include <cstring>
+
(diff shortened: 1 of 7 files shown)


plafosse:
In general I'm a huge fan of this PR. I appreciate all the static asserts which should help this code from breaking in the future.  I don't see downsides to any of this. This seems like it should shave off some bytes on every IL instruction allocation which is a nice win too.

I was trying to see if anything could be made consteval to provide stronger guarantees of compile-time generation but I'm not sure about which parts could be converted or what the compiler support is for consteval.

bdash:
In practice, what we care about is that the static variables are compile-time constants and not initialized at runtime. Being declared as `static constexpr` provides that guarantee. Adding `consteval` would be useful if what we care about is guaranteeing that a given function is only ever used at compile time. That's less important here since the functions are implementation details.

plafosse (commented):
Looks good to me.

PR #7773, opened 2025-12-11
Fix formatting so this displays well on docs site. Fixes #7772

### LINKED ISSUE #7772: Intended Formatting Lost Between Github and Docs Site
Markdown formatting on docs site is slightly less tolerant of spacing than Github's. There are a few places where it loses the intended formatting.

Image

vs

Image

and

Image

vs 

Image

The problem is missing newlines here:

Image

utkonos:


emesare:
Thank you!

PR #7777, opened 2025-12-11
Add language about workflow ephemerality. Fixes #7771

### LINKED ISSUE #7771: Workflow Docs Should Mention Ephemerality of Workflows Per Binja Execution
Workflows

**What concept or API needs more documentation?**
The Workflows documentation should mention that workflows are regenerated every time binja executes and are then modified by plugins as they are loaded. So the workflow changes made in examples are ephemeral for the current binja execution.

Maybe put that here:
https://docs.binary.ninja/dev/workflows.html#workflow-fundamentals

Without this being mentioned, the first example which deletes everything from the metaAnalysis and then registers the empty clone over the original sounds quite destructive.

utkonos:


bpotchik:
Thanks, refined that language some, will show up in a future build.

PR #7779, opened 2025-12-11
[Rust] Restructure type APIs into `types` module

emesare:
This helps with documentation, giving a single module for those working with types to find related APIs

Also split out enumeration and structure APIs into their own file, since they have their own backing data separate from `Type`.

This does not actually break any plugins since these modules that were moved and _not_ re-exported are not used as of yet.

The APIs which are being used have re-exports that allow existing code to correctly import.


emesare:
One unfortunate thing is we keep QualifiedName in the types module, I think we should move it out and re-export it within types module with a deprecation notice.

Issue #7702, opened 2025-11-29
Add std::vector::reserve to BinaryView::GetAllEntryFunctions()


monotonik-guy:
For some reason for this API `reserve` not used. I might to create PR, but don't have PC nearly.

```cpp
vector<Ref<Function>> BinaryView::GetAllEntryFunctions()
{
	size_t count;
	BNFunction** funcs = BNGetAllEntryFunctions(m_object, &count);
	if (count == 0)
		return {};

	vector<Ref<Function>> result;
	for (size_t i = 0; i < count; i++)
		result.push_back(new Function(BNNewFunctionReference(funcs[i])));
	BNFreeFunctionList(funcs, count);
	return result;
}
```

PR #7780, opened 2025-12-11
Add missing std::vector reserve calls

SmoothHacker:
Fixes issue #7702 and adds `vector.reverse()` calls in other places


SmoothHacker:
More reserve calls being added. [performance-inefficient-vector-operation](https://clang.llvm.org/extra/clang-tidy/checks/performance/inefficient-vector-operation.html) is a very good clang-tidy check for this

CouleeApps (commented):
LGTM. Probably a mild perf win too, thanks for the contribution

CouleeApps (approved):

Issue #7784, opened 2025-12-12
Change Plugin Load Order So User Developed Plugins Can Access Official Plugin APIs


utkonos:
**What is the feature you'd like to have?**
Please change the plugin load order to the following:
1. Core
2. Core UI
3. Official
4. Community
5. Local

**Is your feature request related to a problem?**
Today, the load order is such that a local or community plugin is loaded before the official and core ui plugins. Therefore they cannot import from those plugins. The main problem in this case is Sidekick scripting API is not available for import in python plugins.

**Are any alternative solutions acceptable?**
No sure, but I will try to hack something b/c I really need this.

**Additional Information:**

<img width="908" height="677" alt="Image" src="https://github.com/user-attachments/assets/5f85528d-9fa6-428d-8511-50bd620540af" />

ElykDeer:
Here's a workaround I've been using to import official helpers from my unofficial/user plugin: https://github.com/ElykDeer/tanto-devtools/blob/57b97bdf09982a9fb34c8546e308d014274f270b/dominator_graphs.py#L1-L9

PR #7789, opened 2025-12-13
Add Plugin Dependency APIs for the Python API

### LINKED ISSUE #7785: Implement `BNAddOptionalPluginDependency` for Python API
Please implement the equivalent to `BNAddOptionalPluginDependency` for the Python API.

Two problems:

1. If I make a workflow plugin, I want to make sure that I can wait until another workflow plugin has already cloned metaanalysis so that I can then clone metaanalysis.
2. If I make any plugin, I want to be able to wait for another official, community, or local plugin to load in the case that I need to import functions, classes, APIs from those plugins.

Will try to hack something, but I don't know of an alternative now.

SmoothHacker:
Fixes #7785

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -406,6 +406,11 @@ def get_memory_usage_info() -> Mapping[str, int]:
 	core.BNFreeMemoryUsageInfo(info, count.value)
 	return result
 
+def add_required_plugin_dependency(plugin_name: str) -> None:
+	core.BNAddRequiredPluginDependency(plugin_name)
+
+def add_optional_plugin_dependency(plugin_name: str) -> None:
+	core.BNAddOptionalPluginDependency(plugin_name)
 
 def load(*args, **kwargs) -> BinaryView:
 	"""


CouleeApps:
The idea for this is good, but implementing these APIs like this won't have any effect. This is because python is loaded all together as one core plugin, which loads user python plugins separately from the core plugin system, which these APIs inform. Accomplishing something like this will need modifications to the closed-source python plugin itself and likely need modifications to the core too if python plugins have any core plugin dependencies (or vice versa).

0cyn:
This is particularly difficult with python plugins as these are loaded and initialized in the same step (importing the module), whereas core plugins dont initialize until their CorePluginInit function is called (in which case we can check their dependencies before loading them) 

So this functionality would require a change or workaround to that, or more likely a way to list dependencies in the Metadata itself. We do want to discuss this in the future but at the moment its not currently possible for python plugins to specify plugin dependencies before load

utkonos:
@0cyn Without actually implementing an entire dependency feature for Python plugins, is there a middle ground that could be implemented with a more simple change? Having inter-plugin dependencies would be fantastic, but the very most basic issue is that local plugins are loaded first, then community, then official. It there were a way for just that order to be reversed, it would at least alleviate the issue of community and local plugins not being able to access APIs exposed by official plugins: the scripting API in Sidekick being the one that I am looking to use in a workflow plugin. I want to be able to clone and register the metaanalysis workflow after Sidekick has loaded so that I can use its scripting API in my local workflow plugin.

0cyn:
We're looking at merging a refactoring of our plugin loading sometime ideally in January. When this lands we can ensure the load order is core -> user -> official -> community for python plugins.

Thank you for the feedback!

utkonos:
@0cyn Can that be core -> official -> community -> user? This way workflows are cloned and registered in descending "officialdom"?

0cyn:
That would make sense. The override-by-name order in the rewrite is core -> user -> repository however I'm not immediately sure if this specific issue would impact that.

I'll bring this up to make sure we're not missing anything, but that load order should be fine.  Thank you!

utkonos:
@0cyn This is what I'm trying to solve for. I want the code in this plugin to be a workflow node. But `Vector35_Sidekick.scripting.api` isn't loaded yet. And even if I make a Community plugin, it would still have that problem.

I have a separate, specific issue open for this:
https://github.com/Vector35/binaryninja-api/issues/7784

image

This plugin right now gathers data about various functions. I have been making a Sidekick Workbench script that does even more and also stores the results in a Sidekick Index. All of that I now want in a Workflow node.

Issue #7759, opened 2025-12-09
PossibleValueSet merging api


mostobriv:
**What is the feature you'd like to have?**
Sometimes when developing workflows or some scripts I need to traverse PVS and then get the accumulated PVS of some variable (mostly this happened in cases when BN automatically fails to build PVS by itself), in this cases ability to merge one PVS with another is helpful, however current python api is lacking of such feature. 

So, I think, having `__add__` and `__iadd__` for `PossibleValueSet` class would be a handy QoL feature.

- `UndeterminedValue` + `Any` == `UndeterminedValue` or exception
- `ConstantValue / ConstantPointerValue` + `ConstantValue / ConstantPointerValue` == `InSetOfValues` or (in case of equality) `ConstantPointerValue`
- `InSetOfValues` + `InSetOfValues` == `InSetOfValues`
- `NotInSetOfValues` + `NotInSetOfValues` == `NotInSetOfValues`
- etc.

mostobriv:
If it worth a shot, I can try to implement it and make a PR, just didn't want to bother you guys with meaningless or unwanted PRs.

CouleeApps:
We have a bunch of these implemented in the core. It may be worth exposing them or moving the source to the public repo so that others can use them as well. CC core/possiblevalueset.h

negasora:
APIs for PVS operations added in `5.3.8793-dev` using e.g. `.add(other, size)`

PR #7792, opened 2025-12-15
[Python] Added __add__ and __iadd__ to PossibleValueSet

mostobriv:
Addressing #7759 

Current version of PR cover the cases:
- `Undetermined + Any`, `Any + Undetermined`
- `ConstantValue | ConstantPointerValue` + `ConstantValue | ConstantPointerValue`
- `InSetOfValues + InSetOfValues`, `InSetOfValues + Any with "value" attribute`
- `NotInSetOfValues + NotInSetOfValues`

Didn't add any combination with `int` or some other type different from `PossibleValueSet` as i'm not sure api should allow such merging combo.

### DIFF
--- a/python/variable.py
+++ b/python/variable.py
@@ -340,6 +340,35 @@ class PossibleValueSet:
 		if self._type == RegisterValueType.ReturnAddressValue:
 			return "<return address>"
 		return "<undetermined>"
+	
+	def __add__(self, other: "PossibleValueSet") -> "PossibleValueSet":
+		if not isinstance(other, self.__class__):
+			return NotImplemented
+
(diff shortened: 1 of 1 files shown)


seekbytes:
I understand that `+` is a common operator in python and it's practical, but I wouldn't mind if we move this operation to a `union` method. Other than that, looks good :)

negasora:
Thanks for the PR! However, we already have this logic implemented in core so the path forward is to expose PVS operations through the API. Having to duplicate the same PVS operation logic in each API shouldn't be required when they could just call the same core methods, and may lead to inconsistent behavior if the implementations differ.

PR #7798, opened 2025-12-16
Annotate deprecated C++ APIs

### LINKED ISSUE #7802: Annotate deprecated C++ APIs with [[deprecated(msg)]]
C++ APIs that have been deprecated are currently only marked as deprecated in their documentation comments. Only the most diligent API users are aware of these as they require closely read the API header or documentation to identify which methods are deprecated, and then searching their codebase to identify whether they're used.

C++14 introduced the [`[[deprecated(msg)]]`](https://en.cppreference.com/w/cpp/language/attributes/deprecated) attribute that can be added to symbols. Compilers will emit warnings when deprecated symbols are used. This makes it immediately obvious to API users when they are still using a deprecated API.

If a user builds with `-Werror`, meaning deprecation warnings would break their build, and they're unable to immediately update the usage of the deprecated API, they can either:
1. Add `-Wno-error=deprecated-declarations` to disable warnings as errors only for deprecation warnings. I'm not sure if there's an MSVC equivalent for this.
2. Wrap usage of the deprecated API in:
   ```
   #pragma GCC diganostic push
   #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
   // Use of deprecated API here
   #pragma GCC diagnostic pop
   ```

   The MSVC equivalent involves `#pragma warning(push)` and `#pragma warning(disable: 4996)`.

Both options are temporary measures, intended only to quiet noise while the API user works towards updating the code. Deprecated APIs will be eventually be removed and code using them will need to be updated or it will inevitably break.

bdash:
A `BN_DEPRECATED` macro is introduced that expands to a `[[deprecated(msg)]]` attribute. Using functions that are annotated as deprecated will generate a compiler warning. This should help make users aware that they should migrate to replacement APIs, making it safer for us to remove the deprecated APIs in a future release.

Fixes #7802.

### DIFF
--- a/base/compiler.h
+++ b/base/compiler.h
@@ -1,0 +1,73 @@
+// Copyright (c) 2026 Vector 35 Inc
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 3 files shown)


plafosse:
This is great but but looks like some unrelated whitespace changes slipped in too.

plafosse (approved):

bdash:
> This is great but but looks like some unrelated whitespace changes slipped in too.

Not sure if it's what you're referring to, but the changes in platform_linux.cpp are updating code that was using a deprecated API to use the replacement.

PR #7808, opened 2025-12-18
[Rust] Replace `log` with `tracing`

### LINKED ISSUE #6863: [Rust API] Improve logger documentation
- Binary Ninja Version: 5.0.7290-stable
- Edition: Commercial
- OS: macOS
- OS Version: 14.5
- CPU Architecture: Apple M-series (ARM64)

**Bug Description:**
When using the Rust API to develop a plugin, I attempted to enable logging in headless mode using:

```rust
binaryninja::logger::Logger::new("Test").init();
````

However, no log output is produced, even when explicitly setting the log level. This makes it difficult to debug plugins in headless mode.

Most existing Rust plugin examples use `print*` macros instead of the `log` crate, which suggests that `logger.init()` may not be supported in this context — or perhaps I’m missing a required step.

**Steps To Reproduce:**

1. Create a headless plugin using the Binary Ninja Rust API.
2. Call `binaryninja::logger::Logger::new("Test").init()` and attempt to log a message using `log::info!`, `warn!`, etc.
3. Run the plugin in headless mode using `cargo run`.
4. Observe that no output is produced in the console.

**Expected Behavior:**
I expected that `logger.init()` would configure logging correctly and output log messages to the console when running in headless mode.

N/A

**Binary:**
Not applicable — this issue concerns plugin development using the Rust API.

**Additional Information:**
Possibly related to issue #5101 and issue #1990.

Is this the intended behavior? Should I be using a third-party logging solution instead of the built-in Binary Ninja logger when working in headless mode?

### PR DESCRIPTION
**Pull request by emesare**
- Added more documentation
- Replaced global named logger for plugins, fixing the issue when the CU has multiple (e.g. statically linked demo)
- ~~Explicitly name the target for the log event, this is how it must be done to guarantee we send it to the correct one~~
- Simplified some misc code
- Added `session_id` event field, to send logs to a particular session (e.g. binary view scoped logs)

This is a breaking change, but I believe there is no better time to make it, we cannot continue to use the `log` crate, it is too limited for our needs.

~~Still need to finish docs and cross link more between the `logger` module and the `tracing` module, as well as provide some examples in the top level module docs.~~

~~Still need to place more session id spans in relevant callbacks so that they automatically get scoped to a bv.~~

~~TBD if we keep the tracing re-exports for the macros, we might instead let callers use the non re-exported version. or we might move the code into `logger` and then re-export tracing at the top level, or we might not re-export at all and assume the caller includes `tracing` in Cargo.toml.~~

### CODE REVIEW

**Comment by emesare** [comment]
Resolves https://github.com/Vector35/binaryninja-api/issues/6863

**Comment by emesare** [comment]
Force pushed removing explicit targets in log calls, replaced with rewrite rules within the registered tracing `BinaryNinjaLayer`

**Comment by emesare** [comment]
Migration guide:

1. Add/Remove crates:

```diff
diff --git a/plugins/pdb-ng/Cargo.toml b/plugins/pdb-ng/Cargo.toml
--- a/plugins/pdb-ng/Cargo.toml
+++ b/plugins/pdb-ng/Cargo.toml
@@ -16,1 +16,2 @@
 regex = "1"
+tracing = "0.1"
```

2. Replace usage of `log` helper macros:

```diff
diff --git a/plugins/warp/src/plugin/load.rs b/plugins/warp/src/plugin/load.rs
--- a/plugins/warp/src/plugin/load.rs
+++ b/plugins/warp/src/plugin/load.rs
@@ -132,4 +133,4 @@
             Err(e) => {
-                log::error!("Failed to read signature file: {}", e);
+                tracing::error!("Failed to read signature file: {}", e);
                 return;
             }
```

> You can technically keep using `log::info!` and friends, see [tracing documentation](https://docs.rs/tracing/latest/tracing/#log-compatibility) for more details.

3. Replace `Logger::new` at top of plugin init with `tracing_init!`:

```diff
diff --git a/plugins/pdb-ng/src/lib.rs b/plugins/pdb-ng/src/lib.rs
--- a/plugins/pdb-ng/src/lib.rs
+++ b/plugins/pdb-ng/src/lib.rs
@@ -721,1 +729,1 @@
-    Logger::new("PDB").init();
+    binaryninja::tracing_init!("PDB Import");
```

rust/plugin_examples/data_renderer/Cargo.toml
```diff
@@ -10,4 +10,4 @@ crate-type = ["cdylib"]
 binaryninjacore-sys = { path = "../../binaryninjacore-sys" }
 binaryninja = { path = "../.." }
 uuid = "1.18.1"
-log = "0.4.27"
+tracing = "0.1"
```
bdash: May as well leave the newline at EOF here.

rust/src/tracing.rs
```diff
@@ -0,0 +1,362 @@
...
+    /// target will be rewritten to "MyPlugin::commands::analyze", assuming no other rewrites occur.
+    pub target_mappings: Vec<(String, String)>,
+    /// Whether the default target should be formatted to be displayed in the "common" logger name
+    /// format for Binary Ninja. This formatting does _not_ apply to targets explicitly set, rather
+    /// than only when a target is omitted.
+    ///
+    /// # Example
+    ///
+    /// Given the target "my_crate::commands::analyze" the target will be rewritten to "MyCrate.Commands.Analyze".
+    pub formatted_target: bool,
```
bdash: This is describing behavior that will be taken, so I think it should be `format` rather than `formatted`.

rust/src/tracing.rs
```diff
@@ -0,0 +1,362 @@
...
+pub struct BinaryNinjaLayer {
+    /// Rewrite mappings for the default target.
+    ///
+    /// # Example
+    ///
+    /// Given the target "my_crate::commands::analyze" and the mapping ("my_crate", "MyPlugin") the
+    /// target will be rewritten to "MyPlugin::commands::analyze", assuming no other rewrites occur.
+    pub target_mappings: Vec<(String, String)>,
+    /// Whether the default target should be formatted to be displayed in the "common" logger name
+    /// format for Binary Ninja. This formatting does _not_ apply to targets explicitly set, rather
```
bdash: It might be clearer to phrase this in the positive. Something like: "This formatting will only be applied when the target name is implicit. Explicitly provided target names will be preserved as-is."

bdash (approved):
LGTM, with some minor comments.

plafosse:
Mark already approved this. I'm in favor too. Just setting Mark as assigned so my filtering works :)

PR #7861, opened 2026-01-07
Add APIs to allow calling conventions to provide requirements for being considered during heuristic calling convention detection

D0ntPanic:
Several modern languages, like Rust and Swift, have custom calling conventions that input or output in registers that are normally reserved as callee saved registers. To account for this, we need to add calling conventions that define these additional register uses.

Unfortunately, adding these calling conventions will cause the heuristic calling convention detection to run into a lot of cases where there are multiple apparently equivalent solutions to the optimal calling convention. Most of the time we want the base calling convention that doesn't include the non-standard register usage.

To allow for this, this commit adds two new APIs. One provides the set of arguments that must be present, and the other provides the set of registers that must be clobbered.

For non-standard inputs, the calling convention should supply the register as an argument register, and also supply it as a required argument register. This will cause the calling convention detection code to avoid the calling convention unless usage of the non-standard input register is actually found. It will also allow the calling convention code to avoid it even if there are no arguments to the function (in this case, just adding the register as a normal argument register will cause a conflict and may not choose the desired base convention).

For non-standard outputs, supply the register as a required clobbered register. This will avoid detecting the convention when the non-standard output is not written.

### DIFF
--- a/arch/x86/arch_x86.cpp
+++ b/arch/x86/arch_x86.cpp
@@ -3814,6 +3814,11 @@ public:
 		return vector<uint32_t>{ XED_REG_ECX };
 	}
 
+	virtual vector<uint32_t> GetRequiredArgumentRegisters() override
+	{
+		return vector<uint32_t>{ XED_REG_ECX };
+	}
+
 	virtual bool IsStackAdjustedOnReturn() override
 	{
 		return true;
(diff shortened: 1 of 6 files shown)


plafosse:
I'm curious if this will at all help with tricore calling convention. @rssor and @emesare might have some input here.

rssor:
Won't help but won't cause problems for either (won't conflict with the other in-progress cc work either)

emesare:
> I'm curious if this will at all help with tricore calling convention. @rssor and @emesare might have some input here.

Wont help with tricore cc, that is partly addressed in a [separate core pr](https://github.com/Vector35/binaryninja/pull/1215). The other part has to do with hardware preserved registers which is another separate issue yet to be tackled (why we preserve upper ctx registers explicitly around calls).

D0ntPanic:
This was merged with the calling convention refactor

PR #7878, opened 2026-01-13
Make add_segments and add_sections apis pythonic

### LINKED ISSUE #7725: add_auto_segments and add_auto_sections both require native objects
Both of the `add_auto_*` APIs that make batch sections and segments in python currently require you to construct a native object to use the API. This is not pythonic and we should likely match the syntax of the singular APIs and take a list of a list of those parameters. Or at a minimum we should have helper wrappers to make it more straight forward to do that. 

https://dev-api.binary.ninja/search.html?q=add_auto&check_keywords=yes&area=default

`add_auto_segments(segments: List[BNSegmentInfo])→ None[source]`

plafosse:
This introduces two new dataclasses SegmentInfo and SectionInfo which hold information about their respective objects. These are used for specifying information to a BinaryView while class Segment and Section are still the objects used after registration. Additionally I've created two helper functions in Segment/Section which return the respective Info structure which could be useful if you wanted to modify existing Segments/Sections

Currently these two APIs are only used by the add_segments/add_sections APIs but maybe more useful in the future.

Resolves https://github.com/Vector35/binaryninja-api/issues/7725

Unit tests here: https://github.com/Vector35/binaryninja/commit/74f276bd18f384d16dc66ba78cbc986b9e02d9e4

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -1618,6 +1618,27 @@ class BinaryViewType(metaclass=_BinaryViewTypeMetaclass):
 		"""
 		BinaryViewEvent.register(BinaryViewEventType.BinaryViewInitialAnalysisCompletionEvent, callback)
 
+@dataclass
+class SegmentInfo:
+	"""
+	This class helper class holds Segment information used to describe segments when creating a BinaryView.
+	See BinaryView.add_auto_segments and BinaryView.add_user_segments for usage.
+	See class Segment for segment information retrieval from an existing BinaryView.
+	"""
+	start: int
+	length: int
+	data_offset: int
+	data_length: int
+	flags: 'SegmentFlag'
+
+	def _to_core_struct(self) -> core.BNSegmentInfo:
+		info = core.BNSegmentInfo()
+		info.start	  = self.start
+		info.length	  = self.length
+		info.dataOffset = self.data_offset
+		info.dataLength = self.data_length
+		info.flags	  = self.flags
+		return info
 
 class Segment:
 	"""
@@ -1732,6 +1753,16 @@ class Segment:
 	def auto_defined(self) -> bool:
 		return core.BNSegmentIsAutoDefined(self.handle)
 
+	@property
+	def segment_info(self) -> SegmentInfo:
+		return SegmentInfo(
+			start=self.start,
+			length=self.length,
+			data_offset=self.data_offset,
+			data_length=self.data_length,
+			flags=core.BNSegmentGetFlags(self.handle)
+		)
+
 
 class SegmentDescriptorList(list):
 	def __init__(self, image_base: int):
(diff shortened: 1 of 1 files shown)


python/binaryview.py
```diff
@@ -1732,6 +1753,16 @@ def data_end(self) -> int:
 	def auto_defined(self) -> bool:
 		return core.BNSegmentIsAutoDefined(self.handle)
 
+	@property
+	def segment_info(self) -> SegmentInfo:
```
emesare, line 1757: maybe just call this info seeing as this is a property on Segment

python/binaryview.py
```diff
@@ -1886,6 +1949,21 @@ def length(self):
 	def end(self) -> int:
 		return self.start + self.length
 
+	@property
+	def section_info(self) -> SectionInfo:
```
emesare, line 1953: same nitpick as segment_info

python/binaryview.py
emesare: need to update this

emesare: need to update this

emesare (changes requested):
We also need to update the C++ API as per the comment in the referenced issue, could do that as a seperate PR I do not really mind.

python/binaryview.py
plafosse: there is already `info_data` and `info_section`  I was trying to pick something that was more distinct

plafosse:
> We also need to update the C++ API as per the comment in the referenced issue, could do that as a seperate PR I do not really mind.

Much less important IMO.  For SegmentInfo it would basically just be `typedef struct BNSegmentInfo SegmentInfo` For SectionInfo it would we'd probably want something like this
```c++
struct SectionInfo
{
	BNSectionInfo info;
	SectionInfo(const string& name, uint64_t start, uint64_t length, BNSectionSemantics semantics,
	    const string& type, uint64_t align, uint64_t entrySize, const string& linkedSection,
	    const string& infoSection, uint64_t infoData)
	    : info.name(name.c_str()), info.start(start), info.length(length), info.semantics(semantics),
	      info.type(type.c_str()), info.align(align), info.entrySize(entrySize),
	      info.linkedSection(linkedSection.c_str()), info.infoSection(infoSection.c_str()),
	      info.infoData(infoData)
	{
	}
};


void BinaryView::AddAutoSections(const vector<SectionInfo>& sections)
{
	vector<BNSectionInfo> sectionInfos;
	sectionInfos.reserve(sections.size());
	for (const auto& section : sections)
		sectionInfos.push_back(section.info);
	BNAddAutoSections(m_object, sectionInfos.data(), sectionInfos.size());
}

```
Unless I'm missing something.

emesare:
> > We also need to update the C++ API as per the comment in the referenced issue, could do that as a seperate PR I do not really mind.
> 
> Much less important IMO. For SegmentInfo it would basically just be `typedef struct BNSegmentInfo SegmentInfo` For SectionInfo it would we'd probably want something like this
> 
> ```c++
> struct SectionInfo
> {
> 	BNSectionInfo info;
> 	SectionInfo(const string& name, uint64_t start, uint64_t length, BNSectionSemantics semantics,
> 	    const string& type, uint64_t align, uint64_t entrySize, const string& linkedSection,
> 	    const string& infoSection, uint64_t infoData)
> 	    : info.name(name.c_str()), info.start(start), info.length(length), info.semantics(semantics),
> 	      info.type(type.c_str()), info.align(align), info.entrySize(entrySize),
> 	      info.linkedSection(linkedSection.c_str()), info.infoSection(infoSection.c_str()),
> 	      info.infoData(infoData)
> 	{
> 	}
> };
> 
> 
> void BinaryView::AddAutoSections(const vector<SectionInfo>& sections)
> {
> 	vector<BNSectionInfo> sectionInfos;
> 	sectionInfos.reserve(sections.size());
> 	for (const auto& section : sections)
> 		sectionInfos.push_back(section.info);
> 	BNAddAutoSections(m_object, sectionInfos.data(), sectionInfos.size());
> }
> ```
> 
> Unless I'm missing something.

you cant pass in a string by ref, take a pointer to the data and expect the lifetime to be of the SectionInfo, but other than that its fine

plafosse:
Merged without the C++ changes for now

PR #7911, opened 2026-01-22
Fix slice handling in FunctionList, BasicBlockList, and TagList

3rdit:
The standard Python slicing syntax, i.e. `[:]`, `[2:]`, `[:5]`.. (anything with an open end) fail with TypeError on FunctionList, BasicBlockList and TagList. Other valid slices like `[0:len(list)]` would also cause an IndexError.

The slice bounds checking compares `i.start` and `i.stop` directly against integers, but Python sets these to `None` for open-ended slices. This means comparing `None < 0` raises TypeError. The check `i.stop >= len(self)` also rejects valid Python where stop equals length.

This PR replaces the manual bounds checking with `slice.indices()` that will handle None values and normalise the bounds correctly. Other places in the codebase already do this.

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -2466,10 +2466,8 @@ class FunctionList:
 			return _function.Function(self._view, core.BNNewFunctionReference(self._funcs[i]))
 		elif isinstance(i, slice):
 			result = []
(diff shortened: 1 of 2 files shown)


plafosse (approved):
Thanks for the fix. I'm adding tests and merging today.

plafosse:
Merged this yesterday!

PR #7925, opened 2026-01-29
Share context between BB recovery, lifting, and disassembly

zznop:
Adds a GetInstructionTextWithContext callback to the architecture class that can be used to pass data from AnalyzeBasicBlocks. This same context is also supplied to LiftFunction and allows for supplying shared function and/or binary view level information across basic block analysis, function lifting, and disassembly text rendering

### DIFF
--- a/rust/src/architecture.rs
+++ b/rust/src/architecture.rs
@@ -42,6 +42,8 @@ use std::{
     mem::MaybeUninit,
 };
 
+use std::ptr::NonNull;
+
 use crate::function_recognizer::FunctionRecognizer;
 use crate::relocation::{CustomRelocationHandlerHandle, RelocationHandler};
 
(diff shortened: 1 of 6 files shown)


rust/src/architecture.rs
```diff
@@ -1485,6 +1627,15 @@ where
         custom_arch.lift_function(function, &mut context)
     }
 
+    extern "C" fn cb_free_function_arch_context<A>(ctxt: *mut c_void, context: *mut c_void)
+    where
+        A: 'static + Architecture<Handle = CustomArchitectureHandle<A>> + Send + Sync,
+    {
+        let custom_arch = unsafe { &*(ctxt as *mut A) };
+        let context = NonNull::new(context);
+        custom_arch.free_function_arch_context(context);
```
rssor: we don't need this function to be visible to bindings consumers.

rather, just make sure we take ownership of the function arch context here and ensure it gets dropped. see how the custom BinaryView implementation handles closed binary views. the bindings simply take ownership of the whole custom type, 'leak' it over to the core, then when the core is done with it it gets instantiated back into a Box and dropped. contains all the unsafe code to the bindings and the user can't mess things up

rssor (changes requested):

zznop:
Added in `5.3.9128-dev`

PR #7947, opened 2026-02-06
Handle Universal binaries through container transform system.

bpotchik:


bpotchik:
- Refactor Universal (Fat Mach-O) handling from BinaryViewType to Transform
- Preserve headless auto-selection behavior when no architecture preference is set
- Add per-transform last-selected preselection in Container Browser (QSettings)
- Update setting descriptions for architecturePreference and openWithOptions
- Add container transform path tests alongside legacy BinaryViewType tests

PR #7965, opened 2026-02-13
Minor Python bugfixes and type issues

nullableVoidPtr:
- Fixed parameters for `get_incoming_flag_value`
- Disambiguated return type for `TypedDataAccessor.__getitem__` (previously `TypedDataAcccessor | List[TypedDataAccessor]` no matter the type for `key`)
- Fixed parameter type conflicts in `types`, primarily in `TypeBuilder`s
- Disambiguated return type for data/text/line searchers in a BinaryView
- Specify callback type and convert `match` callback param for `BinaryView.find_all_text`
- Allow user-specified joins for a Python `QualifiedName`
- Fix narrowed sequence elements passed into some functions causing a type mismatch
  - `List` is invariant, but `Sequence` is covariant

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -88,6 +88,7 @@ PathType = Union[str, os.PathLike]
 InstructionsType = Generator[Tuple[List['_function.InstructionTextToken'], int], None, None]
 ProgressFuncType = Callable[[int, int], bool]
 DataMatchCallbackType = Callable[[int, 'databuffer.DataBuffer'], bool]
+TextMatchCallbackType = Callable[[int, str, 'lineardisassembly.LinearDisassemblyLine'], bool]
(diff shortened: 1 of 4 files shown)


nullableVoidPtr:
For additional context, these are issues I came across polishing [my plugin](github.com/nullableVoidPtr/LyticEnzyme)

nullableVoidPtr:
That being said, I feel like a `TypeProtocol` would be rather useful for easier typing when `Type` or `TypeBuilder` can be used - though I have not tested this.

CouleeApps:
These are good improvements! I only saw the following that needed changing:

* The separator to QualifiedName is internally called `join`
* Need to construct QualifiedName with `join` from the core struct
* Also need to handle NameSpace type too
* Using `Self` for types is a Python 3.11 feature, which exceeds our minimum supported Python of 3.9

Given these are all pretty trivial changes, I just added them onto your PR. Thanks for the contribution!

PR #7978, opened 2026-02-21
Python: Replace NotImplemented with default values or errors

Alkalem:
In the python API the builtin value [NotImplemented](https://docs.python.org/3/library/constants.html#NotImplemented) is sometimes used as return value. This is intended for operator implementations like `__eq__`, but may cause problems in other contexts. This PR only touches `NotImplemented` outside of operator implementations.

As an example why the current implementation might be problematic: The base `Architecture` class currently returns `NotImplemented` for patching methods. This leads to boolean evaluation of `NotImplemented` for `is_xy_patch_available` (evaluated to True). For that reason, all patches are listed as available, but their execution will finally fail because `NotImplemented` as return value for the patch methods is not of type `Optional[bytes]`.

For unimplemented methods in API base classes `NotImplementedError` seems like a better solution if there are no good default values / default implementations available.

### DIFF
--- a/python/architecture.py
+++ b/python/architecture.py
@@ -2400,7 +2400,7 @@ class Architecture(metaclass=_ArchitectureMetaClass):
 			b'\\x0f\\x84\\x04\\x00\\x00\\x00'
 			>>>
 		"""
-		return NotImplemented
+		raise NotImplementedError
 
 	def is_never_branch_patch_available(self, data: bytes, addr: int = 0) -> bool:
 		"""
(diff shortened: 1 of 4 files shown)


Alkalem:
The getters (for example `get_length`) might have a reasonable default, but a bad default is likely worse than an exception or the current `NotImplemented` return value. Because I have no experience with `fileaccessor`, `filemetadata` and `interaction`, I chose the error as solution.

plafosse (approved):

plafosse:
Thanks for your contribution! Merged in 5.4.10152

PR #7980, opened 2026-02-23
Python bindings for defining relocations and handlers

### LINKED ISSUE #1470: Relocation bindings and documentation
Require Python bindings and documentation for how to implement relocation functionality for a custom binary view (for COFF files, in this specific case).

nullableVoidPtr:
Fixes #1470

### DIFF
--- a/python/relocation.py
+++ b/python/relocation.py
@@ -247,7 +247,7 @@ class RelocationHandler:
 
 		return bytes(dest)
 
-	def perform_get_operand_for_external_relocation(self, data: bytes, addr: int, il: lowlevelil.LowLevelILFunction, reloc: Relocation) -> bool | Optional['lowlevelil.ExpressionIndex']:
+	def perform_get_operand_for_external_relocation(self, data: bytes, addr: int, il: lowlevelil.LowLevelILFunction, reloc: Relocation) -> Optional[bool | int]:
 		return True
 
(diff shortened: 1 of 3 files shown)


nullableVoidPtr:
Need to force-rebase against actual upstream instead of my mistaken APX fork

nullableVoidPtr:
Insofar, reading relocations work fine - needs more testing re: `define_relocation`

python/relocation.py
```diff
@@ -0,0 +1,303 @@
...
+			arch.handle,
+			reloc.handle,
+			dest,
+			length
+		):
+			return None
+
+		return bytes(dest)
+
+	def get_operand_for_external_relocation(self, data: bytes, addr: int, il: lowlevelil.LowLevelILFunction, reloc: Relocation) -> bool | Optional['lowlevelil.ExpressionIndex']:
```
emesare: operand != expr index, just return a int for now

plafosse:
Can you provide a test case demonstrating this code working? I haven't fully digested the code but ApplyRelocation involves directly modifying bytes at a passed in pointer in C++ and when I originally wrote this code I didn't take the time to think through how to accomplish this in pythonic way.

nullableVoidPtr:
Apologies - I forgot to reply this. Really the "Pythonic" way would be ingesting a "src_bytes" and returning something like an optional "overwritten" bytes, with nil value for failure. I haven't written a test for it yet since it does require somewhat more extensive harnesses but I can get around to it this weekend.

PR #8002, opened 2026-03-09
Avoid creating functions at jump table entries in shared cache

### LINKED ISSUE #7992: Functions are being created for jump tables in code sections
- Binary Ninja Version: 5.3.9208-dev Ultimate (8b909a0c)
- Edition: Ultimate
- OS: macOS
- OS Version: 26.3
- CPU Architecture: M1

**Bug Description:**
It seems Binary Ninja is overwriting jump tables in code sections with functions even though it has correctly identified them as jump tables. The symbol of the function is a data symbol as well.

**Steps To Reproduce:**
1. Open a recent copy of the DYLD Shared Cache with default load settings so `libsystem_c.dylib` is loaded automatically.
2. Wait for initial analysis to complete.
3. Go to any of the unnamed functions (beginning with `sub_`).
4. Observe that they all seem to be for jump tables are not actual functions.

**Expected Behavior:**
Jump tables in code sections shouldn't be overwritten with functions.

Navigating to the symbol for one of the unnamed functions thats actually a jump table:

Image

Navigate to the reference to the jump table:

Image

It appears the jump table is working correctly as a switch case.

**Binary:**
DYLD Shared Cache for an iPhone 17 Pro Max running iOS 26.0

**Additional Information:**
I confirmed the behaviour is exactly the same when running Binary Ninja with plugins disabled.

plafosse:
## Summary

- When processing shared cache Mach-O function tables, Binary Ninja was creating functions at addresses that turn out to be jump table entries rather than real function starts, producing spurious `sub_` functions.
- Adds a `HeuristicIsAFunction` check before calling `AddFunctionForAnalysis`. On aarch64, if the first instruction at the candidate address disassembles as `udf`, the entry is skipped — `udf` is used as padding/data in jump tables and is never a valid function prologue.
- This heuristic is admittedly a bit rough and could produce false negatives in unusual cases, but it directly addresses the reported symptom.

Fix https://github.com/Vector35/binaryninja-api/issues/7992

### DIFF
--- a/view/macho/machoview.h
+++ b/view/macho/machoview.h
@@ -5,6 +5,7 @@
 #include <string.h>
 
 #include "binaryninjaapi.h"
+#include "lowlevelilinstruction.h"
 #include "objc.h"
 
 //These are laready defined in one of the osx headers we want to override
@@ -1545,4 +1546,27 @@ namespace BinaryNinja
 	};
 
 	void InitMachoViewType();
+}
+
+// Check whether the instruction at `addr` looks like a valid function entry point.
+// Returns false for undefined or trap instructions that typically indicate jump table entries.
+// `opcode` must contain at least `opLen` bytes read from `addr`.
+inline bool IsValidFunctionStart(BinaryNinja::Architecture* arch, uint64_t addr, const uint8_t* opcode, size_t opLen)
+{
+	if (!opLen)
+		return false;
+
+	BinaryNinja::Ref<BinaryNinja::LowLevelILFunction> ilFunc = new BinaryNinja::LowLevelILFunction(arch, nullptr);
+	ilFunc->SetCurrentAddress(arch, addr);
+	arch->GetInstructionLowLevelIL(opcode, addr, opLen, *ilFunc);
+	for (size_t i = 0; i < ilFunc->GetInstructionCount(); i++)
+	{
+		const auto& instr = ilFunc->GetInstruction(i);
+		if (instr.operation == LLIL_UNDEF)
+			return false;
+		if (i == 0 && instr.operation == LLIL_TRAP)
+			return false;
+	}
+
+	return true;
 }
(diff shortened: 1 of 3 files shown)


view/sharedcache/core/MachOProcessor.cpp
```diff
@@ -17,6 +17,30 @@ SharedCacheMachOProcessor::SharedCacheMachOProcessor(Ref<BinaryView> view, std::
 	}
 }
 
+static bool HeuristicIsAFunction(Platform* targetPlatform, std::shared_ptr<VirtualMemory> vm, uint64_t func)
```
bdash: It's probably best to use the same logic as `MachoView::IsValidFunctionStart` so we get consistent behavior. Ideally we'd share that code, but for whatever reason none of the Mach-O handling is shared at present.

plafosse: yeah that a good point I kind forgot that existed

view/macho/machoview.h
```diff
@@ -1549,3 +1547,26 @@ namespace BinaryNinja
 
 	void InitMachoViewType();
 }
+
+// Check whether the instruction at `addr` looks like a valid function entry point.
+// Returns false for undefined or trap instructions that typically indicate jump table entries.
+// `opcode` must contain at least `opLen` bytes read from `addr`.
+inline bool IsValidFunctionStart(BinaryNinja::Architecture* arch, uint64_t addr, const uint8_t* opcode, size_t opLen)
```
bdash, line 1554: This should be within the namespace.

It's rather unfortunate that this function is inlined in the header. Presumably that's because there is no shared target for Mach-O specific stuff, and the shared cache can't link to machoview.

view/sharedcache/core/MachOProcessor.cpp
```diff
@@ -50,9 +50,16 @@ void SharedCacheMachOProcessor::ApplyHeader(const SharedCache& cache, SharedCach
...
 		{
 			auto targetPlatform = m_view->GetDefaultPlatform();
+			auto arch = targetPlatform->GetArchitecture();
 			auto functions = header.ReadFunctionTable(*m_vm);
 			for (const auto& func : functions)
-				m_view->AddFunctionForAnalysis(targetPlatform, func, false);
+			{
+				uint8_t opcode[BN_MAX_INSTRUCTION_LENGTH];
+				size_t opLen = arch->GetMaxInstructionLength();
+				m_vm->Read(opcode, func, opLen);
```
bdash, line 59: We should handle the fact that `VirtualMemory::Read` can throw.

bdash (changes requested):

PR #8040, opened 2026-03-25
Add a hook to allow BinaryView subclasses to run code after snapshot data is applied

bdash:
Snapshot data is applied when loading from a database, rebasing the view, etc.

### DIFF
--- a/binaryninjaapi.h
+++ b/binaryninjaapi.h
@@ -5773,6 +5773,15 @@ namespace BinaryNinja {
 		void PerformDefineRelocation(Architecture* arch, BNRelocationInfo& info, uint64_t target, uint64_t reloc);
 		void PerformDefineRelocation(Architecture* arch, BNRelocationInfo& info, Ref<Symbol> sym, uint64_t reloc);
 
+		/*! OnAfterSnapshotDataApplied is called when loading a view from a database, after snapshot data has been applied to it.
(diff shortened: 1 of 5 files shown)


negasora:
What do you think about renaming this callback to `OnAfterSnapshotDataApplied`?
"Did apply" to me suggests that it'd return a bool returning whether or not snapshot data was applied.
Also most (all?) existing event-based callbacks use `OnAfter`/`OnBefore` prefixes iirc

bdash:
I don't feel strongly about the naming so I'm happy to adjust it if it makes sense to do so.

In the API code, the only uses of `OnBefore` / `OnAfter` are in `ProjectNotification`. All of the `On*` methods appear to also be on the `FooNotification` types. Do you think it makes sense to continue that naming pattern in this different context?

negasora:
I think it does because people might be familiar with one callback and search `OnBefore`/`OnAfter` to see what other ones are available. `On` also kind of gives the message of "this is a called when an event happens".
Also if we want to add a callback later for before snapshot data is applied (idk why/when this would happen) we can use `OnBeforeSnapshotDataApplied` instead of e.g. `AboutToApplySnapshotData`.

bdash:
Renamed to `OnAfterSnapshotDataApplied`.

PR #8048, opened 2026-03-31
[Thumb2] Fix IT instruction info

### LINKED ISSUE #4430: ninja thumb IT Conditional instruction Bring error analysis
- Binary Ninja Version: [e.g. 3.3.3996-dev] (if version is stable, please also test the latest development build via the "Update Channel" option)
- OS: [win10]
- CPU Architecture: [x64]

**Bug Description:**
IT Conditional append instruction goto instruction ninja does not recognize subsequent instructions, resulting in errors

**Steps To Reproduce:**
1. rebase 0xbf5d6000
2. goto 0xbf5e880a

**Screenshots:**
image
image
image
image

**Additional Information:**
[demo.zip](https://github.com/Vector35/binaryninja-api/files/11812023/demo.zip)

### LINKED ISSUE #8039: Incorrect thumb2 disassembly
- Binary Ninja Version: 5.3.9301-dev Personal (0855e811)
- Edition: Non-Commercial
- OS: macOS
- OS Version: Tahoe 26.3
- CPU Architecture: M1

**Bug Description:**
There seems to be a case where in a thumb2 function, instructions are being disassembled as armv7 instead.
It isn't fully clear to me what the conditions are for this, it seems like it may happen after an `it` block?

In this image, the instruction at address 6 is being disassembled as `stmlt`, which is incorrect:
Image

Disassembling that instruction alone as thumb2 gives the correct result, which is a `b` instruction:
Image

For some reason the instruction is being disassembled as armv7 rather than thumb2, even though the function is being created as thumb2.

**Steps To Reproduce:**
1. Create new binary view
2. Paste following instruction bytes: 04bf0020704705f0aeb8
3. Make thumb2 function at address 0
4. Observe incorrect disassembly at address 6 (`stmlt` instead of `b`)

**Expected Behavior:**
The instruction at address 6 should be disassembled as thumb2.
As an example, this is what the disassembly looks like in IDA.

Image

negasora:
Fixes https://github.com/Vector35/binaryninja-api/issues/8039 and https://github.com/Vector35/binaryninja-api/issues/4430

### DIFF
--- a/arch/armv7/thumb2_disasm/arch_thumb2.cpp
+++ b/arch/armv7/thumb2_disasm/arch_thumb2.cpp
@@ -187,6 +187,117 @@ public:
 
 		result.length = decomp.instrSize / 8;
 
+		if ((decomp.mnem == armv7::ARMV7_IT) && (decomp.fields[FIELD_mask] != 0))
+		{
+			// IT block: consume all instructions and handle contained branches
+			uint32_t offset = decomp.instrSize / 8;
+			uint32_t mask = decomp.fields[FIELD_mask];
+			uint32_t cond = decomp.fields[FIELD_firstcond];
+
+			size_t instrCount;
+			if (mask & 1)
+				instrCount = 4;
+			else if (mask & 2)
+				instrCount = 3;
+			else if (mask & 4)
+				instrCount = 2;
+			else
+				instrCount = 1;
+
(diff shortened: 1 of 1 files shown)


negasora:
Merged in `5.4.9442-dev`

Issue #4028, opened 2022-03-08
"push r16" not correctly lifted in x86 (32 bits)


themaks:
On x86 (32 bits mode), disassembling the opcodes 66 53 66 5B, binary ninja outputs the following disassembly:
```
00000000  6653               push    bx {var_4}
00000002  665b               pop     bx
```
which is correct; however, the lifted IL is not :
```
   0 @ 00000000  push(zx.d(bx))
   1 @ 00000002  bx = pop.w
```

Indeed, it should read instead :
```
   0 @ 00000000  push(bx)  // OperandSize == 16
   1 @ 00000002  bx = pop.w
```
The 0x66 prefix is correctly interpreted for the `pop` operation (since it operates on 16 bits), but not forthe `push` operation.

Binary Ninja Version: 3.0.3280-dev Personal, 8291c569
Platform: Windows 10 Version 2009

xusheng6:
I can confirm this bug. The related code is here: https://github.com/Vector35/arch-x86/blob/master/il.cpp#L2819-L2834. For `push bx` instruction, `stackAdjustment` is 4 and `opOneLen` is 2. This is unexpected, since `stackAdjustment` should really be 2. Seems like a xed bug

xusheng6:
I have removed the milestone for this issue because it depends on an upstream issue, so I cannot really do much for it

PR #8050, opened 2026-04-01
Fix stack adjustment for x86 pop r16

haileys:
Related to https://github.com/Vector35/binaryninja-api/issues/4028

`pop r16` is lifted incorrectly and adjusts the stack by the wrong size:
image

This PR changes the lifting to always pop the memory operand size and insert a `LLIL_LOW_PART` if the destination register size is different:

image

### DIFF
--- a/arch/x86/il.cpp
+++ b/arch/x86/il.cpp
@@ -3141,6 +3141,15 @@ bool GetLowLevelILForInstruction(Architecture* arch, const uint64_t addr, LowLev
 		break;
 
 	case XED_ICLASS_POP:
+	{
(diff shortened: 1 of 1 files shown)


fuzyll (approved):
Sorry for the delay on this.

I was a little confused when starting to review this because I didn't think x86-64 had an `r16`. I now understand that this is fixing lifting for things like segment registers where the destination is smaller than width of the stack pop. Not some weird general-purpose register thing. 😛 

I verified that stuff like `pop fs` and `pop gs` appear improved while stuff like `pop ax` and `pop word [rax]` seem unaffected and the code itself seems like a clean change. I also verified that the change to `push` appears to not change anything and is just a refactor.

Thanks for the PR!

PR #8061, opened 2026-04-05
[Rust] add the process step for the creation of a custom binary view

rbran:
This allows custom implementations of CustomBinaryView to call `self.*` function related to `BinaryView`. 

Currently calling something like `add_segment` inside the `init` function will cause a panic because the `context.state` had being initialized with `CustomViewContextState::Initialized` yet, eg minidump does it:
https://github.com/Vector35/binaryninja-api/blob/e1efc289bfccf86afa8a4bda02ded1674ba018d0/view/minidump/src/view.rs#L261-L265)


emesare:
Closing in favor of https://github.com/Vector35/binaryninja-api/pull/8120, thank you for the PR!

Issue #7974, opened 2026-02-19
Instant crash during stage 1 analysis when a native architecture hook is registered


WeiN76LQh:
**Version and Platform (required):**
- Binary Ninja Version: 5.3.9138-dev Ultimate (546385d4)
- Edition: Ultimate
- OS: macOS
- OS Version: 26.3
- CPU Architecture: M1

**Bug Description:**
A recent change seems to have broken native `ArchitectureHook` plugins. By registering a completely barebones architecture hook that doesn't replace any functionality, Binary Ninja crashes immediately when the first stage analysis begins.

**Steps To Reproduce:**
Please provide all steps required to reproduce the behavior:
1. Compile the native plugin provided below
2. Open any arm64 binary in Binary Ninja
3. Observe immediate crash once analysis begins

**Binary:**
```c++
#include "binaryninjaapi.h"

using namespace BinaryNinja;
class arm64ArchitectureExtension : public ArchitectureHook {
public:
    arm64ArchitectureExtension()
        : ArchitectureHook(Architecture::GetByName("aarch64"))
    {
    }
};

extern "C" {
BN_DECLARE_CORE_ABI_VERSION

BINARYNINJAPLUGIN void CorePluginDependencies()
{
    AddRequiredPluginDependency("arch_arm64");
}

BINARYNINJAPLUGIN bool CorePluginInit()
{
    Architecture* arm64ext = new arm64ArchitectureExtension();
    Architecture::Register(arm64ext);
    return true;
}
}
```

emesare:
IIRC this crash was due to some basic block being freed twice, this crash was introduced as apart of 5.3.9129.

Still yet to determine what about that commit caused this to occur in an architecture hook.

zznop:
Fixed in `5.3.9423`

PR #8064, opened 2026-04-06
Fixes crash with native architecture hooks

zznop:
Broke into two separate commits
* Fix for https://github.com/Vector35/binaryninja-api/issues/7974
* Fix for spammed Python exception when a Python architecture hook is used

### DIFF
--- a/architecture.cpp
+++ b/architecture.cpp
@@ -483,7 +483,7 @@ FunctionLifterContext::FunctionLifterContext(LowLevelILFunction* func, BNFunctio
 	m_blocks.reserve(context->basicBlockCount);
 	for (size_t i = 0; i < context->basicBlockCount; i++)
 	{
-		m_blocks.emplace_back(new BasicBlock(context->basicBlocks[i]));
+		m_blocks.emplace_back(new BasicBlock(BNNewBasicBlockReference(context->basicBlocks[i])));
 	}
 
 	for (size_t i = 0; i < context->noReturnCallsCount; i++)
--- a/binaryninjacore.h
+++ b/binaryninjacore.h
@@ -37,14 +37,14 @@
 // Current ABI version for linking to the core. This is incremented any time
 // there are changes to the API that affect linking, including new functions,
 // new types, or modifications to existing functions or types.
-#define BN_CURRENT_CORE_ABI_VERSION 163
+#define BN_CURRENT_CORE_ABI_VERSION 164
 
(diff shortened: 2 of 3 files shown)


plafosse (approved):
Looks like this is changed to match ownership patterns similar to elsewhere in the code.

architecture.cpp
```diff
@@ -327,7 +327,10 @@ Ref<BasicBlock> BasicBlockAnalysisContext::CreateBasicBlock(Architecture* arch,
 	BNBasicBlock* block = BNAnalyzeBasicBlocksContextCreateBasicBlock(m_context, arch->GetObject(), start);
 	if (!block)
 		return nullptr;
-	return new BasicBlock(block);
+
+	Ref<BasicBlock> result = new BasicBlock(BNNewBasicBlockReference(block));
+	BNFreeBasicBlock(block);
```
bdash: Why are the calls to `BNNewBasicBlockReference` / `BNFreeBasicBlock` necessary here?

 `BNAnalyzeBasicBlocksContextCreateBasicBlock` returns an object with a +1 reference count. The `Ref<BasicBlock>` that is created in the return statement adopts that reference and will release it when it is destroyed.

zznop: Good catch. It is unnecessary and deadweight. The real fix for the crash is the change on line 486. Updating the PR now.

zznop: Well after testing with the Arch hook after changing those lines back to `return new BasicBlock(block)` it results in a crash. So it appears to be needed, investigating why now...

zznop: Ok my understanding is that BNAnalyzeBasicBlocksContextCreateBasicBlock returns a borrowed reference (not a +1 owned one). The core owns the block. The BNNewBasicBlockReference call is needed to create the APIs owned reference and BNFreeBasicBlock releases the borrowed reference.

In the non-architecutre hook case this doesn't cause a crash because the context and local ref's have the same lifetime. But when using an arch hook, the hook calls the base arch through BNArchitectureAnalyzeBasicBlocks. The base arch creates the wrappers, then dies (calling BNFreeBasicBlock on the core-owned blocks). Then the arch hook continues using those blocks resulting in the double-free.

So the pattern appears to be correct. We acquire our owned +1 ref from the borrowed pointer, the API takes ownership of that reference, and BNFreeBasicBlock releases the borrowed pointer.

Issue #8051, opened 2026-04-01
[DSC] Database reanalyses some Obj-C methods every time it is opened


VisualEhrmanntraut:
Binary Ninja Version: 5.3.9392-dev

This is because of the fix to #8030.

bdash:
Retitled to reflect that this is not a full analysis, but an incremental one of functions that reference some Obj-C types and selectors.

PR #8065, opened 2026-04-06
[ObjC] Don't undefine / redefine symbols and types that already exist in the view

bdash:
This can cause functions to be reanalyzed unnecessarily when the view is loaded from a bndb. Fixes
https://github.com/Vector35/binaryninja-api/issues/8051.

### DIFF
--- a/objectivec/objc.cpp
+++ b/objectivec/objc.cpp
@@ -412,10 +412,16 @@ void ObjCProcessor::DefineObjCSymbol(
 
 	auto defineSymbol = [this](Ref<Symbol> symbol, const Confidence<Ref<Type>>& type) {
 		uint64_t symbolAddress = symbol->GetAddress();
+		if (Ref<Symbol> existingSymbol = m_data->GetSymbolByAddress(symbolAddress))
+		{
+			if (existingSymbol->IsAutoDefined() && existingSymbol->GetType() == symbol->GetType() && existingSymbol->GetRawNameRef() == symbol->GetRawNameRef())
+				return;
+
+			m_data->UndefineAutoSymbol(existingSymbol);
+		}
+
 		// Armv7/Thumb: This will rewrite the symbol's address.
 		// e.g. We pass in 0xc001, it will rewrite it to 0xc000 and create the function w/ the "thumb2" arch.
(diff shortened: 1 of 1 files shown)


VisualEhrmanntraut:
wouldn't it be better if this was done at the core level, globally, for all changes

bdash:
I'm not sure what "this" is you're referring to.

VisualEhrmanntraut:
not triggering analysis when there was no difference made from the applied change (e.g. previous variable/symbol/function type is the same as the new)

VisualEhrmanntraut:
not saying that this won't work for now. just saying as a future change.

bdash:
I'm not making a change that alters the behavior of existing APIs at this point in our release cycle.

VisualEhrmanntraut:
I said future. That implies another release cycle. I know the current release cycle ends at 8th April.

objectivec/objc.cpp
```diff
@@ -412,10 +412,16 @@ void ObjCProcessor::DefineObjCSymbol(
 
 	auto defineSymbol = [this](Ref<Symbol> symbol, const Confidence<Ref<Type>>& type) {
 		uint64_t symbolAddress = symbol->GetAddress();
-		// Armv7/Thumb: This will rewrite the symbol's address.
-		// e.g. We pass in 0xc001, it will rewrite it to 0xc000 and create the function w/ the "thumb2" arch.
 		if (Ref<Symbol> existingSymbol = m_data->GetSymbolByAddress(symbolAddress))
+		{
+			if (existingSymbol->IsAutoDefined() && existingSymbol->GetType() == symbol->GetType() && existingSymbol->GetFullNameRef() == symbol->GetFullNameRef())
```
plafosse: is there a reason to use FullName instead of RawName here? I'm not sure about the exact context but RawName should be invariant to the binary were I can see a situation where the demangler changes and the FullName stops matching.... but maybe thats your goal.

bdash: Raw name is more correct as the code stands.

plafosse (approved):
Looks like a reasonable targeted change while we're in code freeze. We should probably investigate if its worth putting these checks into the core after release.

PR #8069, opened 2026-04-08
add missing container trainsforms to the docs

psifertex:
adding missing transform types to docs


psifertex:
Peter already approved. Build (mostly) succeeded. Intermittent network failures unrelated to this.

PR #8082, opened 2026-04-10
[RTTI] Fix Itanium RTTI scan failing to symbolize objects with extern typeinfo

### LINKED ISSUE #8080: Itanium RTTI metadata not symbolized when base type info is extern
Regression from 5.2 that affects VMI class type that has the base type info in another binary (e.g. extern)

1. Open `loss publish wander`
2. Go to `0x100620e58`
3. Observe that there is no symbol named `_typeinfo_for_DetailsTab`

Was caused by new GNU3 demangler changing the naming scheme for the extern symbol:

Old: `_typeinfo_for_QTabWidget`
New `typeinfo_for_QTabWidget`

So just need to update this code here:

https://github.com/Vector35/binaryninja-api/blob/24eabc032f6450001bd799b5ec2dc368467b773a/plugins/rtti/itanium.cpp#L448-L451

emesare:
The issue stems from the gnu3 demangler changing in https://github.com/Vector35/binaryninja-api/commit/260ca61d94134b6743807e29f64b5ce4f6918d73

Fixes https://github.com/Vector35/binaryninja-api/issues/8080

### DIFF
--- a/plugins/rtti/itanium.cpp
+++ b/plugins/rtti/itanium.cpp
@@ -446,9 +446,11 @@ std::optional<ClassInfo> ItaniumRTTIProcessor::ProcessRTTI(uint64_t objectAddr)
             return std::nullopt;
         auto symName = sym->GetShortName();
         // Remove type info prefix.
-        if (symName.rfind("_typeinfo_for_", 0) != 0)
-            return std::nullopt;
-        return symName.substr(14);
+        if (symName.rfind("_typeinfo_for_", 0) == 0)
+            return symName.substr(14);
+        if (symName.rfind("typeinfo_for_", 0) == 0)
+            return symName.substr(13);
+        return std::nullopt;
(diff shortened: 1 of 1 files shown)


plafosse:
Oh dang didn't realize we were doing any string matching on the output of the demangler.

plafosse (approved):

emesare:
Yea its for extern stuff, maybe there is some metadata in the ELF/Mach-O we can use instead of just doing that, but 🤷

plafosse:
I feel like we need to be using the same string or we need a static assert to ensure stuff like this doesn't happen. Also I think this might fail if our own demangler is turned off and the Generic Demangler kicks in.

emesare:
I think I will do a follow up PR to make this specific part more robust, for now this is good to fix the regression for 5.3.

PR #8100, opened 2026-04-16
[Python] Fix handling of zero-width types returned by a custom Demangler

### LINKED ISSUE #8099: Type returned by custom demangler written in Python is not applied
- Binary Ninja Version: 5.4.9471-dev Ultimate, c9252859
- OS: macos
- OS Version: 26.4
- CPU Architecture: arm64

**Bug Description:**
A [user on Slack reported](https://binaryninja.slack.com/archives/C1E616ZSQ/p1776305807211109?thread_ts=1776288787.671849&cid=C1E616ZSQ) that even though they're returning a type from their custom demangler, it is not being applied to the symbol in question. This works correctly for demanglers written in C++ (our GNU3 / MSVC demanglers) and Rust (our Swift demangler).

**Steps to reproduce:**
I was able to reproduce with this unit test:
```python
def test_custom_demangler_applies_type(self):
	"""Test that a custom Demangler subclass correctly returns both type and name via demangle_generic."""
	mangled_prefix = "_CUSTOMTEST_"

	class TypeTestDemangler(Demangler):
		name = "TypeTestDemangler"

		def is_mangled_string(self, name: str) -> bool:
			return name.startswith(mangled_prefix)

		def demangle(self, arch, name, view=None, simplify=False):
			if not name.startswith(mangled_prefix):
				return None
			demangled_name = name[len(mangled_prefix):]
			func_type = Type.function(Type.int(4), [
				FunctionParameter(Type.pointer(arch, Type.char()), "arg1"),
			])
			return (func_type, QualifiedName(demangled_name))

	TypeTestDemangler().register()
	arch = Architecture["x86_64"]

	result = demangle_generic(arch, mangled_prefix + "my_func")
	assert result is not None, "demangle_generic returned None"
	demangled_type, demangled_name = result

	assert demangled_name == ["my_func"], f"Expected ['my_func'], got {demangled_name}"
	assert demangled_type is not None, "Custom demangler type was not returned by demangle_generic"
	assert demangled_type.type_class == bn.TypeClass.FunctionTypeClass, \
		f"Expected FunctionTypeClass, got {demangled_type.type_class}"
```

This gives:
```
FAILED AssertionError: Custom demangler type was not returned by demangle_generic
```

bdash:
Explicitly test the type against `None` rather than testing for truthiness. This ensures that zero-width types returned by a custom demangler will be applied.

Fixes #8099.

---

For some reason `Type` implements `__len__` and returns `self.width`. This causes zero-width types, such as function types, to be considered False. This is somewhere on the spectrum of unnecessarily confusing and just plain wrong.

### DIFF
--- a/python/demangle.py
+++ b/python/demangle.py
@@ -408,7 +408,7 @@ class Demangler(metaclass=_DemanglerMetaclass):
 				var_name = types.QualifiedName(var_name)
 
 			Demangler._cached_name = var_name._to_core_struct()
-			if type:
+			if type is not None:
 				out_type[0] = core.BNNewTypeReference(type.handle)
 			else:
 				out_type[0] = None


CouleeApps (approved):
... I'm surprised python makes the leap from truthiness -> non zero length. What ever happened to "Explicit is better than implicit."

The bugfix change itself is good but the test you wrote (in the private repo) might be better placed in test_demangler.py and likely doesn't need all the extra scaffolding.

bdash:
I'll rework my test. I somehow entirely overlooked the existence of that file 😅 

Python has treated empty lists as false since very early versions, and they expanded that behavior to user defined types via `__len__`. I think we could implement `__bool__` on `Type` to avoid this, but I'm not entirely sure why it implements `__len__`. It's not a sequence or a mapping as Python sees it. `len(some_type)` evaluating to its width seems kinda surprising, and I'm not sure why anyone would reach for that vs `some_type.width`. It may be too late to undo that now, though.

CouleeApps:
I don't think we should make any changes to `Type` at this time, but it would be worth looking around the codebase to see if anywhere else accidentally (or intentionally?) makes use of this behavior

PR #8104, opened 2026-04-16
Make snippets submodule load over HTTPS instead of SSH

CouleeApps:
Loading over SSH breaks GitHub Actions, which does not have an SSH key configured


fuzyll (approved):
This is the change I had suggested in Slack. URL looks correct, no other changes. Approved.

PR #8105, opened 2026-04-17
Fix aarch64 disassembler Makefiles

gdbinit:
Both Makefile have the HEADERS variable with wrong filename for the updated sysregs* files from last commit on this file.

Make was skipping the configured target and using the default because of this.


bdash:
Thanks for the PR! Rebased and merged in f93512c43333.

PR #8112, opened 2026-04-20
[WARP] Fix misc UI thread safety issues

### LINKED ISSUE #8111: Crash due to null dereference in lambda within WarpCurrentFunctionWidget::SetCurrentFunction
Sentry Issue: [BINARYNINJA-4R](https://vector35.sentry.io/issues/7425787872/)

```
EXCEPTION_ACCESS_VIOLATION_READ / 0x10: Fatal Error: EXCEPTION_ACCESS_VIOLATION_READ / 0x10
  File "function.cpp", line 178, in BinaryNinja::Function::GetView
  File "matches.cpp", line 166, in `WarpCurrentFunctionWidget::SetCurrentFunction'::`10'::<T>::operator()
```

@bdash says:
> The [lambda within `SetCurrentFunction` captures `this` and accesses `m_current`](https://github.com/Vector35/binaryninja-api/blob/bfd306437f34ebc4d4170c5fb8170261585f6038/plugins/warp/ui/matches.cpp#L163-L170). There does not appear to be any locking involved, despite the fact this field is accessed from multiple threads. Additionally, nothing prevents a subsequent call to `SetCurrentFunction` from setting `m_current` to nullptr while an instance of this lambda is enqueued. If this happens then the call to `Function::GetView` will crash due to a null `this.`

emesare:
1. Not holding mutex when clearing fetchers processed guids
2. Not managing the lifetime of the fetchers background thread correctly on current function navigation

Fixes https://github.com/Vector35/binaryninja-api/issues/8111

### DIFF
--- a/plugins/warp/ui/matches.cpp
+++ b/plugins/warp/ui/matches.cpp
@@ -161,18 +161,29 @@ void WarpCurrentFunctionWidget::SetCurrentFunction(FunctionRef current)
 			// Don't block the UI thread when fetching, also unlike other cases where we reach for QtConcurrent::run we
 			// do not use a watcher to tie the future to the widget's lifetime, this is because network requests can
 			// take a long time, and we want to avoid blocking the UI thread in the widget destructor. Instead of
-			// insuring the widget is alive, we just use a weak pointer that can tell us when it (self) has been
+			// ensuring the widget is alive, we just use a weak pointer that can tell us when it (self) has been
 			// destructed.
 			auto future = QtConcurrent::run([self = QPointer(this), current, fetcher = m_fetcher]() {
 				if (!self)
 					return;
-				QMetaObject::invokeMethod(self, [self] { self->m_spinner->show(); }, Qt::QueuedConnection);
+				QMetaObject::invokeMethod(
+					self,
+					[self] {
+						if (self)
+							self->m_spinner->show();
+					},
+					Qt::QueuedConnection);
 				BinaryNinja::Ref bgTask = new BinaryNinja::BackgroundTask("Fetching WARP Functions...", true);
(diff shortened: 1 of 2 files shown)


plugins/warp/ui/matches.cpp
```diff
@@ -151,22 +152,27 @@ void WarpCurrentFunctionWidget::SetCurrentFunction(FunctionRef current)
...
 		if (!m_fetcher->m_requestInProgress.exchange(true))
 		{
-			BinaryNinja::WorkerPriorityEnqueue([this]() {
-				QMetaObject::invokeMethod(this, [this] { m_spinner->show(); }, Qt::QueuedConnection);
+			// Don't block the UI thread when fetching, also unlike other cases where we reach for QtConcurrent::run we
+			// do not use a watcher to tie the future to the widget's lifetime, this is because network requests can
+			// take a long time, and we want to avoid blocking the UI thread in the widget destructor. Instead of
+			// insuring the widget is alive, we just use a weak pointer that can tell us when it (self) has been
+			// destructed.
+			auto future = QtConcurrent::run([self = QPointer(this), current, fetcher = m_fetcher]() {
```
bdash, line 166: I'm not super familiar with `QPointer` but I'm not sure the way that it used here is sufficient. AFAIK nothing prevents the pointed-to object from being destroyed concurrently with the execution of the body of this lambda, after the `QPointer` has been checked. For this to be safe I think the checks of `self` would need to be pushed down into the bodies of the lambdas passed to `QMetaObject::invokeMethod`.

emesare: Added guards in https://github.com/Vector35/binaryninja-api/commit/9e849636fbf7af6498d66851d94171da2adf82ea

bdash (approved):

PR #8120, opened 2026-04-23
[Rust] Misc `binary_view` module refactoring

emesare:
- Remove the "viral" `BinaryViewExt` trait and its blanket impl
- Split up the binary view type from the custom trait impl
- Simplify and fix bugs regarding custom binary view initialization
- Rewrite Minidump binary view example, parses the PE headers to create proper sections now
- Add some extra documentation
- Add unit test for custom binary view

This is in preparation for bundling some rust view types such as the aforementioned minidump one, and potential replacements for other existing view types.

The removal of `BinaryViewExt` also fixes some cases where LSPs will trip up do to its blanket impl and suggest functions from it in completely wrong scenarios, although that itself was not the reasoning behind the change, its more so to do with the fact those functions are invalid in the case of a `CustomBinaryView` that has yet to have its core object constructed.

Supersedes https://github.com/Vector35/binaryninja-api/pull/8061


rust/src/binary_view.rs
```diff
@@ -72,15 +74,377 @@ pub use memory_map::{MemoryMap, MemoryRegionInfo, ResolvedRange};
...
+unsafe impl Send for BinaryViewType {}
+unsafe impl Sync for BinaryViewType {}
+
+/// Implemented for custom views, responsible for setting up the view state once the binary is open.
+pub trait CustomBinaryView: BinaryViewBase {
+    /// Initializes the opened binary view state.
+    ///
+    /// Use this to populate the [`BinaryView`] with sections, segments, and other view data.
+    ///
+    /// NOTE: You must add **at-least** one segment to the view, otherwise calls to [`BinaryViewType::create`]
```
bdash: `at least` is two words

rust/src/binary_view.rs
```diff
@@ -241,41 +603,221 @@ impl From<BNAnalysisProgress> for AnalysisProgress {
...
+                    getDefaultEndianness: Some(cb_endianness::<C>),
+                    isRelocatable: Some(cb_relocatable::<C>),
+                    getAddressSize: Some(cb_address_size::<C>),
+                    save: Some(cb_save::<C>),
+                    onAfterSnapshotDataApplied: Some(cb_on_after_snapshot_data_applied::<C>),
+                },
+            )
+        };
+        if handle.is_null() {
+            // TODO: Free here?
```
bdash: Seems like the leaked box does need to be freed here?

rust/src/binary_view.rs
```diff
@@ -241,41 +603,221 @@ impl From<BNAnalysisProgress> for AnalysisProgress {
...
+    }
+
+    pub(crate) unsafe fn ref_from_raw(handle: *mut BNBinaryView) -> Ref<Self> {
+        debug_assert!(!handle.is_null());
+        Ref::new(Self { handle })
+    }
+
+    /// Create a core instance of the [`CustomBinaryView`].
+    pub fn from_custom<C: CustomBinaryView>(
+        view_type_name: &str,
```
bdash, line 650: Why does this need the view type name to be passed in explicitly? Can it not be retrieved from the view?

rust/src/binary_view.rs
```diff
@@ -1524,95 +2061,89 @@ pub trait BinaryViewExt: BinaryViewBase {
...
         let settings_handle =
-            unsafe { BNBinaryViewGetLoadSettings(self.as_ref().handle, view_type_name.as_ptr()) };
-
-        if settings_handle.is_null() {
-            Err(())
-        } else {
-            Ok(unsafe { Settings::ref_from_raw(settings_handle) })
+            unsafe { BNBinaryViewGetLoadSettings(self.handle, view_type_name.as_ptr()) };
+        match settings_handle.is_null() {
+            true => Settings::new(),
```
bdash: Why is this changing to return a new `Settings` instance rather than returning an error or None? The equivalent C++ API looks to have a nullable return type.

bdash (approved):

Issue #7466, opened 2025-10-06
Windows Visual C++ name demangler fails in some cases


michael-zendaro:
**Version and Platform (required):**
- Binary Ninja Version: [e.g. 5.2.8353-dev]
- Edition: Non-Commercial 
- OS: Windows
- OS Version: 11
- CPU Architecture: x64

**Bug Description:**
In some cases demangler failed to demangle function name. So, as well it failed to set function signature.
I can give this case, other is in dump:
```
?Present1@?QIDXGISwapChain4@@CDXGISwapChain@@UEAAJIIPEBUDXGI_PRESENT_PARAMETERS@@@Z
```

**Steps To Reproduce:**
Please provide all steps required to reproduce the behavior:
1. Search for `@@@` in symbols

**Screenshots/Video Recording:**
<img width="1414" height="484" alt="Image" src="https://github.com/user-attachments/assets/c7e9351b-a507-4b72-892b-6b8aed271f1e" />

**Binary:**
Via Binaryninja portal: system drop burns superbly

plafosse:
Fixed in 5.4.9912

Issue #3535, opened 2022-10-13
Support for demangled variable names


op2786:
Let's say I want to rename a varible to this: `??_7Manager@mgr@app@@6B@`. Currently when I do this, I see variable is named like this:

![vname](https://user-images.githubusercontent.com/109607896/195590136-56ac333f-0592-43ec-9982-898c34b9240b.PNG)

Instead of setting variable name into mangled name, I expect it would be converted into something like this: app::mgr::Manager::`vftable'

Supporting demangled variable names would be useful.

plafosse:
Fixed in 5.4.9912

Issue #1653, opened 2020-05-09
demangle_ms bugs/improvements


0x1F9F1:
Binary Ninja Version: 2.0.2138-dev, d031c340
Platform: Windows 10 Version 1903

After using demangle_ms a lot lately, here's a list of problems I have come across:

## Vftables have type_class TypeClass.NamedTypeReferenceClass
```
# const Foo::`vftable'
>>> demangle_ms(arch, '??_7Foo@@6B@')
(<type: const>, ['Foo', "`vftable'"])

>>> demangle_ms(arch, '??_7Foo@@6B@')[0].type_class
<TypeClass.NamedTypeReferenceClass: 11>
```

## Multiple inheritance vftables produce invalid types and contain the parent in the type tokens instead of the name
```
# const Bar::`vftable'{for `Foo'}
>>> demangle_ms(arch, '??_7Bar@@6BFoo@@@') 
Traceback (most recent call last):
  File "C:\Program Files\Vector35\BinaryNinja\plugins\..\python\binaryninja\types.py", line 356, in __repr__
    return "<type: %s>" % str(self)
  File "C:\Program Files\Vector35\BinaryNinja\plugins\..\python\binaryninja\types.py", line 367, in __str__
    return core.BNGetTypeString(self._handle, platform)
  File "C:\Program Files\Vector35\BinaryNinja\plugins\..\python\binaryninja\_binaryninjacore.py", line 11021, in BNGetTypeString
    result = _BNGetTypeString(*args)
OSError: exception: access violation reading 0x0000000000000010

>>> demangle_ms(arch, '??_7Bar@@6BFoo@@@')[0].get_tokens_after_name()
['{', 'for', ' `', 'Foo', "'}"]
```

## Certain symbols contain the attributes in their type
Attributes (static, virtual, access modifier, etc.) could really do with being returned separately instead of discarding them or making them part of the type string.
```
# protected: static bool Foo::Bar
>>> demangle_ms(arch, '?Bar@Foo@@1_NA')
(<type: protected: static bool>, ['Foo', 'Bar'])

>>> demangle_ms(arch, '?Bar@Foo@@1_NA')[0].get_tokens_before_name()
['protected:', ' ', 'static', ' ', 'bool']
```

## Functions with no parameters have unnecessary void parameter
```
# void __cdecl Foo(void)
>>> demangle_ms(arch, '?Foo@@YAXXZ')
(<type: void __cdecl (void)>, ['Foo'])

[...]

Issue #1605, opened 2020-04-22
Large number of symbols in some Microsoft x86_64 PE files fail to get demangled


e1guapo:
On some Windows binaries, there are a number of symbols that do not get properly demangled.

One example I see is "C:\Windows\ShellExperiences\ClockFlyoutExperience.dll."
The sample from Windows 10 (Version 10.0.18363) has 20,252 total symbols in `bv.symbols`. Of these, 7,796 are still mangled. Microsoft's `undname.exe` utility successfully demangles 6,394 of these.

Some example symbols:
```
?set@IsAttachable@XamlMember@InfoProvider@XamlTypeInfo@@UE$AAAX_N@Z

?get@?QIXamlType@Markup@Xaml@UI@Windows@@IsMarkupExtension@XamlUserType@InfoProvider@XamlTypeInfo@@UE$AAA_NXZ

?GetBindingConnector@?QIComponentConnector2@Markup@Xaml@UI@Windows@@MainPage@ClockFlyoutExperience@@UE$AAAPE$AAUIComponentConnector@2345@HPE$AAVObject@Platform@@@Z
```

marpie:
The same seems to be true for any version of the `subst.exe`:

![image](https://github.com/Vector35/binaryninja-api/assets/703785/6fa4de4e-5248-4771-97b8-09537903369a)

![image](https://github.com/Vector35/binaryninja-api/assets/703785/e4b19062-a8d1-4b71-82ec-963e5bcee20e)

plafosse:
Fixed in 5.4.9912

PR #8140, opened 2026-04-30
Add MSVC demangler correctness fixes and generated test corpus

plafosse:
This should be considered a ground up re-write and so you probably won't get any value out of looking at diffs of the actual demangler. It pretty much solves all known demangler accuracy issues. And lays the ground work for more performance unlocks when we get the new simplifier integrated.                                                                              

Major changes:
- Demangling now is backed by the DemangledTypeNode like gnu3 is. The purpose is
to provide an abstraction layer between c++ features and binary
ninja's type objects.
- Substantial performance increase I think around 3x what it was before
this is due to cutting down on extraneous string copies and type object allocations
- Substantial accuracy fixes. The most substantial of which is accurate scoping
 of back references.
- Lots and lots of other small fixes. This commit could have been about 100 commits but
I didn't feel that adding all those commits would be helpful in understanding what's
actually going on here.
- The msvc code roughly twice as fast even while bailing out way less than the previous version. 

Fixes:
 - https://github.com/Vector35/binaryninja-api/issues/7466
 - https://github.com/Vector35/binaryninja-api/issues/3535
 - https://github.com/Vector35/binaryninja-api/issues/1653
 - https://github.com/Vector35/binaryninja-api/issues/1605
 - https://github.com/Vector35/binaryninja-api/issues/1546


plafosse:
I realized that bb9e72008d010289b51a43711670926809535db5 and 7bfbb87a098b686f415b7d1f949f813ca17aafab have regressed performance. And this needs some more changes so I'm converting it back to a draft :\

demangler/msvc/demangle_msvc.h
```diff
@@ -25,47 +26,33 @@
...
 #else
 #include "binaryninjaapi.h"
 #define BN BinaryNinja
 #define _STD_STRING std::string
 #define _STD_VECTOR std::vector
-#define _STD_SET std::set
+#endif
+
+#ifdef BINARYNINJACORE_LIBRARY
+#include "demangler/gnu3/demangled_type_node.h"
```
CouleeApps: The MS demangler plugin shouldn't be importing these headers from the GNU plugin, but instead these headers should live in a common location such that both plugins (and any future/user plugins) can load them. These headers and classes should probably be moved to api/base or some other central location in the repo

demangler/gnu3/demangle_gnu3.cpp
```diff
@@ -23,14 +23,23 @@
...
 #ifdef BINARYNINJACORE_LIBRARY
 using namespace BinaryNinjaCore;
-#define GetClass GetTypeClass
 #else
 using namespace BinaryNinja;
 using namespace std;
 #endif
 
 
-#define MAX_DEMANGLE_LENGTH    262144
```
CouleeApps, line 33: Any reason to remove the max length guard? I know the max nesting depth guard was added but that doesn't seem like it replaces this, and maximum length is a useful guardrail to have.

plafosse: We would need to stringify just to check the length this was a huge source of slow down in the old version

demangler/msvc/demangle_msvc.h
plafosse: Agreed we didn't catch that

demangler/gnu3/demangle_gnu3.cpp
CouleeApps: Good point. It may still be worth having a length size check on individual nodes though. I'm not sure if the mangle format lets you specify token length in a way that would cause problems, but I wouldn't put it past it.

plafosse: Yeah, it might be worth it having a maximum node length

plugins/rtti/rtti.cpp
```diff
@@ -24,9 +38,9 @@ std::optional<std::string> RTTI::DemangleNameMS(BinaryView* view, bool allowMang
 {
     QualifiedName demangledName = {};
     Ref<Type> outType = {};
-    if (!DemangleMS(view->GetDefaultArchitecture(), mangledName, outType, demangledName, true))
+    if (!DemangleMS(view->GetDefaultArchitecture(), mangledName, outType, demangledName, view))
```
CouleeApps, line 41: This is probably right, but historically trying to pass a pointer to this function has caused bugs with implicit type conversion to bool. CLion seems to think this is correct, though.

plafosse:
Also fixes https://github.com/Vector35/binaryninja-api/issues/8230 as we switched to a nesting guard rather than stringification max length

plafosse:
Merged in 5.4.9912

PR #8143, opened 2026-05-01
Fix Thumb platform preservation for ELF mini debug symbols

plafosse:
This fixes ARM/Thumb .gnu_debugdata loading by preserving the original mini-debug symbol address before ELF symbol normalization clears the Thumb bit.

For ARM/Thumb binaries, the parent ELF load passes an internal load option to the embedded mini-debug ELF load. The child ELF records metadata only for function symbols whose normalized address differs from the original symbol address. The parent then reads that metadata from the returned mini-debug BinaryView and uses the original address to recover the associated platform, while still defining the symbol at the normalized address.

This keeps the extra metadata work limited to ARM/Thumb binaries with .gnu_debugdata, avoids re-parsing the embedded ELF, and does not expose the temporary metadata on the final parent view.

### DIFF
--- a/view/elf/elfview.cpp
+++ b/view/elf/elfview.cpp
@@ -13,6 +13,14 @@ using namespace std;
 
 
 static ElfViewType* g_elfViewType = nullptr;
+static const char* ELF_RECORD_MINI_DEBUG_SYMBOL_ADDRESSES_OPTION = "loader.elf.recordMiniDebugSymbolAddresses";
+static const char* ELF_MINI_DEBUG_SYMBOL_ADDRESS_METADATA = "ELFMiniDebugSymbolAddresses";
+
+
+static bool IsFunctionSymbolType(BNSymbolType type)
+{
+	return (type == FunctionSymbol) || (type == ImportedFunctionSymbol) || (type == LibraryFunctionSymbol);
+}
 
 
 void BinaryNinja::InitElfViewType()
@@ -809,6 +817,9 @@ bool ElfView::Init()
 	SetDefaultPlatform(platform);
 	GetParentView()->SetDefaultPlatform(platform);
 
+	if (settings && settings->Contains(ELF_RECORD_MINI_DEBUG_SYMBOL_ADDRESSES_OPTION))
+		m_recordMiniDebugSymbolAddresses = settings->Get<bool>(ELF_RECORD_MINI_DEBUG_SYMBOL_ADDRESSES_OPTION, this);
+
 	// Finished for parse only mode
 	if (m_parseOnly)
 	{
@@ -1436,6 +1447,8 @@ bool ElfView::Init()
 
 	// Process the queued symbols
 	m_symbolQueue->Process();
+	if (!m_miniDebugSymbolAddresses.empty())
+		StoreMetadata(ELF_MINI_DEBUG_SYMBOL_ADDRESS_METADATA, new Metadata(m_miniDebugSymbolAddresses), true);
 	delete m_symbolQueue;
 	m_symbolQueue = nullptr;
 
@@ -2526,12 +2539,36 @@ bool ElfView::Init()
 }
 
 
+Ref<Symbol> ElfView::DefineElfSymbolAndVariableOrFunction(
+	Ref<Platform> functionPlatform, Ref<Symbol> symbol, const Confidence<Ref<Type>>& type)
+{
+	if (IsFunctionSymbolType(symbol->GetType()) && functionPlatform)
+	{
+		DefineAutoSymbol(symbol);
+		auto func = AddFunctionForAnalysis(functionPlatform, symbol->GetAddress());
+		if (type.GetValue() && func)
+			func->ApplyAutoDiscoveredType(type.GetValue());
+		return symbol;
+	}
+
+	return DefineAutoSymbolAndVariableOrFunction(GetDefaultPlatform(), symbol, type);
+}
+
+
 void ElfView::DefineElfSymbol(BNSymbolType type, const string& incomingName, uint64_t addr, bool gotEntry,
-	BNSymbolBinding binding, size_t size, const Confidence<Ref<Type>>& typeObj)
+	BNSymbolBinding binding, size_t size, const Confidence<Ref<Type>>& typeObj, Ref<Platform> functionPlatform)
 {
 	// Ensure symbol is within the executable
 	if (type != ExternalSymbol && !IsValidOffset(addr))
 		return;
+
+	uint64_t originalAddress = addr;
+	if (IsFunctionSymbolType(type) && functionPlatform)
+	{
+		uint64_t adjustedAddress = addr;
+		functionPlatform->GetArchitecture()->GetAssociatedArchitectureByAddress(adjustedAddress);
+		addr = adjustedAddress;
+	}
 
 	string name = incomingName;
 	Confidence<Ref<Type>> symbolTypeRef;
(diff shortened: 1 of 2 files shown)


view/elf/elfview.h
```diff
@@ -516,6 +516,8 @@ namespace BinaryNinja
 		std::map<uint64_t, uint64_t> m_localGotEntries;
 		std::set<uint64_t> m_gotEntryLocations;
 		std::vector<BNRelocationInfo> m_relocationInfo;
+		std::vector<Ref<Metadata>> m_miniDebugSymbolAddresses;
```
zznop, line 519: Recommend naming this `m_miniDebugSymbolOriginalAddresses` or something that indicates these addresses aren't the symbol addresses, but rather the raw addresses that preserve the lsb.

zznop:
Verified that this fixes the issue for `neon crystal gates sequentially`. Some minor feedback, but other than that looks good to me. I do wish we didn't parse that section in a separate view though so we didn't need to use metadata to shuttle over info.

plafosse:
My original implementation used a custom parser to pull that data out. I ended up discarding that in favor of this approach so that we specifically did not have to maintain two ELF parsers. IMO this is a better solution than duplicating all the parse logic.

view/elf/elfview.h
plafosse: they aren't though its a 3-tuple of name, address, original_address

view/elf/elfview.cpp
```diff
@@ -13,6 +13,14 @@ using namespace std;
 
 
 static ElfViewType* g_elfViewType = nullptr;
+static const char* ELF_RECORD_MINI_DEBUG_SYMBOL_ADDRESSES_OPTION = "loader.elf.recordMiniDebugSymbolAddresses";
```
bpotchik, line 16: This is inconsistent with how other setting keys are used. We typically use the string directly at call sites, which also makes it easier to search for usages without an extra layer of indirection.

view/elf/elfview.cpp
```diff
@@ -1436,6 +1447,8 @@ bool ElfView::Init()
 
 	// Process the queued symbols
 	m_symbolQueue->Process();
+	if (!m_miniDebugSymbolAddresses.empty())
+		StoreMetadata(ELF_MINI_DEBUG_SYMBOL_ADDRESS_METADATA, new Metadata(m_miniDebugSymbolAddresses), true);
```
bpotchik, line 1451: Doesn't this end up marking 'analysis changed' via SetAnalysisChanged?

view/elf/elfview.cpp
```diff
@@ -2526,13 +2539,37 @@ bool ElfView::Init()
...
 {
 	// Ensure symbol is within the executable
 	if (type != ExternalSymbol && !IsValidOffset(addr))
 		return;
 
+	uint64_t originalAddress = addr;
+	if (IsFunctionSymbolType(type) && functionPlatform)
+	{
+		uint64_t adjustedAddress = addr;
+		functionPlatform->GetArchitecture()->GetAssociatedArchitectureByAddress(adjustedAddress);
```
bpotchik, line 2569: Why not just pass in `addr` instead of declaring an extra variable?

view/elf/elfview.cpp
```diff
@@ -13,6 +13,14 @@ using namespace std;
 
 
 static ElfViewType* g_elfViewType = nullptr;
+static const char* ELF_RECORD_MINI_DEBUG_SYMBOL_ADDRESSES_OPTION = "loader.elf.recordMiniDebugSymbolAddresses";
+static const char* ELF_MINI_DEBUG_SYMBOL_ADDRESS_METADATA = "ELFMiniDebugSymbolAddresses";
```
bpotchik, line 17: Same for the setting key comment. Unnecessary indirection.

view/elf/elfview.cpp
```diff
@@ -2633,16 +2670,36 @@ void ElfView::DefineElfSymbol(BNSymbolType type, const string& incomingName, uin
...
+			return;
+		if (m_recordMiniDebugSymbolAddresses && IsFunctionSymbolType(symbol->GetType()) &&
+			(symbol->GetAddress() != originalAddress))
+		{
+			map<string, Ref<Metadata>> entry = {
+				{"name", new Metadata(symbol->GetRawName())},
+				{"address", new Metadata(symbol->GetAddress())},
+				{"originalAddress", new Metadata(originalAddress)}
+			};
+			m_miniDebugSymbolAddresses.push_back(new Metadata(entry));
```
bpotchik, line 2686: It's not just a list of addresses is it? Perhaps  `m_miniDebugSymbolEntries` is more clear?

view/elf/elfview.cpp
```diff
@@ -2633,16 +2670,36 @@ void ElfView::DefineElfSymbol(BNSymbolType type, const string& incomingName, uin
 			new Symbol(type, shortName, fullName, rawName, addr, binding, nameSpace), typeRef);
 	};
 
+	auto recordMiniDebugSymbolAddress = [this, originalAddress](const Ref<Symbol>& symbol) {
```
bpotchik, line 2673: recordMiniDebugSymbolEntry?

view/elf/elfview.cpp
```diff
@@ -2786,22 +2894,30 @@ void ElfView::ParseMiniDebugInfo()
...
+		GetStart(), ELF_RECORD_MINI_DEBUG_SYMBOL_ADDRESSES_OPTION, isArmThumb ? "true" : "false");
 	Ref<BinaryView> debugBv = Load(debugElf, false, debugBvOptions);
+
 	if (!debugBv)
 	{
 		m_logger->LogError("Invalid .gnu_debugdata contents: Failed to create BinaryView");
 		return;
 	}
 
+	auto debugPlatforms = GetMiniDebugFunctionPlatforms(debugBv, isArmThumb);
```
bpotchik, line 2909: Why pass `isArmThumb` into GetMiniDebugFunctionPlatforms?

view/elf/elfview.cpp
```diff
@@ -2767,6 +2824,57 @@ bool ElfView::DerefPpc64Descriptor(BinaryReader& reader, uint64_t addr, uint64_t
...
+static map<pair<string, uint64_t>, Ref<Platform>> GetMiniDebugFunctionPlatforms(Ref<BinaryView> debugBv, bool enabled)
+{
+	map<pair<string, uint64_t>, Ref<Platform>> result;
+	if (!enabled || !debugBv)
+		return result;
+
+	Ref<Platform> defaultPlatform = debugBv->GetDefaultPlatform();
+
+	Ref<Metadata> miniDebugSymbolAddresses = debugBv->QueryMetadata(ELF_MINI_DEBUG_SYMBOL_ADDRESS_METADATA);
+	if (!miniDebugSymbolAddresses || !miniDebugSymbolAddresses->IsArray())
```
bpotchik, line 2836: If there are type mismatches that do occur (i.e. !miniDebugSymbolAddresses->IsArray()), should we generate a log warning /error for it?

view/elf/elfview.cpp
```diff
@@ -2767,6 +2824,57 @@ bool ElfView::DerefPpc64Descriptor(BinaryReader& reader, uint64_t addr, uint64_t
...
+		return result;
+
+	for (const auto& entry : miniDebugSymbolAddresses->GetArray())
+	{
+		if (!entry || !entry->IsKeyValueStore())
+			continue;
```
bpotchik, line 2842: Silent failure?

view/elf/elfview.cpp
```diff
@@ -2767,6 +2824,57 @@ bool ElfView::DerefPpc64Descriptor(BinaryReader& reader, uint64_t addr, uint64_t
...
+
+		auto values = entry->GetKeyValueStore();
+		auto name = values.find("name");
+		auto address = values.find("address");
+		auto originalAddress = values.find("originalAddress");
+		if ((name == values.end()) || (address == values.end()) || (originalAddress == values.end()) ||
+			!name->second || !address->second || !originalAddress->second ||
+			!name->second->IsString() || !address->second->IsUnsignedInteger() ||
+			!originalAddress->second->IsUnsignedInteger())
+			continue;
```
bpotchik, line 2852: Silent failure?

bpotchik (commented):
After looking over the change set I realized that the implementation here is way more complicated than it needs to be. The "real" fix should be about 10 lines of code and there shouldn't be any need to funnel a bunch of symbol information through the Metadata APIs.

The first hint that there is something else that needs run to ground is the large spew of errors:
```
[:239 Analysis] Attempting to add function not backed by file: 0x23008
[:239 Analysis] Attempting to add function not backed by file: 0x23ebc
[:239 Analysis] Attempting to add function not backed by file: 0x248f8
[:239 Analysis] Attempting to add function not backed by file: 0x252bc
[:239 Analysis] Attempting to add function not backed by file: 0x252e8
[:239 Analysis] Attempting to add function not backed by file: 0x24cfc
[:239 Analysis] Attempting to add function not backed by file: 0x250a8
[:239 Analysis] Attempting to add function not backed by file: 0x252d2
[:239 Analysis] Attempting to add function not backed by file: 0x24dc4
[:239 Analysis] Attempting to add function not backed by file: 0x2673c
[:239 Analysis] Attempting to add function not backed by file: 0x27c00
```

The BinaryView itself already has the ability to convey the resolved platform information.

The other way to fix this is to define a new setting: `analysis.functions.allowUnbackedMemory`, and pull the arch info from the function itself.

If you want I can push a branch with the 10 line version of this fix.

plafosse:
Closing this as @bpotchik has a superior solution. Here: https://github.com/Vector35/binaryninja-api/tree/fix_gnu_debugdata

PR #8170, opened 2026-05-12
Increment SESSION_COUNT when cloning a Session

### LINKED ISSUE #8166: Rust `Session` has improper `Clone` impl
- Binary Ninja Version: dev
- Edition: Ultimate
- OS: Ubuntu Linux
- OS Version: 24.04
- CPU Architecture: x64

**Bug Description:**
Calling BNCreateDatabaseWithProgress from a headless context returns false very quickly (~7-37µs), with no log output, even with BNLogToStderr(DebugLog) enabled. There is no way for a caller to determine why the call failed. Reproduced under multiple thread / dispatch configurations.

fuzyll:
I believe this would fix #8166. It's pretty tricky to figure out what's going on, but the user gave me a minimal test case that looks something like:

```rust
let session = Session::new().expect("failed to initialize headless session");
{
    let session = session.clone();
    // ...
    let result = tokio::task::spawn_blocking(move || {
        let bv = session.load_with_progress(&input, |_, _| true)
```

I think what's happening is that the `session.clone()` is making a new `Session` while bypassing the registration so `SESSION_COUNT` is never actually incremented. Then, later, when another `Session` has `Drop` called on it, we call `shutdown()` prematurely and that ruins everything else.

It's not clear to me if this solution is the right way to go, or if we should instead just remove `Clone` from `Session` altogether, or if there's a third option that would be better. But, I believe this would at least fix the issue at hand.

### DIFF
--- a/rust/src/headless.rs
+++ b/rust/src/headless.rs
@@ -295,7 +295,7 @@ pub fn license_location() -> Option<LicenseLocation> {
 }
 
 /// Wrapper for [`init`] and [`shutdown`]. Instantiating this at the top of your script will initialize everything correctly and then clean itself up at exit as well.
-#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Default, PartialEq, Eq, Hash)]
 pub struct Session {
     license_duration: Option<Duration>,
 }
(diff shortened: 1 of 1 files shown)


emesare (approved):
LGTM, in the future we may also want to make `Session` impl `Copy` with the same logic as `Clone`, that can be a discussion for later though

fuzyll:
This was rebased and merged as 3021a6889ba7bd36ffd2dd2ce5000284c2d95318. Closing.

PR #8171, opened 2026-05-13
Add little-endian octeon mips arch

jonathanzetier:
I didn't think these actually existed in the wild, but lo and behold, we have stumbled upon a little-endian cavium mips64 ELF.  Unfortunately I can't share it, but I can confirm that this properly disassembles the binary we discovered, whereas before, it would be appear as total garbage because it was trying to decode little-endian MIPS as big-endian.

I'm pretty sure I added the little-endian variant everywhere that had the big-endian variant, but please double-check and let me know if there's somewhere I missed!

### DIFF
--- a/arch/mips/arch_mips.cpp
+++ b/arch/mips/arch_mips.cpp
@@ -3674,6 +3674,21 @@ static Ref<Platform> ElfFlagsRecognize(BinaryView* view, Metadata* metadata)
 	if (!flagsMetadata || !flagsMetadata->IsUnsignedInteger())
 		return nullptr;
 
+	Ref<Metadata> endiannessMetadata = metadata->Get("EI_DATA");
+	if (!endiannessMetadata || !endiannessMetadata->IsUnsignedInteger())
+		return nullptr;
+
+	uint64_t endiannessValue = endiannessMetadata->GetUnsignedInteger();
+	BNEndianness endianness;
+	if (endiannessValue == 1)
+		endianness = LittleEndian;
+	else if (endiannessValue == 2)
+		endianness = BigEndian;
+	else {
+		LogError("ELF endianness value 0x%" PRIx64 " doesn't map to valid value", endiannessValue);
+		return nullptr;
+	}
+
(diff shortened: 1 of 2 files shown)


fuzyll (approved):
I've been working on a thing in the background for awhile that lets  me generate binaries of different architecture/platform combos and I can confirm that this makes a sample binary that does `return 0;` go from:

```
004004b0    int64_t main(int64_t, int64_t, int64_t, int64_t arg4, int32_t arg5 @ $t7)

int64_t arg4  {Register $a3}
int32_t arg5  {Register $t7}

004004b0     0  f0ffbd67   scd     $ra, -0x4299($a3)  {0x1}  {0x0}
❓004004b4     0  0800beff   j       0x2fbfc
004004b8     0  25f0a003   addiu   $s0, $t7, -0x5ffd
```

...to:

```
004004b0    int64_t main() __pure

void var_10  {Frame offset -10}
int64_t __saved_$fp  {Frame offset -8}
void arg_0  {Frame offset 0}

004004b0     0* f0ffbd67   daddiu  $sp, $sp, -0x10
004004b4    10  0800beff   sd      $fp, 8($sp) {__saved_$fp}
004004b8    10  25f0a003   move    $fp, $sp {var_10}
004004bc    10  25100000   move    $v0, $zero  {0x0}
004004c0    10  25e8c003   move    $sp, $fp
004004c4    10  0800bedf   ld      $fp, 8($sp) {__saved_$fp}
004004c8    10* 1000bd67   daddiu  $sp, $sp, 0x10
004004cc     0  0800e003   jr      $ra
004004d0     0  00000000   nop     
```

Which, as far as I can tell, means we've gone from "not real code" to "real code". Seems like an obvious win to me.

Will get this merged as soon as I can. 😄

fuzyll:
This was merged today as c1b0eb7bf58e7234d5930c1c2b4d0987b71c99d1, along with some unit tests I added. Builds of Binary Ninja >= 5.4.9787-dev should have these changes in it. Thanks again for the PR!

PR #8185, opened 2026-05-19
Fix null deref from bad opcode read during lifting

zznop:
Fix for sentry crash [BINARYNINJA-89](https://vector35.sentry.io/issues/7491848949/?alert_rule_id=16897857&alert_type=issue&environment=stable&notification_uuid=f7824bc4-f4e1-48b0-a162-5c286dea66c3&project=4510789549752320&referrer=slack)

This fix adds a few guards:
1. Prevents a possible underflow on `len` before performing the opcode read in `DefaultLiftFunction`
2. Checks the raw pointer returned by `GetDataAt` for `NULL` in `DefaultLiftFunction`
3. Checks data for `NULL` and ensures `len` >= 2 in `populateDecomposeRequest`
4. Adds a guard in the IT-block loop to ensure we don't read OOB

### DIFF
--- a/arch/armv7/thumb2_disasm/arch_thumb2.cpp
+++ b/arch/armv7/thumb2_disasm/arch_thumb2.cpp
@@ -98,9 +98,11 @@ protected:
 		return "thumbv7-none-none";
 	}
 
-	void populateDecomposeRequest(decomp_request *req, const uint8_t *data, size_t len,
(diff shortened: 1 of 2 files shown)


bdash (approved):
Minor nit: your commit message appears to be truncated, mentioning `BINARYNINJA-8`.

Issue #8021, opened 2026-03-18
[DSC] On macOS its not clear what Binary Ninja wants when asking the user to select the primary shared cache file


WeiN76LQh:
**Version and Platform (required):**
- Binary Ninja Version: 5.3.9265-dev Ultimate (564b14be)
- Edition: Ultimate
- OS: macOS
- OS Version: 26.3
- CPU Architecture: M1

**Bug Description:**
On macOS when the DSC plugin can't find the corresponding copy of DSC for a database the file explorer pops up but there's no indication to the user what Binary Ninja is expecting the user to open in the file explorer dialog.

**Expected Behavior:**
Something to indicate to the user why the file explorer is being displayed.

fuzyll:
This should be fixed as of builds >= 5.4.9718-dev and I forgot to close it when my PR (#8188) was merged.

Issue #8020, opened 2026-03-18
[DSC] Failed to load primary shared cache file


WeiN76LQh:
**Version and Platform (required):**
- Binary Ninja Version: 5.3.9265-dev Ultimate (564b14be)
- Edition: Ultimate
- OS: macOS
- OS Version: 26.3
- CPU Architecture: M1

**Bug Description:**
When opening a Binary Ninja database called `dyld_shared_cache_arm64e2.bndb` which is in the same directory (not in a project) as the extracted DYLD Shared Cache and another database called `dyld_shared_cache_arm64e.bndb` I get the following errors:

```
[SharedCache.Builder error] Failed to add file 'dyld_shared_cache_arm64e': File does not start with 'dyld': '/<some path>/dyld_shared_cache_arm64e.bndb'
[SharedCache.View alert] Failed to add primary cache file: 'dyld_shared_cache_arm64e'
[FileMetadata error] View type 'DSCView' could not be created
[Undo error] UndoBuffer deserialization error: Unable to retrieve BinaryView of type: DSCView.
[Undo error] UndoBuffer deserialization error: Unable to retrieve BinaryView of type: DSCView.
[Undo error] UndoBuffer deserialization error: Unable to retrieve BinaryView of type: DSCView.
[Undo error] UndoBuffer deserialization error: Unable to retrieve BinaryView of type: DSCView.
[Undo error] UndoBuffer deserialization error: Unable to retrieve BinaryView of type: DSCView.
[Undo error] UndoBuffer deserialization error: Unable to retrieve BinaryView of type: DSCView.
[Undo error] UndoBuffer deserialization error: Unable to retrieve BinaryView of type: DSCView.
...
```

When the binary view opens its in the hex view and everything is missing in the linear view.

**Steps To Reproduce:**
Please provide all steps required to reproduce the behavior:
1. Extract the DYLD Shared Cache using something like `ipsw`.
2. Open it in Binary Ninja.
3. Save a database with the name `dyld_shared_cache_arm64e.bndb` in the same directory as the extracted DSC.
4. Load an image (might not need this step).
5. Save the binary view again as another database with the name `dyld_shared_cache_arm64e2.bndb` in the same directory as the other one.
6. Close the binary view.
7. Open the database `dyld_shared_cache_arm64e2.bndb`.
8. Observe the aforementioned issue.

**Binary:**
DYLD Shared Cache for iOS 26.2 for an iPhone 17 Pro Max.

[...]

Issue #6630, opened 2025-04-12
[SharedCache] Confusing experience opening DSC database when the DSC files cant be located


WeiN76LQh:
**Version and Platform (required):**
- Binary Ninja Version: 5.0.7218-dev (c480ce0c)
- OS: macOS
- OS Version: 15.2
- CPU Architecture: M1

**Bug Description:**
When opening a DSC database in a project and the DSC files cannot be found, the user experience is confusing and its not clear what the user is being asked to open. Also the open file dialog is the OS one not the Binary Ninja project file browser one.

**Steps To Reproduce:**
Please provide all steps required to reproduce the behavior:
1. Create a new project.
2. Copy all the DSC files into a folder in that project.
3. Open the DSC.
4. Save the database into a different top level folder in the project to the one with the DSC files in it.
5. Close the binary view.
6. Open the database in the project.
7. Observe that the user is prompted to open a file using the OS file browser and no information is provided to explain what they should be opening. Additionally it should not be the OS file browser but the project file browser that Binary Ninja provides for opening files in a project.

**Expected Behavior:**
It should be clear to the user what is happening (the plugin can't find the related DSC files) and when being asked to find a file in a project the user should be provided with a project file browser not an OS one because the names of the files on disk for projects are hard to relate to actual files in the project due to their random names.

**Additional Information:**
This behaviour might be worse on macOS than other platforms due to the file dialogs not have text to indicate what the user should be opening.

emesare:
I am going to add some more metadata to track the entries that were initially added to the shared cache, so that on subsequent loads of the database we can understand where to load the entries from, without having to worry about the original file path that binary ninja saves being overwritten.

That is to say, I believe the prompt you are getting is a little overzealous and we can do better to just _not_ show it instead, with some specific metadata entries for the file name we can just do the lookup for you.

emesare:
I went ahead and pushed what I believe to be the "fix" for this, because this is better (easier) to just show visually I am just going to attach a video.

https://github.com/user-attachments/assets/48e6f4b7-102d-4b1f-a17c-cbab8ca643aa

emesare:
Oops, did not mean to close! Please let me know if the above is acceptable, atleast for the short term, like I mentioned above there is still more that can be done, but it is more cumbersome.

For the above it is available as of dev builds `>7221`.

WeiN76LQh:
[...]

PR #8188, opened 2026-05-20
Primary shared cache file UX improvements

fuzyll:
This PR addresses the following issues:

* #8020
* #8021
* #6630

It also adds some clarification on how analysis databases for shared caches are handled differently from regular analysis databases. Had an Enterprise customer that lost some time because their expectations didn't account for this difference and figured it would be good to spell it out more clearly.

One of the primary changes here is the introduction of a load setting, `loader.dsc.primaryFilePath`, which is intended to be used in headless settings to specify where the primary shared cache file is. An example of using this is now also in the documentation.

This is still pretty annoying, though, so I've changed and expanded how we search for the primary shared cache file. We should now handle cases like accidentally choosing a `.bndb` file better. We should also be able to handle shared caches being stored within projects more gracefully as well. I tried to keep Enterprise collaboration in mind with these changes and I don't believe there should be any strange conflicts introduced.

### DIFF
--- a/view/sharedcache/core/SharedCacheView.cpp
+++ b/view/sharedcache/core/SharedCacheView.cpp
@@ -23,6 +23,25 @@ static bool IsUsablePrimaryCachePath(const std::string& path)
 		&& std::filesystem::is_regular_file(path, ec);
 }
 
+
+static std::string PathRelativeTo(const std::string& path, const std::string& basePath)
+{
+	std::error_code ec;
+	auto relativePath = std::filesystem::relative(path, basePath, ec);
+	auto relativePathString = relativePath.generic_string();
+	if (ec || relativePath.empty() || relativePathString == ".." || relativePathString.find("../") == 0)
+		return path;
+	return relativePath.string();
+}
+
+
+static std::string ResolveRelativePath(const std::string& path, const std::string& basePath)
+{
+	if (path.empty() || std::filesystem::path(path).is_absolute())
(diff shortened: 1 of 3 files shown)


view/sharedcache/core/SharedCacheView.cpp
```diff
@@ -1004,38 +1054,239 @@ void SharedCacheView::LogSecondaryFileName(std::string secondaryFileName)
...
+	Ref<Project> project = nullptr;
+	if (currentProjectFile)
+		project = currentProjectFile->GetProject();
+
+	auto storePrimaryProjectFile = [&](ProjectFile* projectFile) -> std::string {
+		SetPrimaryFileLocation(projectFile->GetPathInProject(), projectFile->GetName());
+		return projectFile->GetPathOnDisk();
+	};
+
+	auto storePrimaryFilePath = [&](const std::string& path) {
```
bdash: Can these lambdas be turned into real functions? It's hard to follow the flow of the function proper with all of these in here, and the fact they each capture `&` makes it harder to tell what they do use from the enclosing scope.

fuzyll: Yeah, sorry, bad habit while prototyping stuff. Will totally make them separate functions instead.

emesare (approved):

fuzyll:
This was merged in 5.4.9718-dev.

PR #8197, opened 2026-05-22
Rust bindings needed for custom function lifters

zznop:
Rust bindings for core functions required to implement a custom function lifter in an architecture plugin. These bindings were required to add support for Texas Instruments TMS320C6x

### DIFF
--- a/rust/src/architecture.rs
+++ b/rust/src/architecture.rs
@@ -27,7 +27,7 @@ use crate::{
     data_buffer::DataBuffer,
     disassembly::InstructionTextToken,
     ffi::INVALID_REGISTER,
-    function::Function,
+    function::{Function, Location, NativeBlock},
     platform::Platform,
     rc::*,
     relocation::CoreRelocationHandler,
@@ -35,6 +35,7 @@ use crate::{
     types::{NameAndType, Type},
     Endianness,
 };
+use std::collections::{HashMap, HashSet};
 use std::ops::Deref;
 use std::{
     borrow::Borrow,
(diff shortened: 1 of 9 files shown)


rust/src/architecture.rs
```diff
@@ -591,13 +594,133 @@ pub trait ArchitectureWithFunctionContext: Architecture {
 
 pub struct FunctionLifterContext {
     pub(crate) handle: *mut BNFunctionLifterContext,
+    pub function: *mut BNLowLevelILFunction,
```
emesare: Seems strange to expose the `function` as a raw pointer, can we ref count it like everything else?

rust/src/architecture.rs
```diff
@@ -591,13 +594,133 @@ pub trait ArchitectureWithFunctionContext: Architecture {
...
+    pub platform: Ref<Platform>,
+    pub logger: Ref<Logger>,
+    pub blocks: Vec<Ref<BasicBlock<NativeBlock>>>,
+    pub no_return_calls: HashSet<Location>,
+    pub contextual_returns: HashMap<Location, bool>,
+    pub inlined_remapping: HashMap<Location, Location>,
+    pub user_indirect_branches: HashMap<Location, HashSet<Location>>,
+    pub auto_indirect_branches: HashMap<Location, HashSet<Location>>,
+    //pub inlined_calls: HashSet<u64>,
```
emesare: Not used anymore? (the inlined_calls field)

zznop: inlined_calls isn't used by rust at the moment because there's a bigger issue and much heavier lift in that function inlining can't be performed right now for archs that override `lift_function` (only TMS320C6x) because we don't have a FunctionLifterContext::CheckForInlinedCall implementation in Rust and the C++ version (in defaultarch.cpp) won't work for TMS320C6x

emesare: Okay, I would add a comment in the source but that is nitpick, ill go ahead and mark as resolved

rust/src/architecture.rs
zznop: Seems like a reasonable change. `FunctionLifterContext::from_raw makes function ref-counted manually. And `Drop` releases it. The lifter callback passes the LLIL function separately as LowLevelILMutableFunction. If we make `function` ref counted then the `Drop` implementation can go away entirely.

rust/src/architecture.rs
zznop: With that said, I've updated it to still populate the inlined_calls HashSet

zznop:
Added in Added in 5.4.9715-dev

PR #8219, opened 2026-05-27
Misc fixes for R2

### LINKED ISSUE #8211: Crash due to null dereference below WarpFunctionInfoWidget::UpdateInfo
Sentry Issue: [BINARYNINJA-91](https://vector35.sentry.io/issues/7504622858/?referrer=github_integration)

```
EXCEPTION_ACCESS_VIOLATION_READ / 0xd0: Fatal Error: EXCEPTION_ACCESS_VIOLATION_READ / 0xd0
  File "qabstractitemmodel.cpp", line 3410, in QAbstractItemModel::beginResetModel
  File "qstandarditemmodel.cpp", line 2267, in QStandardItemModel::clear
  File "function.cpp", line 352, in WarpFunctionInfoWidget::UpdateInfo
  File "sync.rs", line 2685, in alloc::sync::impl$38::drop
  File "mod.rs", line 804, in core::ptr::drop_in_place
...
(30 additional frame(s) were not displayed)
```

@bdash says:
> The stack trace is not reliable beyond the bottom few frames.
>
> One thing I noticed while looking at this code is that [both of the two `ExecuteOnMainThread`](https://github.com/Vector35/binaryninja-api/blob/39201debeaf9440349f56612bfbc80bc352c4de7/plugins/warp/ui/plugin.cpp#L172-L184) calls inside event handlers registered by `WarpSidebarWidget` appear to be unsafe. They assume `this` is still valid, but the widget can be destroyed between the event handler running and the lambda passed to `ExecuteOnMainThread` running. I'm not sure that issue would result in this particular crash though. They seem like they'd crash sooner since the `WarpCurrentFunctionWidget` and `WarpFunctionInfoWidget` would likely also have been destroyed.

emesare:
- [Windows] Fix x64 platform view init race condition
  - This mirrors the fix that was performed on the x86 platform
- [WARP] Fix race condition with sidebar deletion with analysis completion event callback
  - Fixes https://github.com/Vector35/binaryninja-api/issues/8211
- [Objective-C] Fix misc crashes for bad input binaries
  - Fixes https://github.com/Vector35/binaryninja/issues/1470
- [PE] Add a fast fail for malformed PE exception directory table size
  - Fixes https://github.com/Vector35/binaryninja/issues/1472
-  [RTTI] Add a fast-fail to the Itanium RTTI parser for large unbacked sections
   - Fixes   https://github.com/Vector35/binaryninja/issues/1471
- [ELF] Fix MIPS symbol lookup hang with malformed symbol table entry 
  - Fixes   https://github.com/Vector35/binaryninja/issues/1471

### DIFF
--- a/objectivec/objc.cpp
+++ b/objectivec/objc.cpp
@@ -6,6 +6,10 @@
 
 #define RELEASE_ASSERT(condition) ((condition) ? (void)0 : (std::abort(), (void)0))
 
+#define MAX_PROTOCOL_COUNT 0x1000
+#define MAX_METHOD_LIST_COUNT 0x1000
+#define MAX_IVAR_LIST_COUNT 0x1000
+
 using namespace BinaryNinja;
 
 namespace {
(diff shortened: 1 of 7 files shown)


plafosse (approved):
Did review in person and approved given the changes we discussed (regarding integer overflow.)

PR #8222, opened 2026-05-27
[codex] Guard disposed BinaryView handles

### LINKED ISSUE #8161: Crash on use of out of scope binary view in Python
- Binary Ninja Version: 5.4.9605
- Edition: Ultimate
- OS: macOS
- OS Version: 26.4.1
- CPU Architecture: M5

**Bug Description:**
If you load a binary view with a `with` block, then access the variable holding the view outside the `with` block, the process crashes trying to read a null pointer in the FFI layer.

This is buggy Python code, but it shouldn't crash the process.

**Steps To Reproduce:**
Paste the following code in the Python console:
```python
if True:
    with load("/bin/ls") as a:
        pass
    print(a)
```
The process should immediately crash with a null pointer access.

**Expected Behavior:**
Should get a Python exception instead of crashing the entire process.

plafosse:
## Summary

Fixes Vector35/binaryninja-api#8161 by keeping the raw BinaryView handle private and routing public access through a guard that raises `ReferenceError` after disposal. `repr()` now reports disposed views explicitly instead of attempting to query freed state.

## Behavior

```python
>>> with load("/bin/ls") as b:
...  print(b)
... 
... 
<BinaryView: '/bin/ls', start 0x100000000, len 0xe310>
>>> b
<BinaryView: disposed>
>>> b.start
  File "<console>", line 1, in <module>
  File "/Users/peterl/src/binaryninja/cmake-build-relwithdebinfo/out/binaryninja.app/Contents/MacOS/plugins/../../Resources/python/binaryninja/binaryview.py", line 3730, in start
    return core.BNGetStartOffset(self.handle)
  File "/Users/peterl/src/binaryninja/cmake-build-relwithdebinfo/out/binaryninja.app/Contents/MacOS/plugins/../../Resources/python/binaryninja/binaryview.py", line 3248, in handle
    raise ReferenceError("BinaryView has been disposed")
ReferenceError: BinaryView has been disposed
```

## Validation

- `git diff --check -- python/binaryview.py`

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -3197,7 +3197,7 @@ class BinaryView:
 			_handle = core.BNCreateCustomBinaryView(self.__class__.name, file_metadata.handle, _parent_view, self._cb)
 
 		assert _handle is not None
-		self.handle = _handle
+		self._handle = _handle
 		self._notifications = {}
 		self._parse_only = False
 		self._preload_limit = 5
@@ -3210,9 +3210,9 @@ class BinaryView:
 		for i in self._notifications.values():
 			i._unregister()
 		self._notifications.clear()
-		if self.handle is not None:
-			core.BNFreeBinaryView(self.handle)
-			self.handle = None
+		if self._handle is not None:
+			core.BNFreeBinaryView(self._handle)
+			self._handle = None
 
 	def __enter__(self) -> 'BinaryView':
 		return self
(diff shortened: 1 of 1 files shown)


rssor:
This appears to match semantics with how Python behaves with disposed-of files, so +1 here

fuzyll (approved):
Nice, this behavior more closely matches with what I was trying to get at in the office before I left. Looks good to me.

PR #8245, opened 2026-06-05
[idb_import] Broaden IDB import: function folders, variables, value locations, and type fidelity

ChrisKader:
Substantially expands what the IDB importer brings over from an IDB/I64 and applies to the Binary Ninja database, plus correctness and type-translation improvements. Verified end-to-end against a large arm64 database.

### New data imported
- **Function folders**: the IDA Functions-window folder hierarchy (dirtree) is parsed, preserving arbitrary nesting, and recreated as Binary Ninja components (shown as folders in the Symbol List). Functions at the dirtree root are left unfoldered, matching IDA.
- **Stack-frame variables**: named locals/arguments from each function's frame, placed using the IDA frame geometry (frsize/frregs) mapped to BN's stack offset convention.
- **Register variables (regvars)**: registers the user renamed within a function, resolved by name to BN registers.
- **Argument & return value locations**: explicit stack and register storage locations (ArgLoc Stack/Reg1/Reg2/RRel and retloc) resolved via the processor's register names — no hardcoded per-arch tables.
- **No-return functions** and **local labels** that were previously parsed but never applied.

### Type translation fidelity
- C basic-type sizing taken from the TIL header (bool/short/int/long/long long/ long double); correct BoolSized handling; pointer __ptr32/__ptr64 widths; variadic detection; struct/union widths computed by the real layout engine (alignment, bitfields, tail padding); udt extra padding; unknown-unsized type mapped to void.

### Integrity & correctness
- Verifies the IDB's recorded input-file SHA256 against the loaded binary.
- Base-address rebasing aligns to the lowest mapped segment (format agnostic), fixing imported addresses being shifted by the header size.
- Segments are deduplicated by exact range as well as name.

### Cleanup
- Removes the orphaned/dead `types.rs` translator (superseded by the active translator) and resolves the plugin's outstanding TODOs.


plafosse:
This is great thank you for the PR!

ChrisKader:
Is there anything you need me to do? I am trying to transition to BN from using IDA Pro as my main RE tool so I expect more PRs to come for this plugin.

emesare:
Nope! We will have this reviewed and go from there, thanks for the PR!

plugins/idb_import/src/mapper.rs
```diff
@@ -45,13 +67,21 @@ impl IDBMapper {
...
-                    .sections()
+                // Align against the lowest mapped *segment*, not the lowest section.
+                // IDA's `min_ea` is the image base, and it maps the file header region too.
+                // Across executable formats the first *section* generally begins after the
+                // format's headers/metadata (e.g. the Mach-O header + load commands, the PE
+                // headers, the ELF header + program headers), so aligning to the lowest section
+                // over-shifts every imported address by the header size. Segments include the
+                // header region, so the lowest segment matches IDA's image base regardless of
+                // format.
+                let bn_segment_addr = view
```
emesare: In all of the binaries I have tested, this is not the case, for example, loading this binary you will see with the change to use segments we no longer have a correct base address to map with.

[ConsoleApplication1.zip](https://github.com/user-attachments/files/29022012/ConsoleApplication1.zip)

If you can provide a sample that proves otherwise we can discuss how best to go about solving this for both scenarios.

plugins/idb_import/src/mapper.rs
```diff
@@ -73,35 +103,51 @@ impl IDBMapper {
...
             rebased_export.address = rebase(export.address);
-            self.map_export_to_view(view, &til_translator, &rebased_export);
+            if let Some(bn_func) =
+                self.map_export_to_view(view, &til_translator, &rebased_export)
+            {
+                functions_by_address.insert(rebased_export.address, bn_func);
+            }
         }
 
-        // TODO: The below undo and ignore is not thread safe, this means that the mapper itself
```
emesare, line 103: There is a reason this was left as a TODO instead of a NOTE, we do not consider this solved and we would like to fix this behavior, not leave it as-is.

plugins/idb_import/src/mapper.rs
```diff
@@ -170,8 +246,10 @@ impl IDBMapper {
         if let Ok(_used_types) = til_translator.used_types.lock() {
             used_types = _used_types.clone();
         }
-        // TODO: Adding types to view after the types have been applied to the functions is not a
```
emesare, line 173: Again, there is a reason this was left as a TODO.

plugins/idb_import/src/mapper.rs
```diff
@@ -205,8 +283,9 @@ impl IDBMapper {
                 }
             }
 
+            // NOTE: A further lookup source would be the TILs attached to the IDB (e.g. imported
+            // library TILs) before giving up; not yet wired in.
```
emesare: Again, there is a reason this was left as a TODO.

plugins/idb_import/src/mapper.rs
```diff
@@ -314,14 +413,21 @@ impl IDBMapper {
...
             tracing::warn!("Failed to add function for {:0x}", func.address);
-            return;
+            return None;
         };
 
+        // IDA marks functions that never return (e.g. `abort`, `exit`); carry that over so BN's
+        // analysis does not fall through past calls to them.
+        if func.is_no_return {
+            tracing::debug!("Marking function as no-return: {:0x}", func.address);
+            bn_func.set_auto_can_return(false);
```
emesare, line 690: Not entirely sure this will be persisted, will need to save a database with a function this applies with and check.

plugins/idb_import/src/mapper.rs
```diff
@@ -339,7 +445,9 @@ impl IDBMapper {
             }
         }
 
-        // TODO: Attach a platform tuple to the FunctionInfo?
+        // NOTE: We let the function inherit the view's default platform. Carrying an explicit
```
emesare: Leave this as a TODO

plugins/idb_import/src/mapper.rs
```diff
@@ -348,6 +456,182 @@ impl IDBMapper {
...
+            };
+            let reg_width = register.info().size().max(1);
+            let var = Variable::from_register(register);
+            tracing::debug!(
+                "Mapping register variable {} => {} at {:0x}",
+                reg_var.register,
+                reg_var.name,
+                func.address
+            );
+            bn_func.create_user_var(&var, &Type::int(reg_width, false), &reg_var.name, false);
```
emesare, line 823: Use of create_user_var will add an undo action, the caller of this function should wrap this so that the undo history is forgotten (similar to comments).

plugins/idb_import/src/mapper.rs
```diff
@@ -348,6 +456,182 @@ impl IDBMapper {
...
+            match til_translator.translate_type_info(&member.member_type) {
+                Ok(ty) => {
+                    tracing::debug!(
+                        "Mapping stack variable '{}' at frame offset {} (bn {}) for {:0x}",
+                        name,
+                        frame_offset,
+                        bn_offset,
+                        func.address
+                    );
+                    bn_func.create_auto_stack_var(bn_offset, &ty, &name);
```
emesare: We do create_user_var for register variables but then use the auto variant for stack variables?

plugins/idb_import/src/mapper.rs
```diff
@@ -348,6 +456,182 @@ impl IDBMapper {
...
+                    let Some(parent) = parent else {
+                        continue;
+                    };
+                    let rebased = rebase(*address);
+                    // Prefer the function we created during mapping; fall back to a view lookup in
+                    // case it came from another source.
+                    let func = functions_by_address.get(&rebased).cloned().or_else(|| {
+                        view.functions_at(rebased)
+                            .iter()
+                            .find(|func| func.start() == rebased)
```
emesare: What is the point of this? functions_at queries based off the start address of the function.

plugins/idb_import/src/parse.rs
```diff
@@ -85,6 +114,9 @@ pub struct ID0Info {
     pub comments: Vec<CommentInfo>,
     pub labels: Vec<LabelInfo>,
     pub exports: Vec<ExportInfo>,
+    /// Processor register names indexed by IDA register number, used to resolve the registers
+    /// referenced by argument/return value locations into Binary Ninja registers.
+    pub register_names: Vec<String>,
```
emesare, line 174: If this is indexed by IDA register number the assumption here is there are no gaps and the index starts at 0, I think it would be useful to add that invariant as a comment for future reference.

plugins/idb_import/src/parse.rs
```diff
@@ -94,6 +126,20 @@ pub struct DirTreeInfo {
...
+    /// The IDA "Functions" window folder hierarchy (root-level entries).
+    pub function_folders: Vec<FunctionFolderEntry>,
+}
+
+/// An entry in IDA's function folder tree: either a function (by address) or a named folder
+/// containing further entries. Mirrors IDA's dirtree so it can be recreated as Binary Ninja
+/// components.
+#[derive(Debug, Clone, Serialize)]
+pub enum FunctionFolderEntry {
+    Function(u64),
```
emesare, line 193: Use function::Location in places where we expect to "unique" a location for a function, since our definition of a function location is rooted in the absolute address and its architecture.

plugins/idb_import/src/translate.rs
```diff
@@ -192,25 +256,37 @@ impl TILTranslator {
...
                 // so we are going to be liberal and allow unknown basic types to be treated as a sized int.
                 Ok(TypeBuilder::int(*bytes as usize, false))
             }
-            Basic::Bool => Ok(TypeBuilder::bool()),
-            Basic::BoolSized { .. } => {
-                // TODO: This needs to be resized, if that cannot be done, make a NTR to an int named BOOL?
-                Ok(TypeBuilder::bool())
-            }
+            Basic::Bool if self.bool_size == 1 => Ok(TypeBuilder::bool()),
+            // Binary Ninja's `bool` is always a single byte; a wider `bool` is represented as an
```
emesare, line 274: We need to instead make an NTR to int with a more descriptive name then, as the comment suggests.

plugins/idb_import/src/translate.rs
```diff
@@ -219,19 +295,29 @@ impl TILTranslator {
...
 
     pub fn build_pointer_ty(
         &self,
         pointer_ty: &idb_rs::til::pointer::Pointer,
     ) -> anyhow::Result<TypeBuilder> {
-        // TODO: Consult pointer_ty.closure (is this how we can get based pointers?)
+        // A `__ptr32` / `__ptr64` modifier overrides the platform address size for this pointer
+        // (e.g. a 32-bit pointer embedded in an otherwise 64-bit binary).
+        //
+        // NOTE: `closure` (based pointers) and `shifted` pointers have no direct Binary Ninja
```
emesare: Based pointers do exist in Binary Ninja, although we do not expose simple constructors for them and you must use the `TypeBuilder::set_pointer_base` function.

emesare (changes requested):
There are a few issues, if you do not have time to address them I can take over, thank you for your time.

ChrisKader:
Corrections have been made.

ChrisKader:
The thing that IDA does, that is odd, is create "virtual" segments for things like "HEADER" and i had mis-accounted for that before.

emesare:
Thank you for the fixes, we went ahead and merged your parsing changes a few days ago (see idb-rs).

We went ahead and rolled up changes we deemed reasonable for a single "change", in the future it is easier/faster for us to review changes in a more isolated manor, you can see the rolled up commit merged here https://github.com/Vector35/binaryninja-api/commit/6f40e6b10df27456f3f7cc908c1521a7db03004f it is missing a few things primarily the operand application stuff found on your dev branch. For new functionality we encourage you to utilize seperate git branches so we can correctly pull just the commits needed, thank you!

PR #8246, opened 2026-06-05
defaultarch: handle caller-set return addresses when inlining calls

appleflyerv3:
detect registers that are set to the caller's next addr before a call is inlined, and therefore treat the jumps to those regs in the callee IL as returns to the call label.

### DIFF
--- a/defaultarch.cpp
+++ b/defaultarch.cpp
@@ -16,6 +16,37 @@ static bool GetNextFunctionAfterAddress(Ref<BinaryView> data, Ref<Platform> plat
 	return nextFunc != nullptr;
 }
 
+static bool IsZeroConstant(LowLevelILInstruction expr)
+{
+	return ((expr.operation == LLIL_CONST) || (expr.operation == LLIL_CONST_PTR)) && (expr.GetConstant() == 0);
+}
+
+
+static bool IsConstantPointer(LowLevelILInstruction expr, uint64_t value)
+{
+	return ((expr.operation == LLIL_CONST) || (expr.operation == LLIL_CONST_PTR)) && ((uint64_t)expr.GetConstant() == value);
+}
+
+
+static bool IsReturnAddressRegisterExpr(LowLevelILInstruction expr, const set<uint32_t>& returnAddressRegisters)
+{
+	switch (expr.operation)
+	{
+	case LLIL_REG:
+		return returnAddressRegisters.count(expr.GetSourceRegister<LLIL_REG>()) != 0;
+	case LLIL_ADD:
+		return (IsReturnAddressRegisterExpr(expr.GetLeftExpr<LLIL_ADD>(), returnAddressRegisters)
+				&& IsZeroConstant(expr.GetRightExpr<LLIL_ADD>()))
+			|| (IsZeroConstant(expr.GetLeftExpr<LLIL_ADD>())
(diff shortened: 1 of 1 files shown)


appleflyerv3:
This fix allows Binary Ninja to recognize during function inlining when an inlined function returns by jumping to a register that was set to the caller’s fallthrough address. (lifted via LLIL_JUMP)

For example, this is used by RISC-V LLVM outlined functions such as `jal t0, OUTLINED_FUNCTION_*`, followed by `jr t0` within `OUTLINED_FUNCTION_*`, where `jr t0` is lifted as `LLIL_JUMP(t0)`.
For binaries using this pattern, this allows `OUTLINED_FUNCTION_*` helpers, which are typically used as stack prologue/epilogue helpers, to be inlined correctly so the caller’s stack state is recovered.

Example:

```asm
sub_function:
    jal t0, OUTLINED_FUNCTION_1
    ...
    jal t0, OUTLINED_FUNCTION_2

OUTLINED_FUNCTION_1:
    addi sp, sp, -0x10
    sw   ra, 0xc(sp)
    sw   s0, 0x8(sp)
    sw   s1, 0x4(sp)
    sw   s2, 0x0(sp)
    jr   t0

OUTLINED_FUNCTION_2:
    lw   ra, 0xc(sp)
    lw   s0, 0x8(sp)
    lw   s1, 0x4(sp)
    lw   s2, 0x0(sp)
    addi sp, sp, 0x10
    jr   t0
```

zznop:
I have reviewed/tested this PR and added the following commits to the [test_pr_8246](https://github.com/Vector35/binaryninja-api/commits/test_pr_8246/) branch:
* [This](https://github.com/Vector35/binaryninja-api/commit/c9ddc18257447b23d78e0b06483cdff00f04864b) commit avoids treating LLIL_JUMP(reg) as an inline return when the inlined calllee clobbers that register before returning and also adds support for the tailcalls
* [This](https://github.com/Vector35/binaryninja-api/commit/a872917854c79971492114c8f91377af03f6020c) commit fixes lifting of the RISC-V `jal` instruction to set the return register when `jal` is using something other than `ra` (the link register).

defaultarch.cpp
```diff
@@ -16,6 +16,37 @@ static bool GetNextFunctionAfterAddress(Ref<BinaryView> data, Ref<Platform> plat
 	return nextFunc != nullptr;
 }
 
+static bool IsZeroConstant(LowLevelILInstruction expr)
+{
+	return ((expr.operation == LLIL_CONST) || (expr.operation == LLIL_CONST_PTR)) && (expr.GetConstant() == 0);
+}
+
+
+static bool IsConstantPointer(LowLevelILInstruction expr, uint64_t value)
```
plafosse, line 25: this is a misnomer

appleflyerv3: the function name?
what would be a more suitable name for this, then?

plafosse: IsConstantValue perhaps

zznop: I changed the function name to `ConstantCompare` since it's not only checking that the expression is a constant, it's also checking if it is equal to `value`

https://github.com/Vector35/binaryninja-api/commit/10e6eb0e224b041f0e4afc572f414804051d83d1

zznop:
This commit has been merged and is available in `5.4.9920-dev`

PR #8250, opened 2026-06-06
[idb_import] Display operands against their IDA enumeration

ChrisKader:
## Summary

Imports the operands IDA displays against an enumeration and renders them the same way in Binary Ninja (e.g. `orr w1, w8, #VFALSE`, `mov w8, #LUA_ERRORHANDLERINDEX`), plus supporting fixes to the operand-display path.

Builds on the broader `idb_import` work; the headline addition is enum-displayed operands.

## What it does

- **Resolve the enumeration (idb-rs).** Reads the per-operand enum reference from id0 and resolves the member's tid back to the exact owning enumeration by tid range, so it returns the correct enum regardless of member-name collisions and whether the enum lives in a type library or the local types.
- **Find every enum operand.** Probes the enum altval directly for operands 0/1 instead of gating on the operand-representation flag, which does not reflect the enum on instructions like `orr w1, w8, #imm`.
- **Key the override correctly.** `set_int_display_type`'s operand index is the number of operand-separator tokens before the token in the rendered line, not IDA's operand number; both the enum and number-format passes now count separators.
- **Apply after functions exist.** The import runs before functions are created, so overrides were silently dropped. They are now stashed per-view and applied from a `BinaryViewInitialAnalysisCompletionEvent` handler once functions exist, then re-analysis is requested so they render.
- **Enum precedence.** Enum displays are applied before number formats and the format pass skips enum addresses, so IDA's enumeration display is not overwritten where an operand also carries a number-format flag.

## Settings

- `analysis.idb.applyOperandEnums` — display enum operands (small, cheap).
- `analysis.idb.applyOperandFormats` — apply per-operand number formats.
- `analysis.idb.skipDefaultOperandFormats` (default true) — skip number formats already matching Binary Ninja's default (hex), which are the bulk on large databases.

## Note

Includes the binding fix for the use-after-free in `Function::set_int_display_type`, also submitted standalone as #8249; the enum display does not render without it.

## Testing

Verified end-to-end on an arm64 database: enum operands render their members (`LUA_ERRORHANDLERINDEX`, `LUA_GLOBALSINDEX`, `VFALSE`, `VUPVAL`) after a full-analysis load through the plugin's auto-load path.


emesare:
Needs rebase

PR #8252, opened 2026-06-07
Ensure out-of-tree API builds succeed across all platforms

fuzyll:
A user last week complained that we had broken the ability to build the API and some examples with a change we made. It wasn't broken for _us_, since we build in-tree, but customers don't have the rest of our source code.

As it turns out, @CouleeApps wrote a nice little script that lets us test this out-of-tree build process in February 2022 so we wouldn't have this problem. Unfortunately, it appears to have been taken out of our build process pretty quickly because it added a lot of overall time to builds.

I've gone ahead and set up a new nightly process for ensuring we are made aware when we break builds for customers. To do this, I've split the original script @CouleeApps wrote into two–one for us to test in-tree builds with, and one to test out-of-tree builds with. (They were previously both in the same script, so the former would fail with my setup that approximates a customer's environment.)

In the process of doing this, I discovered that the Triage example would no longer build on Windows and the API itself would not build with clang 14 on Ubuntu 22.04. This PR fixes those issues in the quickest and most straight-forward way I could figure out how.

I have separately run a full build with these API changes through CI as well, which has succeeded. Just want someone else to look these over and ensure they look reasonable before I merge them.


fuzyll (commented):
After some changes to the build servers, I've got this branch down to 0 code changes in Binary Ninja or its build scripts. There are now only changes to the `test_build_*.py` scripts.

Since this resolves all of the concerns that people brought up internally, I'm just going to merge it myself so the test will work.

Issue #3077, opened 2022-04-20
Theming System Improvements


cxnder:
## Misc Improvements:
* Certain UI elements utilize seemingly random theme attributes (e.g. New Tab uses Graph View theming attributes) This should probably be cleaned up.
* Allow colors passed as [r, g, b] to have a 4th optional alpha field ([r, g, b, a])
* Documentation for stylesheet support and how it interfaces with binaryninja is *badly* needed.
* Documentation for theming in general needs some improvements.

### Theme Inheritance

One critical pain at the moment is that there are no "defaults", or simple method to grab one of the embedded themes as a "template".

Rather than manually exporting the embedded themes and uploading them somewhere, I'd propose a "parent" optional flag in the bntheme that allows inheriting another theme's dictionaries. 

This would allow making "theme packs" a bit easier as well

### Pallete Brush feature

Currently, themes can define args that specify the palette color, but not the palette *brush*. This would allow for the ability to load QPixmaps (passed as image urls) as pane backdrops.

### Icon Overrides

Not quite sure about this one, but I think it would be a nice feature to have. An access method in the API that fetches the theme override or default would be great.

CX330Blake:
BTW, if we can have a theme manager like plugin manager will be cool.

psifertex:
Worth noting since this was filed that 
https://docs.binary.ninja/dev/themes.html exists now. For the theme manager request, see #1076

fuzyll:
As of builds >= `5.4.9978-dev`, the second bullet-point here about `.bntheme` accepting alpha values has been implemented. I have also open sourced all of our themes as `.bntheme` files in the `themes/` directory of this repository for people to reference.

I also cleaned up some situations where we had hard-coded colors that should've been themeable. This doesn't entirely address the first bullet-point, but it's progress.

Leaving this open as there's still more to do, but wanted to provide a status update.

PR #8258, opened 2026-06-11
Theme system improvements

fuzyll:
This branch makes a number of changes and improvements to how we handle themes within Binary Ninja:

* No themes are constructed with code, all themes use the `.bntheme` JSON system
* All default themes are now open source so people can inspect them, copy them, and make PRs against them
* We no longer map unknown keys to black and, instead, map them to `nullopt`, which will provide a descriptive error instead of silently loading the theme with colors blacked out
* The `.bntheme` JSON system now accepts a 4-value color that provides alpha as the 4th value (required for some of our default themes, also see request in Vector35/binaryninja-api#3077)


fuzyll:
Ah, wrong repository. Off my game today. 😣 Closing this because it also comes with core changes and the PR should be elsewhere.

PR #8260, opened 2026-06-11
Improve compatibility with iOS / macOS 27 shared caches

bdash:
* `objc_msgSend$stub` functions no longer appear in the `__objc_stubs` section of their dylib. Instead they're coalesced across multiple dylibs and appear in a stub island region of the shared cache. This means that `AnalyzeStubFunction` can no longer determine the type of stub it is processing purely based on the containing section name. It now considers the target of the call to determine the type of the stub.
* `objc_msgSend` and friends now have definitions in multiple dylibs throughout the shared cache (`/usr/lib/objc/libobjcMsgSendN.dylib`). This means that loading the target of `objc_msgSend` calls within `objc_msgSend$stub` functions is not sufficient to make selector definitions visible to analysis. Instead, we explicitly load `/usr/lib/libobjc.A.dylib` whenever we process a stub function that references `libobjcMsgSendN.dylib`.
* Add support for method type strings that are pointers relative to selector base address.
* Introduce a new activity to rename `objc_msgSend$stub` stub functions. They no longer have symbol information since they live in stub island regions outside of images. This also helps with regular stripped Mach-O binaries.

### DIFF
--- a/plugins/workflow_objc/src/activities/name_stubs.rs
+++ b/plugins/workflow_objc/src/activities/name_stubs.rs
@@ -1,0 +1,99 @@
+use binaryninja::{
+    low_level_il::instruction::{InstructionHandler as _, LowLevelILInstructionKind},
+    symbol::{Symbol, SymbolType},
+    variable::PossibleValueSet,
+    workflow::AnalysisContext,
+};
+
+use crate::{
+    activities::objc_msg_send_calls::{call_target_type, selector_from_call, MessageSendType},
+    error::ILLevel,
+    metadata::GlobalState,
+    Error,
+};
+
+// Reconstruct names for Objective-C selector stubs, e.g. `_objc_msgSend$length`.
+// The compiler emits one of these stubs per selector. Each one does nothing but load the selector and tail-call
+// `objc_msgSend`.
+pub fn process(ac: &AnalysisContext) -> Result<(), Error> {
+    let view = ac.view();
+    if GlobalState::should_ignore_view(&view) {
+        return Ok(());
+    }
+
+    let func = ac.function();
+    let func_start = func.start();
+
+    // Don't override a name the user has assigned to this function.
+    if let Some(symbol) = view.symbol_by_address(func_start) {
+        if !symbol.auto_defined() {
+            return Ok(());
+        }
+    }
+
+    // Bail out early for functions that do not look like a stub: a single basic block of a few instructions.
+    const MAX_STUB_SIZE: u64 = 32;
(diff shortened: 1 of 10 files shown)


plugins/workflow_objc/src/activities/name_stubs.rs
```diff
@@ -0,0 +1,99 @@
...
+// Reconstruct names for Objective-C selector stubs, e.g. `_objc_msgSend$length`.
+// The compiler emits one of these stubs per selector. Each one does nothing but load the selector and tail-call
+// `objc_msgSend`.
+pub fn process(ac: &AnalysisContext) -> Result<(), Error> {
+    let view = ac.view();
+    if GlobalState::should_ignore_view(&view) {
+        return Ok(());
+    }
+
+    let func = ac.function();
```
emesare, line 24: I think we need to update `AnalysisContext::function` to be an `Option<Ref<Function>>`, since it can be executed in the context of the binary view for a module level workflow.

This isn't an issue with your PR I just noticed it when reviewing.

plugins/workflow_objc/src/activities/name_stubs.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    // The tail call terminates the stub's single basic block, so it can only be the last instruction.
+    let blocks = ssa.basic_blocks();
+    let Some(block) = blocks.iter().next() else {
+        return Ok(());
+    };
+    let Some(insn) = block.iter().last() else {
+        return Ok(());
+    };
+    let LowLevelILInstructionKind::TailCallSsa(call_op) = insn.kind() else {
```
emesare, line 63: Is this also suppose to capture jumps as well? https://github.com/Vector35/binaryninja-api/pull/8261/changes#diff-153013bd9df2d3608be397874180b01d8e12bf3415cfe2cf26f001ab287d4120R47 I see this and it seems like it should? Not sure.

bdash: The code you link to deals with cross-image stub functions into an image that is not yet loaded. In that case the call will often be represented as a jump. This case is a little different.

`objc_msgSend$stub` functions always end with a call to `objc_msgSend`. `AnalyzeStubFunction` in SharedCacheWorkflow.cpp automatically loads `libobjc` (and/or the new `/usr/lib/objc/libobjcMsgSendN.dylib`) when it processes one of these stub functions, so any jump instruction that may have existed will have been converted to a tailcall.

emesare (approved):
Seeing as the comment above about the `AnalysisContext::function` call relates to the bindings and not the workflow itself, we can do that separately.

PR #8261, opened 2026-06-11
[DSC] Resolve cross-image stub functions to their target image in context menu actions

bdash:
This ensures the "Load /usr/lib/libFoo.dylib" context menu item will be offered when right-clicking on a stub function that calls into libFoo.dylib.

### DIFF
--- a/view/sharedcache/ui/SharedCacheUINotifications.cpp
+++ b/view/sharedcache/ui/SharedCacheUINotifications.cpp
@@ -4,6 +4,7 @@
 
 #include "SharedCacheUINotifications.h"
 #include <sharedcacheapi.h>
+#include "mediumlevelilinstruction.h"
 #include "ui/sidebar.h"
 #include "ui/linearview.h"
 #include "ui/viewframe.h"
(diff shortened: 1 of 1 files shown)


view/sharedcache/ui/SharedCacheUINotifications.cpp
```diff
@@ -15,6 +16,58 @@ using namespace SharedCacheAPI;
...
+	auto mlil = func->GetMediumLevelIL();
+	if (!mlil)
+		return std::nullopt;
+
+	for (const auto& block : mlil->GetBasicBlocks())
+	{
+		for (size_t i = block->GetStart(), end = block->GetEnd(); i < end; ++i)
+		{
+			const auto instr = mlil->GetInstruction(i);
+			if (instr.operation != MLIL_JUMP && instr.operation != MLIL_TAILCALL)
```
emesare: since both these ops will terminate the block u can just check the last instr in the block, and since u check to see if the blocks in the native function are just one, u can skip doing the loop over blocks as well

emesare: I would also probably swap out the maxStubLength from checking native block size and go off the number of mlil instructions, just to make it more streamlined

bdash: The check on native size is so it can bail out without generating MLIL for functions that are clearly not stubs.

emesare (approved):

PR #8262, opened 2026-06-11
[armv7] Implement IsELFDataRelocation more efficiently

bdash:
There's no reason for these to create `std::map`s when a switch does the trick.


fuzyll (approved):
Good find! I'm not really sure why this was written that way. I also wouldn't have immediately realized this was a performance problem if it hadn't come up in the context of the user's issue today.

fuzyll:
This landed in build 5.4.9815-dev.

PR #8263, opened 2026-06-11
[ELF] Improve performance of loading binaries with many sections not mapped into the address space

bdash:
These sections are added to the parent view. Do the work within a bulk segment modification to significantly reduce the overhead of adding sections.


fuzyll (approved):
Looks good to me. Crazy that this is all we needed.

fuzyll:
This landed in build 5.4.9815-dev.

PR #8270, opened 2026-06-15
[BNTL] Add compiler flags processing to BNTL utils

xitska:
Small change to allow adding of clang compiler flags when creating BNTLs.
The CLI is untested as I have no headless license, but I'm happy to make changes where needed for ergonomics.

### DIFF
--- a/plugins/bntl_utils/cli/src/create.rs
+++ b/plugins/bntl_utils/cli/src/create.rs
@@ -19,6 +19,9 @@ pub struct CreateArgs {
     pub include_directories: Vec<PathBuf>,
     #[clap(long)]
     pub dry_run: bool,
+
+    #[arg(last = true, allow_hyphen_values = true, num_args = 0..)]
+    compiler_flags: Vec<String>,
 }
 
 impl CreateArgs {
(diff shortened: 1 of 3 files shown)


plugins/bntl_utils/src/process.rs
```diff
@@ -434,6 +434,8 @@ pub struct TypeLibProcessor {
     include_directories: Vec<PathBuf>,
     /// Whether to process existing type libraries when processing a binary file.
     process_existing_type_libraries: bool,
+    /// The compiler flags to use when processing header files.
+    options: Vec<String>,
```
emesare: Rename to compiler_options or compiler_flags

plugins/bntl_utils/src/command/create.rs
```diff
@@ -124,6 +124,23 @@ impl PlatformField {
     }
 }
 
+pub struct FlagsField;
```
emesare: Rename to compiler flags or compiler options to align with the processor field

plugins/bntl_utils/cli/src/create.rs
```diff
@@ -19,6 +19,9 @@ pub struct CreateArgs {
     pub include_directories: Vec<PathBuf>,
     #[clap(long)]
     pub dry_run: bool,
+
+    #[arg(last = true, allow_hyphen_values = true, num_args = 0..)]
+    compiler_flags: Vec<String>,
```
emesare: Add a comment like include_directories that tells the user this is only used when parsing from C header files.

emesare (approved):

PR #8277, opened 2026-06-18
[DSC] Add a Strings table to the shared cache triage view

bdash:
The table displays strings from every image in the shared cache. Double-clicking on a string loads its corresponding image or region and navigates to it.

The entire shared cache is scanned asynchronously on the worker pool to discover strings when the Strings tab is first selected. Table filtering and sorting is performed on background threads due to the sheer amount of data this table contains (over 15 millions of rows). Data is discarded shortly after the strings table is hidden.

This change introduces a `BackgroundSortFilterRows` helper class that handles sorting and filtering of flat UI models on background threads, and a `TriageTablePanel` that encapsulates a lot of the logic that is common across DSC's different panes. These are both used by the new strings table, and I updated the DSC's symbols table to use them as well.

### DIFF
--- a/ui/backgroundsortfilterrows.h
+++ b/ui/backgroundsortfilterrows.h
@@ -1,0 +1,589 @@
+#pragma once
+
+#include <algorithm>
+#include <atomic>
+#include <cstdint>
+#include <functional>
+#include <memory>
+#include <numeric>
+#include <optional>
+#include <utility>
+#include <vector>
+
+#include <QFuture>
+#include <QFutureWatcher>
+#include <QThread>
+#include <QThreadPool>
+#include <QtConcurrent/QtConcurrentMap>
+#include <QtConcurrent/QtConcurrentRun>
+
+#include "base/assertions.h"
+#include "binaryninjacore.h"
+
+/*! The non-template portion of `BackgroundSortFilterRows`.
+	Contains the background job state machine, watcher lifecycle, and abandonment protocol,
+	which are independent of the row type.
+
+	\ingroup filter
+*/
+class BackgroundSortFilterRowsBase
+{
+public:
+	// Model integration points. The reset and insert hooks bracket changes to the display rows,
+	// matching the corresponding QAbstractItemModel methods.
+	struct ModelHooks
+	{
+		std::function<void()> beginResetModel;
+		std::function<void()> endResetModel;
+		std::function<void(int first, int last)> beginInsertRows;
+		std::function<void()> endInsertRows;
+		// Optional. Called when background filtering starts and finishes.
+		std::function<void(bool active)> filteringChanged;
+		// Optional. Called when background sorting starts and finishes.
+		std::function<void(bool active)> sortingChanged;
+	};
+
+	explicit BackgroundSortFilterRowsBase(ModelHooks hooks) : m_hooks(std::move(hooks)) {}
+	virtual ~BackgroundSortFilterRowsBase() { shutdown(); }
+
+	BackgroundSortFilterRowsBase(const BackgroundSortFilterRowsBase&) = delete;
+	BackgroundSortFilterRowsBase& operator=(const BackgroundSortFilterRowsBase&) = delete;
+
+	bool busy() const { return m_jobActive; }
+	bool filtering() const { return m_jobActive && m_jobKind == JobKind::Filter; }
+	bool sorting() const { return m_jobActive && m_jobKind == JobKind::Sort; }
+
+protected:
+	enum class JobKind
+	{
+		Filter,
+		Sort,
+	};
+
+	void assertOwningThread() const { BN_ASSERT(QThread::currentThread() == m_owningThread); }
+
+	// Abandon any in-flight job, wait for its workers to notice, and destroy the watcher
+	// immediately so a queued finished signal cannot run a commit handler afterwards. Derived
+	// destructors must call this before the row storage the workers read is destroyed.
+	void shutdown()
+	{
+		m_filterGeneration.fetch_add(1);
+		m_future.waitForFinished();
+		m_watcher.reset();
+	}
+
+	// Like `shutdown`, but leaves the object reusable. Returns it to the idle state and notifies
+	// that the abandoned job is no longer running. The job's result is discarded, not committed.
+	void abandonJob()
+	{
+		if (!m_jobActive)
+			return;
+		shutdown();
+		m_jobActive = false;
+		if (m_jobKind == JobKind::Filter && m_hooks.filteringChanged)
+			m_hooks.filteringChanged(false);
+		else if (m_jobKind == JobKind::Sort && m_hooks.sortingChanged)
+			m_hooks.sortingChanged(false);
+	}
+
+	// Registers `future` as the running job. Its completion invokes `commitJob` back on the
+	// owning thread. `filterGeneration` must be the value the job's workers compare against.
+	void beginJob(JobKind kind, uint64_t filterGeneration, QFuture<void> future)
+	{
+		BN_ASSERT(!m_jobActive);
+		m_jobActive = true;
+		m_jobKind = kind;
+		m_jobFilterGeneration = filterGeneration;
+		m_future = std::move(future);
+		m_watcher = std::make_unique<QFutureWatcher<void>>();
+		QObject::connect(
+			m_watcher.get(), &QFutureWatcherBase::finished, m_watcher.get(), [this] { finishJob(); });
+		if (kind == JobKind::Filter && m_hooks.filteringChanged)
+			m_hooks.filteringChanged(true);
+		else if (kind == JobKind::Sort && m_hooks.sortingChanged)
+			m_hooks.sortingChanged(true);
+		m_watcher->setFuture(m_future);
+	}
+
+	// Whether the filter changed after the current or just-finished job captured its generation.
+	bool filterChangedDuringJob() const { return m_jobFilterGeneration != m_filterGeneration.load(); }
+
+	// Called on the owning thread when the registered job's future finishes.
+	virtual void commitJob(JobKind kind) = 0;
+
+	ModelHooks m_hooks;
+	// Bumped when the filtered set becomes stale: on a filter-predicate change, and on teardown.
+	// Filter workers watch it to drop a superseded filter pass. Sort workers watch it too, because a
+	// filter change invalidates the display snapshot they are reordering, and commitSortJob reads it
+	// (via filterChangedDuringJob) to recognize that case and re-filter before re-sorting.
+	std::atomic<uint64_t> m_filterGeneration = 0;
+
+private:
+	void finishJob()
+	{
+		// This is invoked by the watcher's finished signal so the watcher cannot be deleted
+		// directly here.
+		m_watcher.release()->deleteLater();
+		m_jobActive = false;
+		commitJob(m_jobKind);
+	}
+
+	QThread* m_owningThread = QThread::currentThread();
+	uint64_t m_jobFilterGeneration = 0;
+	bool m_jobActive = false;
+	JobKind m_jobKind = JobKind::Filter;
+	std::unique_ptr<QFutureWatcher<void>> m_watcher;
+	QFuture<void> m_future;
+};
+
+/*! Append-only row storage for table models that filters and sorts rows on the worker pool.
+
+	Filtering and sorting run as QtConcurrent jobs. Changing the filter abandons a running job at
+	its next check and re-runs with the latest predicate, so rapid filter changes stay responsive.
+	Appends and sorts requested while a job is running are deferred and applied when the job
+	commits, allowing the job to read the rows without locking.
+
+	All methods must be called on the thread the object was created on (the GUI thread for table
+	models): the job state machine and row vectors are single-thread-confined, with the worker
+	threads communicating back only through the watcher's queued finished signal, and reading only
+	the atomic generation counters and the unmutated rows. The filter predicate and comparator are
+	called concurrently from worker threads and must be thread-safe.
+
+	\ingroup filter
+*/
+template <typename Row>
+class BackgroundSortFilterRows : public BackgroundSortFilterRowsBase
+{
+public:
+	using Predicate = std::function<bool(const Row&)>;
+	using PredicateFactory = std::function<Predicate()>;
+	using Comparator = std::function<bool(const Row&, const Row&)>;
+	// Display entries are indices into the master rows. 32 bits suffices: a QAbstractItemModel
+	// addresses rows with int, so it can never display more than INT_MAX of them.
+	using Index = uint32_t;
+
+	explicit BackgroundSortFilterRows(ModelHooks hooks) : BackgroundSortFilterRowsBase(std::move(hooks)) {}
+
+	~BackgroundSortFilterRows() override
+	{
+		// The workers read the row storage, so they must be stopped before it is destroyed.
+		shutdown();
+	}
+
+	// Number of rows a view should currently display. Either all rows, or the matches of the
+	// committed filter.
+	size_t displayCount() const { return m_display.size(); }
+
+	// The row at display position `index`, resolving through the filter/sort index.
+	const Row& displayAt(size_t index) const
+	{
+		BN_ASSERT(index < m_display.size());
+		return m_rows[m_display[index]];
+	}
+
+	// Count of all rows, ignoring any filter, including rows whose append is deferred.
+	size_t totalCount() const { return m_rows.size() + m_pendingRows.size(); }
+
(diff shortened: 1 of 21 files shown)


ui/backgroundsortfilterrows.h
```diff
@@ -0,0 +1,573 @@
...
+		std::vector<Range> chunks;
+		chunks.reserve((display.size() + chunkSize - 1) / chunkSize);
+		for (size_t begin = 0; begin < display.size(); begin += chunkSize)
+			chunks.emplace_back(begin, std::min(begin + chunkSize, display.size()));
+
+		QtConcurrent::blockingMap(chunks, [&](const Range& chunk) {
+			size_t comparisons = 0;
+			try
+			{
+				std::sort(display.begin() + chunk.first, display.begin() + chunk.second, guarded(comparisons));
```
fuzyll, line 334: Would it be possible to have non-deterministic ordering here? If I'm following this correctly, results will arrive in the order in which workers complete. Since there's no tie-break on address/length/something else, wouldn't this have ties ordered by worker completion?

(Even if I'm right, I don't know how often this would occur, whether this would be noticeable, etc. Just an observation.)

ui/backgroundsortfilterrows.h
```diff
@@ -0,0 +1,573 @@
...
+		const uint64_t sortGeneration = m_sortGeneration.load();
+		m_sortResult = std::make_shared<std::vector<Index>>();
+		auto result = m_sortResult;
+
+		// The display indices are not mutated while the job is active (appends and sorts are
+		// deferred), so the worker sorts a copy and reads m_rows without locking. An abandoned sort
+		// leaves the shared result empty, which commitSortJob never reads.
+		auto future = QtConcurrent::run(
+			[this, filterGeneration, sortGeneration, comparator, result, display = m_display]() mutable {
+				if (auto sorted = parallelSort(std::move(display), m_rows, comparator, filterGeneration,
```
fuzyll, line 406: Because `parallelSort` uses `blockingMap`, if we're thread-starved, I think this could wind up in a state where this `run` would occupy one thread and then not be able to start the inner thread for `parallelSort`? Does Qt handle this situation for us? I genuinely don't know how `QtConcurrent` works and https://doc.qt.io/qt-6/qtconcurrentmap.html wasn't super helpful when trying to figure this out.

ui/backgroundsortfilterrows.h
```diff
@@ -0,0 +1,573 @@
...
+			if (m_hooks.filteringChanged)
+				m_hooks.filteringChanged(false);
+			applySortAfterDisplayChange();
+			return;
+		}
+
+		const uint64_t filterGeneration = m_filterGeneration.load();
+		PredicateFactory factory = m_filterFactory;
+
+		constexpr size_t chunkSize = 1024 * 1024;
```
fuzyll, line 481: Should this be defined somewhere more visible/configurable?

view/sharedcache/api/python/sharedcache.py
```diff
@@ -92,6 +92,71 @@ def image_to_api(image: CacheImage) -> sccore.BNSharedCacheImage:
...
     )
 
+@dataclasses.dataclass
+class CacheString:
+    string_type: sccore.StringTypeEnum
+    address: int
+    raw_length: int
+    text: str
+    region_start: int
+    image_start: int
```
fuzyll: In `sharedcache.cpp`, you state that "A zeroed imageStart means the region is not associated with an image". Should this detail be surfaced to Python as well in some capacity (even if just a doc comment)?

view/sharedcache/core/CacheStringScanner.cpp
```diff
@@ -0,0 +1,209 @@
...
+	constexpr size_t maxLength = CacheStringScanner::kMaxDisplayTextLength;
+	switch (ref.type)
+	{
+	case AsciiString:
+	case Utf8String:
+		return std::string(reinterpret_cast<const char*>(data), std::min(ref.length, maxLength));
+	case Utf16String:
+	{
+		char* converted = BNUnicodeUTF16ToUTF8(data, std::min(ref.length, maxLength * 2));
+		std::string text(converted);
```
fuzyll: If we fail to convert here, we'd try to construct a `std::string` from a `nullptr`. This seems like it might only happen if we fail to allocate the string? Which, is probably not likely to happen, but we still might want to check.

This caused me to feel like a fraud because it turns out I have no idea what actually happens if you try and construct a string from `nullptr`. Fortunately, it appears C++ [also had no idea what should happen](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2166r1.html), but C++23 appears to be fixing this.

view/sharedcache/core/CacheStringScanner.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+	{
+		std::lock_guard lock(state->completionMutex);
+		state->remainingRegions = state->regions.size();
+	}
+	for (size_t i = 0; i < state->regions.size(); i++)
+	{
+		// Scanning starts only when the user opens the Strings tab and is watching it fill, so it
+		// runs at priority above the normal-priority background analysis of off-screen code.
+		WorkerPriorityEnqueue([state, i] {
```
fuzyll, line 85: If we have a large number of cache regions, would this cause any problems if enqueued workers is significantly greater than threads available? Also wonder if not doing some kind of bounding/batching could cause issues with the destructor simply marking things as `abort`.

The design is totally fine and avoids deadlocks/use-after-frees which are worse, only wondering if we could potentially cause performance issues by leaving a large number of enqueued threads running while they try to abort.

view/sharedcache/core/CacheStringScanner.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+		std::span<const uint8_t> data;
+		try
+		{
+			data = state.vm->ReadSpan(cur, dataLen);
+		}
+		catch (std::exception& e)
+		{
+			// This happens if we have not mapped in all the relevant entries.
+			state.logger->LogErrorF("Failed to read region {:#x} while scanning for strings: {}", region.start, e.what());
```
fuzyll, line 162: If this is expected, should this be a warn instead of an error? Or even an info/debug? Not clear to me if this is something a user could take an action to recover from or if this is something they should report to us because it indicates a bug, or if this is kind of normal.

view/sharedcache/core/CacheStringScanner.cpp
```diff
@@ -0,0 +1,209 @@
...
+// Builds the UTF-8 display text for a string found at `data`, truncated to
+// `CacheStringScanner::kMaxDisplayTextLength` bytes of output.
+std::string DisplayTextForString(const uint8_t* data, const BNStringReference& ref)
+{
+	constexpr size_t maxLength = CacheStringScanner::kMaxDisplayTextLength;
+	switch (ref.type)
+	{
+	case AsciiString:
+	case Utf8String:
+		return std::string(reinterpret_cast<const char*>(data), std::min(ref.length, maxLength));
```
fuzyll: If we're truncating on `maxLength` bytes, I think this could cut through a codepoint and create an invalid UTF-8 character. Only thought about this because of that bug that came up earlier where we had a crash on save because of an invalid UTF-8 character.

view/sharedcache/core/CacheStringScanner.cpp
```diff
@@ -0,0 +1,209 @@
...
+	case Utf16String:
+	{
+		char* converted = BNUnicodeUTF16ToUTF8(data, std::min(ref.length, maxLength * 2));
```
fuzyll: Same comment about codepoints here as above.

view/sharedcache/core/ffi.cpp
fuzyll, line 83: Do we intentionally not restore `region.imageStart` here? Noticed in passing that the to/from here don't match and there's no comment explaining why that should be.

view/sharedcache/core/SharedCacheController.cpp
fuzyll: Also, should this be `m_loadedImages` here? I think this is a copy/paste fail?

view/sharedcache/ui/dsctriage.cpp
```diff
@@ -315,81 +335,298 @@ QWidget* DSCTriageView::initImageTable()
...
+{
+	auto controller = SharedCacheController::GetController(*m_data);
+	if (!controller)
+		return;
+
+	auto region = controller->GetRegionAt(string.regionStart);
+	if (!region.has_value())
+		return;
+
+	QPointer<QFutureWatcher<bool>> watcher = new QFutureWatcher<bool>(this);
```
fuzyll, line 623: Do we leak this `watcher` here?

view/sharedcache/ui/dsctriage.cpp
```diff
@@ -315,81 +335,298 @@ QWidget* DSCTriageView::initImageTable()
...
 		{
-			if (button == dialog->button(QMessageBox::Yes))
-				loadImagesWithAddr({image->headerAddress});
-		});
+			navigateToAddress(string.address);
+			return;
+		}
 
-		dialog->exec();
+		promptToLoadImage(image->name, string.address, string.address);
```
fuzyll: As far as I can tell, this works just fine. But, I was surprised it wasn't a bug. I think this could be clearer/more correct if we passed `*string.imageStart` or `image->headerAddress` since we're loading the image.

view/sharedcache/ui/stringstable.cpp
```diff
@@ -0,0 +1,333 @@
...
+QVariant StringsTableModel::data(const QModelIndex& index, int role) const
+{
+	if (!index.isValid())
+		return QVariant();
+
+	switch (role)
+	{
+	case Qt::DisplayRole:
+	{
+		const auto& string = stringAt(index.row());
```
fuzyll, line 95: Is checking `.isValid()` enough to prevent issues here? https://doc.qt.io/qt-6/qmodelindex.html#isValid suggests this just means it's non-negative, not that it actually fits within the correct upper bound. I don't know that this could ever crash, but it might be possible to return wrong results if the model is ever stale?

view/sharedcache/ui/symboltable.cpp
```diff
@@ -1,61 +1,129 @@
...
 		return QVariant();
-	}
 
 	switch (role)
 	{
 	case Qt::DisplayRole:
 	{
-		auto symbol = symbolAt(index.row());
-		auto symbolType = GetSymbolTypeAsString(symbol.type);
+		const auto& symbol = symbolAt(index.row());
```
fuzyll: Same comment here regarding `.isValid`.

view/sharedcache/ui/triagetable.cpp
```diff
@@ -0,0 +1,530 @@
...
+QString ImageNameLookup::tooltip(std::optional<uint64_t> imageStart, uint64_t regionStart) const
+{
+	return tooltip(*m_state, imageStart, regionStart);
+}
+
+
+QString ImageNameLookup::widestImageName() const
+{
+	QString widest;
+	for (const auto& [_, name] : m_state->imageNames)
```
fuzyll: The image column can display region names for non-image rows, but we only consider image names for sizing. In practice, this is probably fine..?

fuzyll (changes requested):
Lots of questions and a few potential bugs. Ironically, the two things I am _most_ concerned about aren't even in these code changes and are just things I saw when scrolling around and reading. 🙃 

Good work regardless, thanks for taking this on. 🙂

view/sharedcache/core/CacheStringScanner.cpp
bdash: This is copied as is from `SharedCacheController::ApplyRegion` where it reads the region contents into memory. It is never hit in practice, and either indicates a bug in our code or a malformed shared cache. I'm inlined to leave it as is for consistency with the code it is based on.

view/sharedcache/core/ffi.cpp
bdash: Folded the fix into this change.

view/sharedcache/core/SharedCacheController.cpp
bdash: Folded the fix into this change.

view/sharedcache/ui/triagetable.cpp
bdash: In practice it is probably fine because the longest image names are longer than the non-image region names in the shared cache's I've tested with. But it's a bit dumb to rely on that, so I will update it to look at non-image region names as well.

ui/backgroundsortfilterrows.h
bdash: If we find in the future that different uses of this need wildly different values we can make it configurable at that point.

view/sharedcache/ui/dsctriage.cpp
bdash: Sorta. It would be cleaned up when `this` was destroyed, but there's no reason for it to live that long. Fixed.

view/sharedcache/ui/stringstable.cpp
bdash: I've added some checks that the indexes are in range, and assertions so that trying to use invalid indexes is caught in debug builds.

view/sharedcache/ui/dsctriage.cpp
bdash: I've updated this to be the obvious thing so it isn't surprising to read.

view/sharedcache/core/CacheStringScanner.cpp
bdash: In practice the background jobs drain very quickly once they're canceled so I'm not sure it's worth trying to do anything more complicated here.

ui/backgroundsortfilterrows.h
bdash: Yeah, it is possible to get non-deterministic results if the comparator doesn't break ties itself. And the comparators the two table models were using didn't do that.

I decided to address this by revamping how sorting works in `TriageTableRowsModel`. Rather than sorting by a single column, it tracks the columns that the user has sorted by and builds a comparator based on multiple columns. The address is always used as a tiebreaker. This gives the effect of using a stable sort when re-sorting by a different column. For example, if the user sorts by symbol name then re-sorts by image name, the rows will be ordered by image name and within each image they'll be sorted by symbol name. This seems strictly more useful than only ever sorting by a single column.

Using a stable sort by itself wouldn't be sufficient here as any filtered rows would be excluded from the sort, so widening the filter criteria would result in inconsistent sorting for the newly-matched rows. Hence building a compound comparator instead.

view/sharedcache/api/python/sharedcache.py
bdash: I've updated these so non-image regions have an `image_start` of None

ui/backgroundsortfilterrows.h
bdash: Yeah, the documentation isn't super clear about whether this is a problem in practice.

To avoid the potential problem you called out I've gone with bracketing the work with calls to [`QThreadPool::releaseThread`](https://doc.qt.io/qt-6/qthreadpool.html#releaseThread) / `QThreadPool::reserveThread`. This has the effect of increasing the effective thread pool size by one for the duration of the work, compensating for this thread blocking on one of the thread pool's threads. The documentation mentions this pattern at least.

view/sharedcache/core/CacheStringScanner.cpp
bdash: I've decided to address this and the other unicode-related issue you mentioned by adding some Unicode helpers to `bn::base`, and using them here. I'll send the change that adds those helpers out as its own PR and rebase this on top of it.

view/sharedcache/ui/dsctriage.cpp
```diff
@@ -119,21 +133,26 @@ void DSCTriageView::loadImagesWithAddr(const std::vector<uint64_t>& addresses, b
...
+				// Run analysis.
+				this->m_data->AddAnalysisOption("linearsweep");
+				this->m_data->AddAnalysisOption("pointersweep");
+				this->m_data->UpdateAnalysis();
+			}
 
-			// Update the triage to display the images as loaded.
-			for (const auto& image : loadedImages)
-				setImageLoaded(image.headerAddress);
+			if (navigateTo)
```
fuzyll, line 152: I think we might navigate even when image loading fails? Not sure how likely that is to happen, just thought I'd point it out.

fuzyll (approved):
Have one comment, but otherwise I think this is good? There's a sorting difference in the symbol table as well that I noted while looking at a local copy I built from this branch, but I dunno if the change is good or bad.

I think it's possible I might've found two more bugs while reviewing this, but they're not the result of these changes and I see no reason to hold this up over bugs that aren't yours. I'll do some more digging after this is merged. It's late and I should be asleep.

PR #8289, opened 2026-06-26
Add bn::base::SegmentedVector

bdash:
A back-growing segmented array, consisting of a directory of pointers to geometrically sized chunks. It gives portable, controlled chunk sizing (unlike `std::deque`), O(1) random access, and a guarantee that growth never moves, copies, or reallocates existing elements.

Chunk capacities follow a geometric ramp (`B`, `2B`, `4B`, ..., `C`) and then a fixed plateau (`C`, `C`, `C`, ...), where `B` is `InitialChunkElems` and `C` is `MaxChunkElems`, both powers of two. A 0 means "use the byte-budget default for `T`", which gives three usage modes:
```c++
SegmentedVector<T> v;        // byte-budget defaults (small ramp to ~64 KiB)
SegmentedVector<T, N> v;     // pure fixed-size N-element chunks (no ramp)
SegmentedVector<T, B, C> v;  // explicit ramp from B to C
```

### DIFF
--- a/base/segmented_vector.h
+++ b/base/segmented_vector.h
@@ -1,0 +1,958 @@
+// Copyright (c) 2026 Vector 35 Inc
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+
+#pragma once
+
+// A back-growing segmented array: a directory of pointers to geometrically
+// sized chunks. It gives portable, controlled chunk sizing (identical on every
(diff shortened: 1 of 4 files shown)


rssor:
having benchmarked this on my branch, it immediately cuts about 40% of allocations related to instruction storage and will be useful in tons of other places, so in favor of merging and then pointing it towards those asap. I've only checked it with the fixed size 128 chunks, and it's probably worth exploring a few others (32->128 geometric might be a further optimization, but fixed-128 is already ~93% efficient).

plafosse:
@bdash did you say there were som eunit tests that went along with this?

bdash:
Yes, on the branch of the same name in the internal repository.

plafosse:
Seems like there may be a potential inconsistency. bulk_construct ensures we're less than max_size but the other reserve, emplace_ back and push_ back do not.

base/segmented_vector.h
```diff
@@ -0,0 +1,952 @@
...
+
+
+	bool empty() const noexcept { return m_size == 0; }
+	size_type size() const noexcept { return m_size; }
+	size_type max_size() const noexcept { return AllocTraits::max_size(m_alloc); }
+
+	// Total element slots across all currently allocated chunks.
+	size_type capacity() const noexcept { return chunk_base(m_directory.size()); }
+
+	void reserve(size_type n)
```
bpotchik, line 572: Should this check max_size similar to `bulk_construct`?

base/segmented_vector.h
```diff
@@ -0,0 +1,952 @@
...
+				// potentially throwing m_alloc assignment is performed first.
+				Allocator newAlloc(other.m_alloc);
+				Directory newDir{DirAlloc(newAlloc)};
+
+				destroy_and_free_all();
+				m_alloc = std::move(newAlloc);
+				std::destroy_at(&m_directory);
+				std::construct_at(&m_directory, std::move(newDir));
+
+				append_range(other.begin(), other.end());
```
bpotchik, line 355: Can `append_range` throw here after modifying *this? If so, you could possibly create a `SegmentedVector tmp(other, other.m_alloc, CopyTag{});` at the top which moves the throwing bit before the *this modification.

base/segmented_vector.h
bdash: I have added the `max_size` check here.

I did not add it to `push_back` or `emplace_back` because in practice the biggest risk is a bogus size value being passed in. Growing by one element at a time will throw if the allocation fails.

base/segmented_vector.h
bdash: `operator=` on most containers only strives to provide a basic exception guarantee (the object is in a valid but undetermined state, with all invariants preserved, if an exception is thrown). That's the best `operator=` can do here if we want to be able to reuse existing allocations in the common "same allocator" code path.

Doing the copy then swap approach would allow us to provide the strong exception guarantee (the object is rolled back to its initial state if an exception is thrown), but that would either be limited to the less common "different allocator" path or it would prevent reusing allocations. Given that, I don't think it's worth doing.

plafosse:
LGTM

PR #8295, opened 2026-06-29
[KernelCache] Improve compatibility with iOS 27 kernel caches

bdash:
iOS 27 / macOS 27 route cross-image calls through `__auth_stubs`, loading their target from `__auth_got`. `__auth_got` is now marked as read-only to allow the target of the cross-image call to be resolved.

Additionally, a workflow activity was added to name these stub functions, and the context menu actions for loading additional images were updated to resolve the target image through the stub functions. Most of this logic is a simplified version of the equivalent code from the dyld shared cache plug-in.


fuzyll (approved):
Tested this against a new iOS 27 beta kernelcache, looks like everything works exactly as advertised. No notes. Thanks!

PR #8299, opened 2026-06-30
[Python] Fix crash after registering Architecture or ArchitectureHook multiple times

### LINKED ISSUE #8195: UAF In Architecture.register
- Binary Ninja Version: 5.4.9678-dev
- Edition: Commercial
- OS: Windows
- OS Version: 10 Pro
- CPU Architecture: x64

**Bug Description:**
There appears to be a use after free, triggerable in python by calling `Architecture.register()` twice on the same class.

The workaround is simply "Don't register twice".

**Steps To Reproduce:**
Run the following snippet

```python
import gc
import binaryninja as bn
from binaryninja import Architecture, RegisterInfo, InstructionInfo
from binaryninja.function import InstructionTextToken
from binaryninja.enums import Endianness


class MinimalArch(Architecture):
    name = "UAF-Repro"
    address_size = 4
    default_int_size = 1
    instr_alignment = 1
    max_instr_length = 1
    endianness = Endianness.LittleEndian
    regs = {"R0": RegisterInfo("R0", 4)}
    stack_pointer = "R0"

    def get_instruction_info(self, data, addr):
        info = InstructionInfo()
        info.length = 1
        return info

    def get_instruction_text(self, data, addr):
        return [InstructionTextToken(0, "nop")], 1

    def get_instruction_low_level_il(self, data, addr, il):
        return 1


MinimalArch.register()
MinimalArch.register() 
gc.collect()  # Force a collection for the sake of expediency, not necessity

arch = bn.Architecture["UAF-Repro"]  # crashes here
```

bdash:
The second registration could result in the callback structure being deallocated, despite core still holding a reference to it. Architecture now rejects multiple attempts to register the same architecture. Registering an ArchitectureHook adds it to a global list to ensure the instance and its callback struct remains alive.

Fixes https://github.com/Vector35/binaryninja-api/issues/8195.

### DIFF
--- a/python/architecture.py
+++ b/python/architecture.py
@@ -1023,6 +1023,8 @@ class Architecture(metaclass=_ArchitectureMetaClass):
 		binaryninja._init_plugins()
 		if cls.name is None:
 			raise ValueError("architecture 'name' is not defined")
+		if core.BNGetArchitectureByName(cls.name) is not None:
+			raise ValueError(f"architecture '{cls.name}' is already registered")
 		arch = cls()
 		cls._registered_cb = arch._cb
 		arch.handle = core.BNRegisterArchitecture(cls.name, arch._cb)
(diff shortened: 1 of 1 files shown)


fuzyll (approved):
This looks like a correct fix to the reported issue to me.

One note: I think we can still register the same hook over and over? Might be missing something, and even if I'm right I dunno if this is a problem. I was just confused because the title includes ArchitectureHook.

bdash:
I chose to let hooks be registered multiple times as it seemed plausible that you might want to apply the same hooks to multiple variants of an architecture. But for architectures themselves, the use of class properties for a bunch of stuff like the name made that seem less useful. I'm honestly not sure if that's the right decision though.

PR #8300, opened 2026-06-30
Improve x86 flag lifting semantics

### LINKED ISSUE #4044: Implement C, P, A, and O flags on x86

zznop:
## Summary

This improves x86 LLIL flag lifting coverage and correctness across arithmetic, test, rotate, shift, double-shift, and signed multiply instructions. It replaces previously unimplemented or ad hoc flag behavior with proper flag-system callbacks where possible, using newer LLIL operations such as `POPCNT` (for PF) and `Unknown` (to model flags that are expectedly undefined).

## Changes

- Model parity flag computation using `POPCNT`.
- Model auxiliary carry for arithmetic instructions using low-nibble carry/borrow formulas.
- Represent architecturally undefined flags with `LLIL_UNKNOWN`.
- Add or fix CF/OF behavior for:
  - `ADC`
  - `SBB`
  - one-, two-, and three-operand `IMUL`
  - `ROL` / `ROR`
  - `RCL` / `RCR`
  - `SHL` / `SAL`
  - `SHR`
  - `SAR`
  - `SHLD`
  - `SHRD`
- Fix count-sensitive shift/rotate behavior:
  - masked count `0` preserves destination/flags
  - count `1` uses defined OF behavior
  - larger counts produce undefined OF where required
  - memory destinations still avoid writes for count `0`
- Keep register shift/rotate value lifts cleaner by moving count-specific flag behavior into flag callbacks instead of emitting unnecessary LLIL control flow.
- Fix `PTEST` / `VPTEST`:
  - correct CF operand polarity
  - clear OF, SF, AF, and PF
  - use the flag system instead of explicit `SetFlag` writes
- Fix APX conditional flag handling:
  - `CCMPNL` / `CTESTNL` now use signed greater-or-equal
  - EVEX default flag assignment no longer derives PF from CF
- Fix `SHLD` / `SHRD` issues:
  - count-zero no-op behavior
  - correct source merge direction
  - compute ZF/SF/PF from the final merged result
  - model SHRD count-1 OF from the merged result
- Improve `ENTER` lifting for nonzero nesting levels by emitting the frame pointer display setup described by the instruction semantics.

## Testing

I've constructed the attached binary that contains each affected instruction along with `lahf` and `seto` to ensure that flags are consumed and emitted during lifting for observation. I've also tested with other binaries shared by users.

[x86_flag_lift_variants.zip](https://github.com/user-attachments/files/29522657/x86_flag_lift_variants.zip)

Closes #4044


arch/x86/arch_x86.cpp
```diff
@@ -2124,7 +2188,494 @@ size_t X86CommonArchitecture::GetFlagWriteLowLevelIL(BNLowLevelILOperation op, s
...
+						il.Const(size, 1)),
+					il.Const(size, 0));
+				if (!operands[1].constant)
+					return boolChoice(countIsZero, countNotZero, il.Flag(flag), carry);
+				return carry;
+			}
+
+			switch (op)
+			{
+			case LLIL_LSL:
```
emesare: I don't think this is reachable given the top level switch case only covers `LLIL_LSR` and `LLIL_ASR`

zznop: Good catch, that was left over from before I moved `LLIL_LSL` to the top-level switch-case. I've updated the PR.

xusheng6 (approved):
Looks fantastic! I also tested it on a some VMP binaries and I don't see unlifted instructions. Cant wait to dust off my emulator PR and test with along with this

PR #8303, opened 2026-07-02
Extend rbran's parser robustness fixes with file-backed bounds and configurable limits

Weitao-Sun:
## Summary

This PR builds on rbran's parser robustness fixes (commits `1b3afd460`–`cd135b972`,
rebased onto dev) with the following improvements:

**Bounds derived from file structure** rather than arbitrary constants where possible:
- `nindirectsyms` is clamped to the number of 4-byte entries that physically fit
  between `indirectsymoff` and end-of-file, using the OS-reported file size
- `ParseExportTrie` uses `buffer.GetLength()` as its node visit cap, which also
  bounds traversal of cyclic tries and fixes an `endGuard` underflow on empty buffers
- `ReadStringTable` rejects sections whose declared size exceeds the actual file length,
  preventing a large allocation before any data is read

**Configurable per-table budgets** replacing per-opcode modulo caps:
- `files.macho.maxRebaseBindEntriesK` (default 1024 K, max 65536 K): single counter
  shared across all opcodes in one rebase or bind table
- `files.elf.maxMipsGotMB` (default 4 MB, max 64 MB): mirrors `fs/binfmt_elf.c`
- `files.elf.maxStringTableTypeSizeMB` (default 4 MB, max 256 MB): guards
  `ApplyTypesToParentStringTable` against oversized section declarations

**Additional fixes not covered by the original commits:**
- `ReadChainedImportAddend32` / `ReadChainedImportAddend64`: apply the same
  `name_offset` bounds check and `strnlen` guard already present in `ReadChainedImport32`
- `ParseSymbolTable`: compare `n_strx` against `stringList.GetLength()` (actual bytes
  returned by `Read()`) rather than `symtab.strsize`, preventing underflow on truncated files
- `BinaryView` C ABI callbacks: extend exception handling to
  `OnAfterSnapshotDataAppliedCallback` and `SaveCallback`
- `PEView`: replace magic-number import-table loop guards with `IsRVARangeBackedByFile`
  checks that validate both endpoints with wraparound protection

### DIFF
--- a/view/macho/chained_fixups.cpp
+++ b/view/macho/chained_fixups.cpp
@@ -240,18 +240,21 @@ auto FixupReaderForFormat(int format) -> std::pair<uint64_t, FixupInfo>(*)(Binar
 	throw std::invalid_argument("Unknown chained pointer format: " + std::to_string(format));
 }
 
+// Returns the NUL-terminated string starting at `offset` within `symbolData`, or an
+// empty view if `offset` does not fall within `symbolData`.
+std::string_view SymbolNameAt(std::span<const char> symbolData, uint32_t offset)
+{
+	if (symbolData.size() <= offset)
+		return std::string_view();
+	return std::string_view(&symbolData[offset], strnlen(&symbolData[offset], symbolData.size() - offset));
+}
+
 ImportEntry ReadChainedImport32(BinaryReader& reader, std::span<const char> symbolData)
 {
 	dyld_chained_import import;
 	reader.Read(&import, sizeof(import));
-	std::string_view view;
-	if (symbolData.size() > import.name_offset) {
-		view = std::string_view(&symbolData[import.name_offset]);
-	} else {
-		view = std::string_view();
-	}
 	return {
-		view,
+		SymbolNameAt(symbolData, import.name_offset),
 		0,
 		import.lib_ordinal > 0xF0 ? static_cast<int8_t>(import.lib_ordinal) : static_cast<int32_t>(import.lib_ordinal),
 		(bool)import.weak_import,
@@ -263,7 +266,7 @@ ImportEntry ReadChainedImportAddend32(BinaryReader& reader, std::span<const char
 	dyld_chained_import_addend import;
 	reader.Read(&import, sizeof(import));
 	return {
-		std::string_view(&symbolData[import.name_offset]),
+		SymbolNameAt(symbolData, import.name_offset),
 		static_cast<uint32_t>(import.addend),
 		import.lib_ordinal > 0xF0 ? static_cast<int8_t>(import.lib_ordinal) : static_cast<int32_t>(import.lib_ordinal),
 		(bool)import.weak_import,
@@ -275,7 +278,7 @@ ImportEntry ReadChainedImportAddend64(BinaryReader& reader, std::span<const char
 	dyld_chained_import_addend64 import;
 	reader.Read(&import, sizeof(import));
 	return {
-		std::string_view(&symbolData[import.name_offset]),
+		SymbolNameAt(symbolData, import.name_offset),
 		import.addend,
 		import.lib_ordinal > 0xFFF0 ? static_cast<int16_t>(import.lib_ordinal) : static_cast<int32_t>(import.lib_ordinal),
 		(bool)import.weak_import,
--- a/view/macho/chained_fixups.cpp
+++ b/view/macho/chained_fixups.cpp
@@ -313,10 +313,9 @@ std::vector<ImportEntry> ChainedFixupProcessor::ProcessImports() const
 
 	auto header = ReadHeader(reader);
 
-	uint64_t symbolDataSize = m_fixupsSize - header.symbols_offset;
-	if (!symbolDataSize) {
+	if (header.symbols_offset >= m_fixupsSize)
 		return imports;
-	}
+	uint64_t symbolDataSize = m_fixupsSize - header.symbols_offset;
 	m_symbolData.resize(symbolDataSize);
 	m_raw->Read(&m_symbolData[0], OffsetInFixups(header.symbols_offset), symbolDataSize);
 
--- a/view/macho/chained_fixups.cpp
+++ b/view/macho/chained_fixups.cpp
@@ -311,6 +311,9 @@ std::vector<ImportEntry> ChainedFixupProcessor::ProcessImports() const
 	auto header = ReadHeader(reader);
 
 	uint64_t symbolDataSize = m_fixupsSize - header.symbols_offset;
+	if (!symbolDataSize) {
+		return imports;
+	}
 	m_symbolData.resize(symbolDataSize);
 	m_raw->Read(&m_symbolData[0], OffsetInFixups(header.symbols_offset), symbolDataSize);
 
--- a/view/macho/chained_fixups.cpp
+++ b/view/macho/chained_fixups.cpp
@@ -244,8 +244,14 @@ ImportEntry ReadChainedImport32(BinaryReader& reader, std::span<const char> symb
 {
 	dyld_chained_import import;
 	reader.Read(&import, sizeof(import));
+	std::string_view view;
+	if (symbolData.size() > import.name_offset) {
+		view = std::string_view(&symbolData[import.name_offset]);
+	} else {
+		view = std::string_view();
+	}
 	return {
-		std::string_view(&symbolData[import.name_offset]),
+		view,
 		0,
 		import.lib_ordinal > 0xF0 ? static_cast<int8_t>(import.lib_ordinal) : static_cast<int32_t>(import.lib_ordinal),
 		(bool)import.weak_import,
@@ -435,7 +441,7 @@ void ChainedFixupProcessor::ProcessChainsInSegment(const dyld_chained_starts_in_
 
 		bool done = false;
 		while (!done)
-		{ 
+		{
 			uint64_t position = reader.GetOffset();
 			auto [raw, fixupInfo] = fixupReader(reader);
 
--- a/view/macho/machoview.cpp
+++ b/view/macho/machoview.cpp
@@ -3826,8 +3826,13 @@ bool MachoViewType::IsTypeValidForData(BinaryView* data)
 uint64_t MachoViewType::ParseHeaders(BinaryView* data, uint64_t imageOffset, mach_header_64& ident, Ref<Architecture>* arch, Ref<Platform>* plat, string& errorMsg)
 {
 	DataBuffer sig = data->ReadBuffer(imageOffset, 4);
+	if (sig.GetLength() != 4)
+	{
+		errorMsg = "signature too small";
+		return 0;
+	}
 	uint32_t magic = *(uint32_t*)sig.GetData();
-	if ((sig.GetLength() != 4) || !(magic == MH_CIGAM || magic == MH_CIGAM_64 || magic == MH_MAGIC || magic == MH_MAGIC_64))
+	if (!(magic == MH_CIGAM || magic == MH_CIGAM_64 || magic == MH_MAGIC || magic == MH_MAGIC_64))
 	{
 		errorMsg = "invalid signature";
 		return 0;
--- a/view/macho/machoview.cpp
+++ b/view/macho/machoview.cpp
@@ -3338,7 +3338,8 @@ void MachoView::ParseSymbolTable(BinaryReader& reader, MachOHeader& header, cons
 			if (sym.n_strx >= symtab.strsize || ((sym.n_type & N_TYPE) == N_INDR))
 				continue;
 
-			string symbol((char*)header.stringList.GetDataAt(sym.n_strx));
+			const char* symbolName = (const char*)header.stringList.GetDataAt(sym.n_strx);
+			string symbol(symbolName, strnlen(symbolName, header.stringList.GetLength() - sym.n_strx));
 			m_symbols.push_back(symbol);
 			//otool ignores symbols that end with ".o", startwith "ltmp" or are "gcc_compiled." so do we
 			if (symbol == "gcc_compiled." ||
--- a/view/macho/machoview.cpp
+++ b/view/macho/machoview.cpp
@@ -2884,7 +2884,8 @@ void MachoView::ParseExportTrie(BinaryReader& reader, linkedit_data_command expo
 		stack.reserve(64);
 		stack.push_back({ /* cursor */ 0, /* text */ "" });
 
-		while (!stack.empty())
+		uint32_t stack_guard = 0;
+		while (!stack.empty() && stack_guard++ < 0x1000)
 		{
 			m_logger->LogTraceF("Export Trie: Processing node {:?} with cursor {:#x}", stack.back().text, stack.back().cursor);
 			Node node = std::move(stack.back());
--- a/view/macho/machoview.cpp
+++ b/view/macho/machoview.cpp
@@ -1700,13 +1700,21 @@ bool MachoView::Init()
 	Ref<Type> filesetEntryCommandType = Type::StructureType(filesetEntryCommandStruct);
 	m_typeNames.filesetEntryCommandQualName = DefineType(filesetEntryCommandTypeId, filesetEntryCommandName, filesetEntryCommandType);
 
-	if (!InitializeHeader(m_header, true, preferredImageBase, preferredImageBaseDesc, platformSetByUser))
-		return false;
-
-	for (auto& it : m_subHeaders)
+	try
 	{
-		if (!InitializeHeader(it.second, false, it.first, "", platformSetByUser))
+		if (!InitializeHeader(m_header, true, preferredImageBase, preferredImageBaseDesc, platformSetByUser))
 			return false;
+
+		for (auto& it : m_subHeaders)
+		{
+			if (!InitializeHeader(it.second, false, it.first, "", platformSetByUser))
+				return false;
+		}
+	}
+	catch (std::exception& e)
+	{
+		m_logger->LogErrorForExceptionF(e, "MachoView failed to InitializeHeader! {:?}", e.what());
+		return false;
 	}
 
 	std::chrono::steady_clock::time_point endTime = std::chrono::steady_clock::now();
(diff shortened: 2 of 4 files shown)


view/macho/machoview.cpp
```diff
@@ -3824,6 +3883,11 @@ bool MachoViewType::IsTypeValidForData(BinaryView* data)
 
 uint64_t MachoViewType::ParseHeaders(BinaryView* data, uint64_t imageOffset, mach_header_64& ident, Ref<Architecture>* arch, Ref<Platform>* plat, string& errorMsg)
 {
+	if (data->GetLength() < 4)
```
bdash: I think a better fix would to move the existing `sig.GetLength()` check to immediately after `sig` is initialized, before `sig.GetData()` is called. This would mean we're checking what actually matters (the size of `sig`).

view/macho/machoview.cpp
```diff
@@ -3335,10 +3387,17 @@ void MachoView::ParseSymbolTable(BinaryReader& reader, MachOHeader& header, cons
...
+			// while still being past the end of the actual buffer.
+			if (sym.n_strx >= header.stringList.GetLength() || ((sym.n_type & N_TYPE) == N_INDR))
 				continue;
 
-			string symbol((char*)header.stringList.GetDataAt(sym.n_strx));
+			size_t symbol_max_len = strnlen(
+				(char*)header.stringList.GetDataAt(sym.n_strx),
+				header.stringList.GetLength() - sym.n_strx
+			);
+			string symbol((char*)header.stringList.GetDataAt(sym.n_strx), symbol_max_len);
```
bdash: I think this could be written as:
```c++
const char* name = (const char*)header.stringList.GetDataAt(sym.n_strx);
string symbol(name, strnlen(name, header.stringList.GetLength() - sym.n_strx));
```

It's shorter and avoids calling `GetDataAt` multiple times.

view/macho/machoview.cpp
```diff
@@ -2875,7 +2908,11 @@ void MachoView::ParseExportTrie(BinaryReader& reader, linkedit_data_command expo
 		stack.reserve(64);
 		stack.push_back({ /* cursor */ 0, /* text */ "" });
 
-		while (!stack.empty())
+		// Each trie node consumes at least one byte from the buffer, so buffer.GetLength()
+		// is a sound upper bound on the number of reachable nodes and also prevents
+		// infinite traversal when the trie contains cycles.
+		uint32_t nodeCount = 0;
```
bdash: The name `nodeCount` is a little confusing since the comment talks about nodes being visited multiple times in the case of cycles. That suggests this is actually counting visits rather than nodes.

This should probably be a `size_t` to be consistent with the result type of `DataBuffer::GetLength`.

view/macho/machoview.cpp
```diff
@@ -2268,9 +2292,16 @@ bool MachoView::InitializeHeader(MachOHeader& header, bool isMainHeader, uint64_
 		// Handle indirect symbols
 		if (header.dysymtab.nindirectsyms)
 		{
-			indirectSymbols.resize(header.dysymtab.nindirectsyms);
+			// Clamp to the number of 4-byte entries that fit between indirectsymoff and
+			// end-of-file; GetParentView()->GetLength() is the OS-reported file size and
+			// is not derived from any field inside the binary.
+			const uint64_t fileRemaining = (GetParentView()->GetLength() > header.dysymtab.indirectsymoff)
```
bdash: The logic here is kind of hard to follow. I suspect it's incorrect when parsing a universal (aka fat) Mach-O binary since in that case `BinaryReader::SetVirtualBase` is used to give every seek an implicit offset, which this computation does not appear to account for.

It also looks like it'll trigger undefined behavior when this clamping does trigger, since `fileRemaining` will be zero, which means `indirectSymCount` will also be zero, and thus `&indirectSymbols[0]` is accessing an object that is past the vector's bounds.

view/macho/machoview.cpp
```diff
@@ -1700,13 +1711,26 @@ bool MachoView::Init()
 	Ref<Type> filesetEntryCommandType = Type::StructureType(filesetEntryCommandStruct);
 	m_typeNames.filesetEntryCommandQualName = DefineType(filesetEntryCommandTypeId, filesetEntryCommandName, filesetEntryCommandType);
 
-	if (!InitializeHeader(m_header, true, preferredImageBase, preferredImageBaseDesc, platformSetByUser))
-		return false;
-
-	for (auto& it : m_subHeaders)
+	try
```
bdash: Is this necessary / beneficial now that you have also added a try / catch within `BinaryView::InitCallback`?

view/macho/machoview.cpp
```diff
@@ -363,6 +363,17 @@ void BinaryNinja::InitMachoViewType()
 	static MachoViewType type;
 	BinaryViewType::Register(&type);
 	g_machoViewType = &type;
+
+	Settings::Instance()->RegisterSetting("files.macho.maxRebaseBindEntriesK",
```
bdash: Other settings that control Mach-O loading appear to be under `loader.macho` rather than `files.macho`.

view/macho/machoview.cpp
```diff
@@ -363,6 +363,17 @@ void BinaryNinja::InitMachoViewType()
...
+		R"~({
+		"title" : "Maximum Mach-O Rebase/Bind Table Entry Count (K)",
+		"type" : "number",
+		"default" : 1024,
+		"minValue" : 0,
+		"maxValue" : 65536,
+		"description" : "Maximum total number of address entries in thousands (K) produced across all opcodes in a single rebase or bind table",
```
bdash: I'm a little confused by why this value is not just the limit itself, but is instead the limit divided by 1000 (or is it 1024)?

view/macho/machoview.cpp
```diff
@@ -363,6 +363,17 @@ void BinaryNinja::InitMachoViewType()
...
+		"ignore" : ["SettingsProjectScope"]
```
bdash: I'm not sure whether this is correct. No other Mach-O settings have it.

view/macho/machoview.cpp
```diff
@@ -2862,9 +2893,11 @@ bool MachoView::AddExportTerminalSymbol(
 void MachoView::ParseExportTrie(BinaryReader& reader, linkedit_data_command exportTrie)
 {
 	try {
-		uint32_t endGuard = exportTrie.datasize;
 		DataBuffer buffer = GetParentView()
 								->ReadBuffer(m_universalImageOffset + exportTrie.dataoff, exportTrie.datasize);
+		if (buffer.GetLength() == 0)
+			return;
+		uint32_t endGuard = buffer.GetLength() - 1;
```
bdash: It took me a moment to realize that this `- 1` is necessary because it is used as an index. Renaming it to `endIndex` could make this clearer.

It's a little surprising that this is a `uint32_t` given that it's being used to bound `uint64_t` values. That's not new in this change, though.

view/macho/machoview.cpp
```diff
@@ -2967,6 +3004,8 @@ void MachoView::ParseRebaseTable(BinaryReader& reader, MachOHeader& header, uint
 	if (tableSize == 0 || tableOffset == 0)
 		return;
 
+	uint64_t remainingIterations = Settings::Instance()->Get<uint64_t>("files.macho.maxRebaseBindEntriesK") * 1024;
```
bdash: Passing the view as the second argument to `Get` will ensure this respects a setting that is overridden at the view level. I believe we do that where we read the other loader settings in this code.

view/macho/machoview.cpp
```diff
@@ -363,6 +363,17 @@ void BinaryNinja::InitMachoViewType()
 	static MachoViewType type;
 	BinaryViewType::Register(&type);
 	g_machoViewType = &type;
+
+	Settings::Instance()->RegisterSetting("files.macho.maxRebaseBindEntriesK",
+		R"~({
+		"title" : "Maximum Mach-O Rebase/Bind Table Entry Count (K)",
+		"type" : "number",
+		"default" : 1024,
```
bdash: Is this sufficient to only trigger on binaries that are in some way problematic? Most of the large binaries I have around are using chained fixups rather than the older binding tables, but at least one of them (Google Chrome Framework) has over 1 million rebase operations.

I'm not entirely clear what problem this limit is trying to solve, but it seems like a limit could be derived from the Mach-O slice's size. In practice there's no reason to have opcodes rebasing the same offsets multiple times. Perhaps a limit of slice size / address size would be sufficient?

view/macho/machoview.cpp
```diff
@@ -3019,7 +3058,9 @@ void MachoView::ParseRebaseTable(BinaryReader& reader, MachOHeader& header, uint
 				address += immediate * m_addressSize;
 				break;
 			case RebaseOpcodeDoRebaseImmediateTimes:
-				count = immediate;
+				// immediate is the low 4 bits of the opcode byte; its value is 0-15.
+				count = std::min((uint64_t)immediate, remainingIterations);
+				remainingIterations -= count;
```
bdash: It took me a moment to figure out why this was adjusting count like this. It feels a little confusing to modify `count` like this in order to enforce the iteration limit.

I think it'd be clearer to introduce a helper like:

```c++
uint64_t remainingEntries = /* the setting, or some other limit derived from the Mach-O slice size */;
auto emitRebase = [&]() {
	m_logger->LogTraceF("Rebasing address {:#x}", address);
	if (address < segmentStartAddress || address >= segmentEndAddress)
	{
		m_logger->LogError("Rebase address out of segment bounds");
		throw ReadException();
	}

	if (remainingEntries == 0)
	{
		m_logger->LogWarn("Rebase table encodes more entries than fit in the file, ignoring the remainder");
		return false;
	}

	remainingEntries--;
	memset(&rebaseRelocation, 0, sizeof(rebaseRelocation));
	rebaseRelocation.nativeType = BINARYNINJA_MANUAL_RELOCATION;
	rebaseRelocation.address = address;
	rebaseRelocation.size = m_addressSize;
	header.rebaseRelocations.push_back(rebaseRelocation);
	return true;
};
```

Then we could leave `count` alone and change the loop body to:
```c++
if (!emitRebase())
{
	tableDone = true;
	break;
}
address += m_addressSize;
```

The same would apply to the other places in this function that apply relocations, and a similar approach would work for binding in `ParseDynamicTable` as well.

view/macho/chained_fixups.cpp
```diff
@@ -244,8 +244,15 @@ ImportEntry ReadChainedImport32(BinaryReader& reader, std::span<const char> symb
 {
 	dyld_chained_import import;
 	reader.Read(&import, sizeof(import));
+	std::string_view view;
```
bdash: This same pattern is repeated in three different functions, but they're each expressed in slightly different ways (same semantics, different syntax). IMHO it'd be worth extracting it into a helper function so each of these functions can just do `SymbolNameAt(symbolData, import.name_offset)`.

bdash (commented):
Thanks for tackling these fixes! I have only reviewed the Mach-O portion of the changes, as that is code I'm familiar with. I did not look at ELF or PE at all.

One broader comment I have is that it seems like this is split into more commits than it needs to be, including commits that are either redundant or fix up earlier commits in the series and should be squashed back into those commits instead. A few examples that stood out to me:
* It is surprising that the changes to `ReadChainedImport32` are in a separate commit from the identical changes in `ReadChainedImportAddend32` and `ReadChainedImportAddend64`.
* https://github.com/Vector35/binaryninja-api/commit/1e00fc0de7637599e52abfecd604a95193c9f868 and https://github.com/Vector35/binaryninja-api/commit/bb601c5d7751bf9eee3f627308ca22fd8e0e7568 appear to be ideas that were tried but rejected in favor of the larger change in https://github.com/Vector35/binaryninja-api/commit/c484c568174736e4281cbea4b187df9e83ba29bc. They should be dropped.
* The settings registration changes in https://github.com/Vector35/binaryninja-api/commit/051c69700569cd81014d1cfbb1254c74ef1774b9 should be squashed back into the commits that introduced the settings so that they are correct when they're introduced.

psifertex:
We also should be adding unit tests wherever possible. I know at least my issues that some of these address included example files that reproduced the issue. Those should be added as unit test as they're all pretty straight forward. I don't know about the other changes from the rbrans commits but it's probably worth investigating.

Weitao-Sun:
Thanks for the detailed review! I've reworked the branch to address all of it. Summary of what changed:

**Commit structure**

- Squashed the `ReadChainedImportAddend32`/`ReadChainedImportAddend64` fix into the same commit as `ReadChainedImport32`, and extracted the shared bounds-check logic into a `SymbolNameAt` helper so all three functionsshare one implementation.
- Dropped `1e00fc0` and `bb601c5` (the earlier per-function attempts that were superseded by the consolidated change).
- Squashed the settings-registration fixes back into the commits that originally introduced each setting, sothe settings are correct at the point they're introduced rather than fixed up afterward.
- Updated a few commit messages whose bodies had gone stale relative to later changes to the same commit (the Mach-O entry-limit description and the ELF/PE "endpoint validation" wording, both superseded by later revisions below).

**Mach-O**

- `ParseHeaders`: moved the length check to immediately after `sig` is read (before `sig.GetData()` is dereferenced), so the check reflects what's actually being read regardless of `imageOffset`.
- `ParseSymbolTable`: simplified the symbol-name read to a single `GetDataAt` call.
- `ParseExportTrie`: renamed `nodeCount`/`endGuard` to `visitCount`/`endIndex` (matching what they actually represent) and changed both to `size_t`.
- `InitializeHeader`: the indirect-symbol-count clamp now accounts for the universal (fat) binary's slice offset when computing how much file data remains, and skips the read entirely when the clamped count is zero (previously this could take the address of the first element of an empty vector).
- Removed the try/catch in `MachoView::Init()` that wrapped `InitializeHeader`, since `BinaryView::InitCallback` now catches everything at the C ABI boundary.
- Renamed the rebase/bind entry-limit setting to `loader.macho.maxRebaseBindEntriesMultiplier` (matching the`loader.macho.*` prefix used by the other Mach-O settings), and changed its semantics: the limit is now derived from the size of the Mach-O slice divided by its pointer size, scaled by this multiplier (default 1.0, range 0.01–10), rather than a fixed constant. The multiplier is kept rather than using the derived bound directly because each recorded relocation entry occupies more memory internally than one pointer-sized slot, so a purely 1:1 slice-size-to-entry-count mapping would still let a relatively small file drive a comparatively larger allocation; the multiplier lets this be tuned without going back to a constant that's disconnected from the file's own size.
- Reads the setting with `Get<double>(key, this)` so a per-view override is respected.
- Introduced `emitRebase`/`emitBind` helpers in `ParseRebaseTable`/`ParseDynamicTable` that own the bounds check, budget decrement, and relocation recording in one place. This also closes a gap where the single-entry opcodes (`RebaseOpcodeDoRebaseAddAddressUleb`, `BindOpcodeDoBind`, `BindOpcodeDoBindAddAddressULEB`, `BindOpcodeDoBindAddAddressImmediateScaled`) didn't consume the budget at all before — only the loop-based opcodes did.

**ELF**

- `ApplyTypesToParentStringTable`/MIPS GOT budget settings kept their `files.elf.*` prefix (already consistent with the other ELF settings).
- The MIPS GOT entry-count bound now also validates that the containing `PT_LOAD` segment's declared file extent doesn't exceed the file's actual length (previously it trusted `virtualAddress + fileSize` without an overflow check or a check against the real file size).
- `IsRangeBackedByFile` now checks every byte in the range rather than only its first and last byte, so a range that happens to span a gap between two file-backed regions is correctly rejected instead of passing basedon its endpoints alone.

**PE**

- The import lookup table loop's section-derived bound now also validates that the section's declared raw-data extent doesn't exceed the file's actual length.
- `IsRVARangeBackedByFile` now resolves and checks every RVA in the range independently (each byte finds itsown containing section and validates that specific file offset), rather than resolving only the first RVA's section and checking just the first/last byte against it — which could give a wrong answer if the range crosses a section boundary.

**Other**

- `BinaryView::InitCallback`, `OnAfterSnapshotDataAppliedCallback`, and `SaveCallback` now all catch and logexceptions at the C ABI boundary instead of only some of them.

All of the above compiles cleanly and I've re-verified each of the review comments against the current code before posting this.

Unit tests: noted the request to add tests where reproducer files exist — will follow up on that separately.

view/macho/machoview.cpp
```diff
@@ -2268,9 +2278,22 @@ bool MachoView::InitializeHeader(MachOHeader& header, bool isMainHeader, uint64_
...
 		{
-			indirectSymbols.resize(header.dysymtab.nindirectsyms);
-			reader.Seek(header.dysymtab.indirectsymoff);
-			reader.Read(&indirectSymbols[0], header.dysymtab.nindirectsyms * sizeof(uint32_t));
+			// Clamp to the number of 4-byte entries that fit between indirectsymoff and
+			// end-of-file; GetParentView()->GetLength() is the OS-reported file size and
+			// is not derived from any field inside the binary. indirectsymoff is relative
+			// to the start of this slice (reader has m_universalImageOffset as its virtual
+			// base), so that offset must be added here to compute the real file position.
+			const uint64_t readOffset = m_universalImageOffset + header.dysymtab.indirectsymoff;
```
bdash, line 2286: This exact multi-line pattern of clamping element count based on bytes remaining in the file is repeated in `GetRebaseBindEntryLimit`. Given how awkward it is with the complex ternary expression, I'd suggest a helper function so the callsites could be more descriptive. Something like:

```c++
const uint32_t indirectSymCount = std::min(header.dysymtab.nindirectsyms,
    MaxElementsRemainingInSlice(header.dysymtab.indirectsymoff, sizeof(uint32_t)));
```

Feel free to pick a better name if something else seems clearer to you.

Adding a helper would also mean that the comment about using the parent view's length and slice offset can be in one place, and they're not something each place doing this type of check needs to concern themselves with.

Weitao-Sun:
Splitting this PR by subsystem: ELF and PE are being pulled out into separate PRs, and this PR (#8303) will be narrowed to contain only the Mach-O portion (plus the two `BinaryView` C ABI exception-handling commits, since the Mach-O simplification relies on `InitCallback` catching everything).

Force-pushing this branch shortly to reflect that — flagging it first since the existing review comments are all against the Mach-O portion and should still apply.

bdash (approved):
Looks great!

Weitao-Sun:
Rebased onto latest `dev` to pick up an unrelated lint fix that was causing CI to fail — no content changes.

Issue #7435, opened 2025-09-27
Incomplete decompilation in golang binary


farvend:
**Version and Platform (required):**
- Binary Ninja Version: 5.2.8286-dev
- Edition: Non-Commercial
- OS: Windows
- OS Version: 10
- CPU Architecture: x64

**Bug Description:**
BInary NInja doesn't fully decompile the main function, decompiler output ends at `runtime.ncpu_3 = runtime.ncpu_3;`, but there is more code. The bug happens because MLIL thinks that branch at `0x004cdc90` is always false

**Steps To Reproduce:**
1. Open asd.exe
2. Go to `main.main` function
3. Open Pseudo C and MLIL (or any other lower BNIL)
4. Go to `0x004cdbe9`
5. See that in MLIL there is more code.

**Expected Behavior:**
Full decompilation of main function

**Binary:**
[asd.zip](https://github.com/user-attachments/files/22569390/asd.zip)

emesare:
Related to https://github.com/Vector35/binaryninja-api/issues/5840

Forwarding response from @bdash on slack:
>It looks like the calling convention that is applied to os/exec.Command or os/exec.(*Cmd).Output does not reflect the registers that Go uses for return values and so rdi is assumed to be unmodified from its previous value. This means the test at 0x004cdc90 is always false and one side of the conditional branch is unreachable.

Issue #5840, opened 2024-08-08
golang support


psifertex:
Right now there are many golang plugins that load debug information from golang metadata.  However they're limited by the internal support for the calling convention which is tracked in https://github.com/Vector35/binaryninja-api/issues/3340.

This issue tracks broader golang support which includes replacing the community plugins with a built-in workflow and is also dependent on #3340 to be resolved first.

ret2void:
Golang binaries are more and more! HOPE it will be a feature in next stable release!

CX330Blake:
> Golang binaries are more and more! HOPE it will be a feature in next stable release!Golang binaries 越來越多！ 希望在下一個穩定版本中成為一個功能！

Indeed. Go and Rust are taking over the world of malware lol.

psifertex:
We're actively tracking this and I expect some movement over the next few stables. Don't have a specific date yet, but definitely not forgotten.

PR #8304, opened 2026-07-03
Add Go calling convention and string recovery

seekbytes:
Based on the two rust plugins we have for Objective-C and Swift, here's a third one that provides some enhancements around golang support for binaries.

Features include:

* a MLIL workflow that sets of the correct calling convention for golang binaries. As reported in [Go internal ABI specification](https://go.googlesource.com/go/+/refs/heads/master/src/cmd/compile/abi-internal.md), we have two main calling conventions for go: one based from the stack, and one that uses only mainly registers.  First, we detect if we're dealing with a golang binary; second, if the function has a symbol for it and the symbol ends with `.abi0` we apply the cc based on the stack. At third, we try to base the heuristic with the first basic block of the routine. If we detect we're accessing a possible argument within LLIL into the stack, then we apply the stack cc otherwise the reg one.

* a HLIL workflow that adjusts the strings length passed as parameters. This is done without checking symbols (🥳 ) by simply taking advantage of the correct calling convention. A string in golang is a structure that is defined by two fields: a set of characters and the length. Given any structure with less than 3 parameters is splitted into registers/stack, we check in the HLIL call for every argument: if we're able to detect a const ptr + len within two arguments (I'm using `window(2)` which can be definitely improved). If yes, we cap the length of the datavar defined for the const ptr and we set the len correctly.

Some questions:
- I just saw I enforced `2024` as rust version, should I downgrade it?
- I defined most of the calling conventions related to golang explicitly inside this plugin (although I see that x86 has go-stack). Should I consider to move them outside of this plugin and implement them in `arch`?

Todo:
- [ ] CMakeList.txt

Should be related to:
- #5840 
- #7435

### DIFF
--- a/plugins/workflow_go/src/activities/calling_convention.rs
+++ b/plugins/workflow_go/src/activities/calling_convention.rs
@@ -1,0 +1,164 @@
+use binaryninja::architecture::{Architecture, ArchitectureExt, CoreRegister, Register};
+use binaryninja::binary_view::AnalysisContext;
+use binaryninja::low_level_il::VisitorAction;
+use binaryninja::low_level_il::expression::LowLevelILExpressionKind::Load;
+use binaryninja::low_level_il::expression::{
+    ExpressionHandler, LowLevelILExpression, LowLevelILExpressionKind, ValueExpr,
+};
+use binaryninja::low_level_il::function::{Mutable, NonSSA};
+use binaryninja::low_level_il::instruction::InstructionHandler;
+use binaryninja::low_level_il::instruction::LowLevelILInstructionKind;
+
+/// The workflow runs over the functions of a golang binary and applies the correct
+/// Go calling convention, so that register-passed arguments are recovered as
+/// parameters instead of appearing as spurious register reads.
+///
+/// Go uses two ABIs: the register-based `ABIInternal` (go1.17+) and the older
+/// stack-based `ABI0`. Binary Ninja ships neither, so we register both as custom
+/// conventions (`go-abiinternal` / `go-stack`, per architecture) and select one
+/// per function.
+///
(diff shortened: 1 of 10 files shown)


seekbytes:
Before:
go_before


After:
go_after

plafosse:
Very cool thanks for this. I'm going to assign someone to get this reviewed and potentially merged prior to the next stable. Ideally @D0ntPanic looks at it as he's most recently touched calling conventions and he's working on some stuff that should help with display of non-null terminated strings.

plugins/workflow_go/src/activities/calling_convention.rs
```diff
@@ -0,0 +1,164 @@
...
+            // thunk functions are trampolines for the external functions and for that we
+            // do not require to set a specific golang calling convention
+            if Self::is_thunk(ctx) {
+                return;
+            }
+        }
+
+        // 1) if the function has a symbol and ends with ".abi0", then we apply the go-stack
+        // convention
+        let name = func.symbol().full_name();
```
emesare: Since Function::symbol always returns a symbol we should maybe consider Function::defined_symbol

https://github.com/Vector35/binaryninja-api/blob/beee8650a717b65a047612bca530dbc55499ec08/rust/src/function.rs#L386

then down where you do "n if n.is_empty() || n.starts_with("sub_") you can instead check "does this function have a symbol" instead of checking  for "sub_".

seekbytes: Done

psifertex:
Unfortunately we weren't able to get this into the K release as you can tell from us bumping it but we're still very much looking forward to merging this. It just needs to be part of a much broader effort we plan to make on getting better golang support that's happening post-6.0. We do appreciate the PR!

PR #8314, opened 2026-07-13
Add the framework for BNIL emulator and implement LLIL emulator

### LINKED ISSUE #1212: IL emulator

xusheng6:
This PR adds the emulator framework and implements LLIL emulation. It closes https://github.com/Vector35/binaryninja-api/issues/1212 and https://github.com/Vector35/debugger/issues/320

The emulator is implemented as an open-source plugin in the API repo to enable 3rd party inspection and contribution. It may also later serve as a "spec" of the definition of our IL

The emulator does its only C FFI and Python/C++ API binding in the same way as BN core and the debugger. 

The scope of this PR is to get the core LLIL emulation done, i.e., we should be able to emulate the LLIL instructions and functions properly. Full system emulation and API emulation is out of scope, though some of them are added to make it possible to verify the result of the emulation, e.g., things can be printed to stdout to validate the emulation

This PR is also accompanied by some core changes (https://github.com/Vector35/binaryninja/pull/1656) the a BNIL debug adapter (https://github.com/Vector35/debugger/tree/test_emulator) which enables the user to visually inspect the behavior of the emulator. The core changes would need to land with this PR. And I would also prefer to have the debugger changes land together because 1) it changes how the intx headers are shared between the emulator and the debugger, 2) it is the more straightforward way to inspect the emulator's status

The tests cover every LLIL instruction we have, plus a few small binaries that does computations (hash, encryption) (to be added)

The IL emulator will be marked as experimental and disabled by default.

Expected follow-up work:

1. Polish the handling of different calling conventions during function calls/returns (e.g., Rust, go binaries)
2. Emulate non-trivial programs and identify new issues

Note: this PR also enables the boss plugin (https://github.com/xusheng6/boss) which uses emulation to detect obfuscated strings in the same way as flare-floss (https://github.com/mandiant/flare-floss)


plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3753 @@
...
+void LLILEmulator::SetArgument(size_t index, const intx::uint512& value)
+{
+	if (!m_view || !m_arch)
+		return;
+
+	Ref<Platform> platform = m_view->GetDefaultPlatform();
+	if (!platform)
+		return;
+
+	Ref<CallingConvention> cc = platform->GetDefaultCallingConvention();
```
emesare: This does not seem correct, and also seems to be in conflict with what the docs say

> Arguments are placed using the function's default calling convention:

Yet this is using the default platform and default calling convention, not the function arg list or cc.

binaryninjaapi.h
```diff
@@ -31,6 +31,23 @@
 #include "binaryninjacore.h"
 #include "exceptions.h"
 
+// intx provides the wide-integer type used by the emulator plugin's API. If a
```
bdash: I'm confused why this is necessary in binaryninjaapi.h when this header does not provide the emulator's API. Why isn't this in emulatorapi.h instead?

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3753 @@
...
+
+	auto argRegs = cc->GetIntegerArgumentRegisters();
+	if (index < argRegs.size())
+	{
+		SetRegister(argRegs[index], value);
+		return;
+	}
+
+	// Stack argument
+	size_t addrSize = m_arch->GetAddressSize();
```
emesare, line 166: We need to be careful using address size for stack slot size given those are two completely different things, but incidentally this will probably be fine assuming we never touch the platform GetAddressSize which on something like the x32 abi for linux x86-64 will report as 4 bytes instead of the 8 byte stack slot size.

binaryninjacore.h
```diff
@@ -9428,6 +9427,7 @@ extern "C"
 	BINARYNINJACOREAPI BNPossibleValueSet BNPossibleValueSetNegate(const BNPossibleValueSet* object, size_t size);
 	BINARYNINJACOREAPI BNPossibleValueSet BNPossibleValueSetNot(const BNPossibleValueSet* object, size_t size);
 
+
```
bdash: Seems unnecessary.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3753 @@
...
+		intx::uint512 toMask = (toSize >= 64) ? ~intx::uint512(0) : (intx::uint512(1) << (toSize * 8)) - 1;
+		result |= toMask & ~fromMask;
+	}
+	return MaskToSize(result, toSize);
+}
+
+
+BNEndianness LLILEmulator::GetEndianness() const
+{
+	if (m_arch)
```
emesare: How is the emulator suppose to function without an architecture? I feel like we should hold that as an invariant instead of trying to half-way support having no architecture in some random member functions.

emesare (changes requested):

plugins/emulator/api/python/generator.cpp
```diff
@@ -0,0 +1,614 @@
...
+
+#include <stdio.h>
+#include <inttypes.h>
+#include "binaryninjaapi.h"
+
+using namespace BinaryNinja;
+using namespace std;
+
+
+map<string, string> g_pythonKeywordReplacements = {
```
bdash, line 27: I know we've done this elsewhere, but do we really need a fifth copy of this in the API repository? Every existing copy of the file is already different, and I have no idea if the differences are significant.

Can we not use the executable produced by python/generator.cpp as part of building the Python bindings?

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3753 @@
...
+	{
+		uint32_t reg = expr.GetRawOperandAsRegister(0);
+		if (LLIL_REG_IS_TEMP(reg))
+			return MaskToSize(GetTempRegister(LLIL_GET_TEMP_REG_INDEX(reg)), sz);
+		return MaskToSize(GetRegister(reg), sz);
+	}
+
+	case LLIL_REG_SPLIT:
+	{
+		uint32_t hi = expr.GetRawOperandAsRegister(0);
```
bdash: Is there a reason this code is using `GetRawOperandAs…` everywhere rather than using the typed accessors? We've had bugs in the past from code like this where operand indices were mixed up.

plugins/emulator/core/llilemulator.h
```diff
@@ -0,0 +1,189 @@
...
+		bool IsLogLibcCalls() const;
+		void SetNopUnknownExternals(bool enabled);
+		bool IsNopUnknownExternals() const;
+
+		void Reset() override;
+
+		// State serialization
+		std::string SaveState() const;
+		bool LoadState(const std::string& json, BinaryNinja::BinaryView* view);
+		BinaryNinja::BinaryView* GetView() const { return m_view; }
```
emesare: It seems ill advised to pass back a raw pointer to the view in a public member function especially because the view member itself is ref counted.

plugins/emulator/core/llilemulator.h
```diff
@@ -0,0 +1,189 @@
...
+		void WriteReturnValue(uint64_t value);
+		intx::uint512 ReadArgumentWide(size_t index);
+		void WriteReturnValueWide(const intx::uint512& value);
+		std::string ResolveCallTargetName(uint64_t addr);
+		static std::string NormalizeLibcName(const std::string& name);
+		bool HandleBuiltinCall(uint64_t dest);
+		bool HandleUnknownCall(uint64_t dest);
+		void EnsureHeapMapped();
+
+		// Stub implementations
```
emesare, line 114: IMO we should not be writing stub functions in the body of the emulator like this, I would personally try and move these out into platform, dll, specific files

binaryninjacore.h
emesare: erroneous formatting?

binaryninjaapi.h
```diff
@@ -23338,6 +23354,7 @@ namespace BinaryNinja {
 		std::optional<DerivedString> RecognizeConstantData(
 			const HighLevelILInstruction& instr) override;
 	};
+
```
emesare: erroneous formatting?

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3753 @@
...
+			SetStopReason(ILEmulatorStopReason::Error, "division by zero");
+			return 0;
+		}
+		return MaskToSize(left / right, sz);
+	}
+
+	case LLIL_DIVS:
+	{
+		// For signed division, work with 64-bit values since intx doesn't provide signed types
+		int64_t left = static_cast<int64_t>(static_cast<uint64_t>(SignExtend(EvalExpr(expr.GetRawOperandAsExpr(0)), sz, 8)));
```
bdash: It is surprising that a bunch of operations on signed integers are implemented such that they silently do the wrong thing for sizes > 8.

plugins/emulator/api/emulatorapi.h
```diff
@@ -0,0 +1,155 @@
...
+
+#include "binaryninjaapi.h"
+#include "ffi.h"
+#include <functional>
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+namespace BinaryNinja
```
plafosse: Seems odd for this to live in the BinaryNinja namespace if its a plugin.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3753 @@
...
+		SetStopReason(ILEmulatorStopReason::CallHook,
+			fmt::format("tailcall to 0x{:x}", dest));
+		return;
+	}
+
+	case LLIL_RET:
+	{
+		// Evaluate the return address expression for side effects
+		// (pops RA from stack on x86, reads LR on ARM — no-op)
+		EvalExpr(instr.GetRawOperandAsExpr(0));
```
bdash: Should this check `m_stopReason` before continuing on?

The pattern of having to manually check `m_stopReason` seems rather error-prone.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3753 @@
...
+		return intx::uint512((flagVal >> bitIndex) & 1);
+	}
+
+	// --- Two-operand arithmetic ---
+	case LLIL_ADD:
+	{
+		intx::uint512 left = EvalExpr(expr.GetRawOperandAsExpr(0));
+		intx::uint512 right = EvalExpr(expr.GetRawOperandAsExpr(1));
+		intx::uint512 result = MaskToSize(left + right, sz);
+		m_lastArithmetic = {MaskToSize(left, sz), MaskToSize(right, sz), result, sz, LLIL_ADD, true};
```
bdash: I'm a little confused by the `MaskToSize` calls here. The add is evaluated here without masking the operands, but is recorded in `m_lastArithmetic` with them masked to `sz`.

Other instructions seem to mask one or more operands before they're used. Sometimes things are masked to `sz`, other times to `sz / 2`. It's hard to tell when reading this code why different instructions do things differently.

xusheng6:
We also want to add at least one test for each LLIL instruction

binaryninjacore.h
xusheng6: Fixed in 9b0180cf5 — removed the stray extra blank line.

xusheng6: Fixed in 0d1183db4 — removed the stray blank line before the namespace close.

binaryninjaapi.h
xusheng6: Fixed by removing it

plugins/emulator/core/llilemulator.cpp
xusheng6: Fixed in 8a42cba00 — SetArgument now derives the calling convention from the function being emulated (via its LLIL function) and only falls back to the platform default when the function has none.

plugins/emulator/core/llilemulator.cpp
xusheng6: Yeah I think it is fine to leave to as is, though I still added a comment for it in 8a42cba00

plugins/emulator/core/llilemulator.cpp
xusheng6: I think the intention is the emulator does not work without an arch. This is just AI being too tempting to add a nullptr check on everything. 

Fixed in 53fab4b83 — a valid architecture is now a hard invariant: BNCreateLLILEmulator* reject creation (with a logged error) when the view/IL has no architecture, and the scattered null-arch guards are gone.

plugins/emulator/core/llilemulator.h
xusheng6: Fixed in 02fd0ac4a — GetView() now returns Ref<BinaryView> instead of a raw pointer.

plugins/emulator/core/llilemulator.h
xusheng6: Agree in principle; leaving this for a follow-up since relocating the ~40 libc/Win32 stubs into platform/DLL-specific files is a larger restructure than the rest of this pass. 

This is also somewhat outside of the scope of the BNIL emulator for now -- I mainly want the emulator to be able to emulate the BNIL, and these are just added so that it can run some small programs as a test

plugins/emulator/api/emulatorapi.h
xusheng6: Fixed in 067252aa2 — moved the public C++ API to a BinaryNinjaEmulatorAPI namespace (mirroring BinaryNinjaDebuggerAPI); the debugger adapter follows the rename in 6bd1699.

plugins/emulator/api/python/generator.cpp
xusheng6: Not changed yet. The four plugin generators (warp/sharedcache/kernelcache/emulator) each keep their own copy, and the emulator's differs substantively from the core python/generator.cpp (~389 lines), so reusing the core executable would need a cross-plugin generator refactor rather than a local change. Happy to take that on separately if we want to unify all of them.

plugins/emulator/core/llilemulator.cpp
xusheng6: Fixed in 75fc5039f — converted ~130 sites in EvalExpr/ExecuteCurrentInstruction to the typed accessors (GetLeftExpr/GetRightExpr/GetCarryExpr, GetDestRegister, GetHighRegister/GetLowRegister, etc.), each verified to read the same operand index. Only LLIL_EXTERN_PTR is left raw (no equivalent typed constant+offset pair). Full test suite passes.

plugins/emulator/core/llilemulator.cpp
xusheng6: Fixed in dbbd4be83 — the CALL/CALL_STACK_ADJUST/TAILCALL cases now check m_stopReason after the builtin-stub/hook handling and bail if a stop was requested, instead of falling through into the call.

plugins/emulator/api/python/generator.cpp
bdash: How does the emulator's copy of this differ and why?

plugins/emulator/core/llilemulator.cpp
xusheng6: Fixed in 8d069d987 — added wide-signed helpers so DIVS/MODS/MULS_DP/CMP_S*/MINS/MAXS operate on the full 512-bit value instead of truncating through int64_t; this also fixes the INT_MIN / -1 case (it wraps instead of trapping). The double-precision divides were fixed in 7864ee307.

plugins/emulator/core/llilemulator.cpp
xusheng6: Fixed in 868ce1bf9 — the basic arithmetic ops now mask both operands to sz once up front and use those masked values consistently (computation + flag context), with a comment explaining why the double-precision ops mask to sz/2.

I think we have single and double precision multiplications, and for a double precision one, the result of the calculation is sz wide, so we need to mask the inputs to sz/2. I am not entirely sure this is correct, my memory about these double precision things are a bit vague. @plafosse do you think this is the right way to do it?

plugins/emulator/api/python/generator.cpp
xusheng6: I had a closer look and it is more nuanced than what I believed. To start with, the debugger version of the generator.cpp contains strings like "BNDebuggerFreeString", "BNDebugger", etc (see https://github.com/Vector35/debugger/blob/f52fbef2ac734a23c673e98a1e42e9fffa6e8ba0/api/python/generator.cpp#L445), so I assumed each of them would have a few different ones and it might be challenging to merge them. However, I noticed that the ones added after the debugger (i.e., warp, sharedcache, kernel cache), as well as the emulator one we are looking at, also contains the debugger related string. These are definitely not required for them, yet it appears to be working, so there should definitely an effort to clean them up, and then hopefully merge all of the copies

That said, I do not think it is a good idea to deal with it in this PR, so I created https://github.com/Vector35/binaryninja/issues/1673 to track it seperately

binaryninjaapi.h
xusheng6: Before this PR, the debugger already uses intx and has a copy of it. Now, since the emulator also uses it, I moved the intx headers to the vendor folder of binaryninja-api, and have the emulator and the debugger both use it. However, this leaves to a build failure related to the definition of MIN and MAX. Previously, the debugger has a `#define NOMINMAX` and that is why the debugger can compile just fine. I was unaware of this and made the chagnes in binaryninjaapi.h. I am now fixing the build in a different way in https://github.com/Vector35/binaryninja-api/commit/fd58c1793453b85da6eb520ce1fc9aad0f20c9fd and https://github.com/Vector35/debugger/commit/cd65c42059176fdbc879497c423efbe4809d4cf0

plugins/emulator/core/llilemulator.cpp
xusheng6: I believe this fixed the issue and I marked it as resolved, but feel free to let me know if you think more changes are needed

plugins/emulator/test/emulator_test.py
xusheng6, line 1: Per Peter's comment, we need to have at least one test for each LLIL instruciton

xusheng6:
@emesare @bdash I have updated the PR according to your review comments, please have a look at them and see if these are adequate, thanks!

docs/guide/emulator.md
```diff
@@ -0,0 +1,353 @@
+# BNIL Emulator — Python API Guide
```
emesare, line 1: Typically we have reserved this kind of content for the developer guide.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3883 @@
...
+	m_callStack.clear();
+}
+
+
+void LLILEmulator::SetArgument(size_t index, const intx::uint512& value)
+{
+	if (!m_view)
+		return;
+
+	// Place arguments using the calling convention of the function being emulated, falling
```
emesare: How does this work with non default argument locations?

Image

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3883 @@
...
+
+
+void LLILEmulator::SetArguments(const std::vector<uint64_t>& values)
+{
+	for (size_t i = 0; i < values.size(); i++)
+		SetArgument(i, values[i]);
+}
+
+
+intx::uint512 LLILEmulator::MaskToSize(const intx::uint512& value, size_t size)
```
emesare: Why not implement this in the base class, or in a helper file? I more so am just curious why this is tied to the LLIL emulator specifically.

Also goes for the similar functions below

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3883 @@
...
+uint8_t LLILEmulator::GetFlag(uint32_t flag) const
+{
+	auto it = m_flags.find(flag);
+	return (it != m_flags.end()) ? it->second : 0;
+}
+
+
+void LLILEmulator::SetFlag(uint32_t flag, uint8_t value)
+{
+	m_flags[flag] = value ? 1 : 0;
```
emesare: why not just use a bool here

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3883 @@
...
+
+	case LLIL_CALL:
+	{
+		uint64_t dest = static_cast<uint64_t>(EvalExpr(instr.GetDestExpr()));
+		if (m_stopReason != ILEmulatorStopReason::Running)
+			return;
+
+		// 1) Try built-in stubs first
+		if (m_builtinLibcStubs && HandleBuiltinCall(dest))
+			break;
```
emesare, line 1911: the mixed usage of break and return in this function body is kinda confusing and might lead to unexpected bugs in the future, pick one and stick with it

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3883 @@
...
+uint64_t LLILEmulator::ReadArgument(size_t index)
+{
+	if (!m_view)
+		return 0;
+
+	Ref<Platform> platform = m_view->GetDefaultPlatform();
+	if (!platform)
+		return 0;
+
+	Ref<CallingConvention> cc = platform->GetDefaultCallingConvention();
```
emesare: SetArgument was updated to read the function calling convention, this was not.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3883 @@
...
+		return "";
+
+	Ref<Symbol> sym = m_view->GetSymbolByAddress(addr);
+	if (!sym)
+		return "";
+
+	BNSymbolType type = sym->GetType();
+	if (type == ImportedFunctionSymbol || type == FunctionSymbol
+		|| type == LibraryFunctionSymbol || type == ExternalSymbol)
+		return std::string(sym->GetShortName());
```
emesare: GetShortName returns a string already.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3883 @@
...
+	return (this->*(it->second))(dest);
+}
+
+
+bool LLILEmulator::HandleUnknownCall(uint64_t dest)
+{
+	if (!m_nopUnknownExternals)
+		return false;
+
+	std::string name = ResolveCallTargetName(dest);
```
emesare: I would make it return std::optional<string>

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3883 @@
...
+
+// ============================================================================
+// LLILEmulator
+// ============================================================================
+
+LLILEmulator::LLILEmulator(BinaryView* backingView) :
+	m_view(backingView), m_il(nullptr), m_arch(nullptr)
+{
+	if (m_view)
+		m_arch = m_view->GetDefaultArchitecture();
```
emesare, line 20: For binaries with multiple architecture (thumb) we should probably add some warning.

xusheng6:
We have also decided to mark the emulator as experimental (which we already did), and disable it in binja be default

docs/guide/emulator.md
xusheng6: I agree, but the only way to use the emulator is through its API (well except for using the BNIL debug adapter), so this should be *fine*

plugins/emulator/core/llilemulator.cpp
xusheng6: Fixed in 8ccc93966. Now it places the incoming arguments into the correct register/stack location when it can, and falls back to the default calling convention's location otherwise

plugins/emulator/core/llilemulator.cpp
xusheng6: Moved to the base class in 54bd3d68b

plugins/emulator/core/llilemulator.cpp
xusheng6: Changed the storage and the setter/getter to use a bool for flags. 7634629f2

plugins/emulator/core/llilemulator.cpp
xusheng6: Fixed in 1735af004

plugins/emulator/core/llilemulator.cpp
xusheng6: Fixed in 6143af047. Created a helper `ResolveCallingConvention` and use it for a few places

plugins/emulator/core/llilemulator.cpp
xusheng6: Fixed in 84abd97d6. I guess this is a leftover from the emulator being originally in the core

plugins/emulator/core/llilemulator.cpp
xusheng6: Changed to std::optional in 84abd97d6

plugins/emulator/core/llilemulator.cpp
xusheng6: Added. Rather than scanning every function up front, it now warns lazily: the architecture switches (SetEntryPoint/EnterFunction/tailcall/state-restore) go through a SetActiveArchitecture helper that logs a one-time warning when the arch being emulated differs from the binary's default. 609d0acdd

Moving forward, we might be able to properly support arch switching code, though that would require more work and I will leave it as future work

plugins/emulator/core/ilemulator.cpp
```diff
@@ -0,0 +1,964 @@
...
+			fmt::format("failed to write {} bytes at 0x{:x}", size, addr));
+	}
+}
+
+
+intx::uint512 ILEmulator::MaskToSize(const intx::uint512& value, size_t size)
+{
+	if (size >= 64)
+		return value;
+	if (size == 0)
```
bdash, line 490: I think this may give incorrect behavior for booleans that are created from comparisons against flags (`LLIL_CMP_*`) as they can have size 0, but non-zero values when true.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+	case LLIL_REG_SPLIT:
+	{
+		uint32_t hi = expr.GetHighRegister();
+		uint32_t lo = expr.GetLowRegister();
+		intx::uint512 hiVal = LLIL_REG_IS_TEMP(hi)
+			? GetTempRegister(LLIL_GET_TEMP_REG_INDEX(hi)) : GetRegister(hi);
+		intx::uint512 loVal = LLIL_REG_IS_TEMP(lo)
+			? GetTempRegister(LLIL_GET_TEMP_REG_INDEX(lo)) : GetRegister(lo);
+		size_t loSize = LLIL_REG_IS_TEMP(lo) ? sz / 2 : m_arch->GetRegisterInfo(lo).size;
+		return MaskToSize((hiVal << (loSize * 8)) | loVal, sz);
```
bdash: Is dividing the low size by 2 and masking the output to `sz` correct here? Core's equivalent code seems to think that the output size should be `instr.size * 2`

 I'd guess `LLIL_SET_REG_SPLIT` needs to be changed if this is changed.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+	}
+
+	case LLIL_FLAG:
+		return intx::uint512(GetFlag(expr.GetSourceFlag()));
+
+	case LLIL_FLAG_BIT:
+	{
+		uint8_t flagVal = GetFlag(expr.GetSourceFlag());
+		size_t bitIndex = expr.GetBitIndex();
+		return intx::uint512((flagVal >> bitIndex) & 1);
```
bdash: This looks like it is incorrect. `GetFlag` returns a boolean, so any non-zero `bitIndex` will be zero. Core's equivalent code performs a left shift when handling `LLIL_FLAG_BIT_SSA`.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+			z = val;
+			break;
+		case NegativeSignFlagRole:
+			s = val;
+			break;
+		case PositiveSignFlagRole:
+			s = val ? 0 : 1;
+			break;
+		case CarryFlagRole:
+		case CarryFlagWithInvertedSubtractRole:
```
bdash: Is treating `CarryFlagRole` the same as `CarryFlagRoleWithInvertedSubtractRole` correct? Core distinguishes between them, and inverts the condition flag for certain flags.

plugins/emulator/core/ilemulator.cpp
```diff
@@ -0,0 +1,964 @@
...
+			SetStopReason(ILEmulatorStopReason::Halt, "pre-instruction hook stopped execution");
+			break;
+		}
+
+		size_t prevIndex = m_instrIndex;
+		ExecuteCurrentInstruction();
+		m_instructionsExecuted++;
+
+		// If ExecuteCurrentInstruction didn't change the index, advance to next
+		if (m_instrIndex == prevIndex && m_stopReason == ILEmulatorStopReason::Running)
```
bdash: Is this check correct? `m_instrIndex` appears to be an index into the current function's instructions, but operations like calls or returns can change the function while keeping the instruction index the same.

This exact same check appears to be duplicated in a bunch of places across two different files, which feels like it is asking for incomplete changes if this is modified in the future.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+		return;
+	}
+
+	case LLIL_TAILCALL:
+	{
+		uint64_t dest = static_cast<uint64_t>(EvalExpr(instr.GetDestExpr()));
+		if (m_stopReason != ILEmulatorStopReason::Running)
+			return;
+
+		if (m_builtinLibcStubs && HandleBuiltinCall(dest))
```
bdash: When a built-in call or call hook handles a tail call, what updates the instruction index + current function? I'm not sure if there's logic missing here that should handle it, or if it's already handled elsewhere and I missed it.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+	}
+}
+
+
+Ref<CallingConvention> LLILEmulator::ResolveCallingConvention()
+{
+	if (!m_view)
+		return nullptr;
+
+	// Use the calling convention of the function being emulated, falling back to the platform
```
bdash, line 2219: This seems problematic given that this is used when finding arguments within stub functions? Consider code within a thiscall function that is calling a typical cdecl function. Using the calling convention from the calling thiscall function will give the wrong argument locations.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+	}
+
+	case LLIL_CALL:
+	{
+		uint64_t dest = static_cast<uint64_t>(EvalExpr(instr.GetDestExpr()));
+		if (m_stopReason != ILEmulatorStopReason::Running)
+			return;
+
+		// 1) Try built-in stubs first
+		if (m_builtinLibcStubs && HandleBuiltinCall(dest))
```
bdash, line 1910: Is using built-in stubs unconditionally correct? For statically linked binaries this will prefer the built-in stub version of functions over implementations that exist in the binary.

Additionally, this logic for resolving the target of calls / handling calls is duplicated in three different places. It should be shared.

plugins/emulator/core/ilemulator.cpp
```diff
@@ -0,0 +1,964 @@
...
+}
+
+
+ILEmulatorStopReason ILEmulator::StepN(size_t n)
+{
+	m_stopReason = ILEmulatorStopReason::Running;
+	m_stopMessage.clear();
+	m_stopRequested.store(false, std::memory_order_relaxed);
+
+	for (size_t i = 0; i < n && m_stopReason == ILEmulatorStopReason::Running; i++)
```
bdash, line 369: There's a bunch of logic within the body of `ILEmulator::Run`'s loop that is absent here. Is that intentional?

It seems like `ILEmulator::Run`, `ILEmulator::StepN`, and `LLILEmulator::StepOver` should really be sharing code rather than reimplementing very similar logic for driving execution.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+
+bool LLILEmulator::StubFread(uint64_t)
+{
+	// fread(buf, size, count, stream) — reads size*count bytes
+	uint64_t bufAddr = ReadArgument(0);
+	uint64_t elemSize = ReadArgument(1);
+	uint64_t elemCount = ReadArgument(2);
+	// arg 3 (stream) is ignored — always treated as stdin
+
+	size_t totalBytes = (size_t)(elemSize * elemCount);
```
bdash, line 3613: Should we clamp this?

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+	doc.Accept(writer);
+	return std::string(sb.GetString(), sb.GetSize());
+}
+
+
+bool LLILEmulator::LoadState(const std::string& json, BinaryView* view)
+{
+	rapidjson::Document doc;
+	doc.Parse(json.c_str());
+	if (doc.HasParseError() || !doc.IsObject())
```
bdash, line 762: RapidJSON is configured to throw exceptions on parse errors, so this `HasParseError` check is dead. You need to use try/catch instead.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+	m_callStack.clear();
+	if (doc.HasMember("callStack") && doc["callStack"].IsArray() && view)
+	{
+		for (auto& frame : doc["callStack"].GetArray())
+		{
+			if (!frame.IsObject())
+				continue;
+
+			CallFrame cf;
+			cf.arch = nullptr;
```
bdash, line 867: Creating call frames with a null architecture seems like it'll crash when this is used later on? We might need to bail out of if we're unable to resolve the function below.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+
+			if (frame.HasMember("funcAddr"))
+			{
+				uint64_t funcAddr = strtoull(frame["funcAddr"].GetString() + 2, nullptr, 16);
```
bdash, line 871: I'm a little confused by what the `+ 2` is doing here.

plugins/emulator/api/ilemulator.cpp
```diff
@@ -0,0 +1,555 @@
...
+#include "ffi.h"
+#include "emulatorapi.h"
+
+using namespace BinaryNinja;
+using namespace BinaryNinjaEmulatorAPI;
+
+
+LLILEmulator::LLILEmulator(Ref<BinaryView> view)
+{
+	m_object = BNCreateLLILEmulatorForView(view->GetObject());
```
bdash, line 31: I think `BNCreateLLILEmulatorForView` can return null on error, but I don't see any way for a caller to detect that has happened. Any use of this object after a null return here would crash.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+	case LLIL_NOP:
+		break;
+
+	case LLIL_SET_REG:
+	{
+		uint32_t reg = instr.GetDestRegister();
+		m_lastArithmetic.valid = false;
+		intx::uint512 val = EvalExpr(instr.GetSourceExpr());
+		if (m_stopReason != ILEmulatorStopReason::Running)
+			return;
```
bdash, line 1776: These early returns seem problematic. `EvalExpr` can have side-effects (`LLIL_POP` jumped out at me). If we return after an `EvalExpr` but before performing the rest of the instruction, only some of the instruction's effects are applied.

I'm not entirely clear what happens if the emulator is resumed after one of these early returns (or if it can be resumed from that state?). It seems like it may reevaluate the same instruction again, in which case any side-effects would be performed a second time.

plugins/emulator/core/ilemulator.cpp
```diff
@@ -0,0 +1,964 @@
...
+		m_instructionsExecuted++;
+
+		if (m_instrIndex == prevIndex && m_stopReason == ILEmulatorStopReason::Running)
+			m_instrIndex++;
+	}
+
+	m_currentAddress = GetCurrentInstructionAddress();
+
+	if (m_stopReason == ILEmulatorStopReason::Running)
+		SetStopReason(ILEmulatorStopReason::Halt);
```
bdash, line 378: Using `Halt` to mean both "ran off end of IL" and "successfully executed exactly N instructions as requested" feels confusing. How is a client to distinguish between those? One seems like a condition they can happily continue execution from, while the other very much isn't.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+}
+
+
+// ============================================================================
+// Flag condition evaluation
+// ============================================================================
+
+intx::uint512 LLILEmulator::EvalFlagCondition(BNLowLevelILFlagCondition cond, uint32_t semClass)
+{
+	auto requiredFlags = m_arch->GetFlagsRequiredForFlagCondition(cond, semClass);
```
bdash, line 1589: Is this reachable on aarch64 or other similar architectures?

`Architecture::GetFlagsRequiredForFlagCondition` appears to be optional. aarch64 does not implement it, so this will return an empty set and presumably flags will always evaluate to zero.

I'm not familiar enough with how flags work on architectures to know what this should be doing instead.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+
+	// --- Flag conditions (Lifted IL) ---
+	case LLIL_FLAG_COND:
+	{
+		BNLowLevelILFlagCondition cond = expr.GetFlagCondition();
+		uint32_t semClass = expr.GetSemanticFlagClass();
+		return EvalFlagCondition(cond, semClass);
+	}
+
+	case LLIL_FLAG_GROUP:
```
bdash, line 1550: Core's handling of this involves `GetSemanticFlagGroupLowLevelIL`, which I don't see called anywhere here. Is that expected?

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+		m_lastArithmetic.valid = false;
+		intx::uint512 val = EvalExpr(instr.GetSourceExpr());
+		if (m_stopReason != ILEmulatorStopReason::Running)
+			return;
+		if (LLIL_REG_IS_TEMP(reg))
+			SetTempRegister(LLIL_GET_TEMP_REG_INDEX(reg), val);
+		else
+			SetRegister(reg, val);
+		if (instr.flags != 0)
+			WriteFlags(instr.flags);
```
bdash, line 1782: Is it correct to only handle flag writes only on instructions, and not on their subexpressions? Is it correct to only write flags on these two instructions (`LLIL_SET_REG`, `LLIL_SET_REG_SPLIT`)?

The fact `m_lastArithmetic` is set only by some subexpressions makes me think `WriteFlags` can end up settings flags based on the wrong expression.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+		SetStopReason(ILEmulatorStopReason::CallHook,
+			fmt::format("tailcall to 0x{:x}", dest));
+		return;
+	}
+
+	case LLIL_RET:
+	{
+		// Evaluate the return address expression for side effects
+		// (pops RA from stack on x86, reads LR on ARM — no-op)
+		EvalExpr(instr.GetDestExpr());
```
bdash, line 2068: Should we respect the return address that we evaluate here, rather than discarding it?

Should this be checking `m_stopReason`? Having to explicitly check it after every call to `EvalExpr` seems pretty error prone.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+
+	// Fallback: use the next LLIL instruction's address
+	if (m_instrIndex + 1 < m_il->GetInstructionCount())
+		return m_il->GetInstruction(m_instrIndex + 1).address;
+
+	return callAddr;
+}
+
+
+void LLILEmulator::Reset()
```
bdash, line 534: Exactly what is this intended to reset? It looks like the IL function and architecture are left intact.

plugins/emulator/core/ilemulator.cpp
```diff
@@ -0,0 +1,964 @@
...
+
+
+ILEmulatorStopReason ILEmulator::Run()
+{
+	m_stopReason = ILEmulatorStopReason::Running;
+	m_stopMessage.clear();
+	m_stopRequested.store(false, std::memory_order_relaxed);
+
+	// Remember the address we're resuming from so we skip the breakpoint on the
+	// very first instruction (avoids re-hitting the same breakpoint on resume).
```
bdash, line 311: Multiple IL instructions can be created when lifting a single native instruction. Is breaking after the first IL instruction correct / expected, or should it continue until the address changes?

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+
+	// Stop state
+	if (doc.HasMember("stopReason"))
+		m_stopReason = (ILEmulatorStopReason)doc["stopReason"].GetInt();
+	if (doc.HasMember("stopMessage"))
+		m_stopMessage = doc["stopMessage"].GetString();
+
+	m_view = view;
+
+	return true;
```
bdash, line 926: There's a lot of logic above that clears members but skips doing work if members don't exist in the JSON. Does the object end up in a coherent state with all of that data missing?

Should this be validating that the view we're loading the state into is the same (whatever that means) as the one it was saved for? Or at the least for the same platform/architecture?

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+		m_instructionsExecuted = doc["instructionsExecuted"].GetUint64();
+
+	// Registers
+	m_registers.clear();
+	if (doc.HasMember("registers") && doc["registers"].IsObject())
+	{
+		for (auto it = doc["registers"].MemberBegin(); it != doc["registers"].MemberEnd(); ++it)
+		{
+			uint32_t regId = std::stoul(it->name.GetString());
+			m_registers[regId] = HexStringToUint512(it->value.GetString());
```
bdash, line 783: Are an architecture's register IDs really considered to be stable? Should this be serializing register names instead?

plugins/emulator/core/ilemulator.cpp
```diff
@@ -0,0 +1,964 @@
...
+{
+	m_stopRequested.store(true, std::memory_order_relaxed);
+}
+
+
+void ILEmulator::Reset()
+{
+	m_memory.Reset();
+	m_breakpoints.clear();
+	m_instructionsExecuted = 0;
```
bdash, line 677: Is there no way to reset `m_instructionsExecuted` short of resetting the emulator entirely? It seems that once you hit this limit the emulator will be wedged and refuse to run any further until it is reset.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+		// %n: reject
+		if (conv == 'n')
+		{
+			LogWarn("BNIL Emulator: printf %%n rejected (attempted write via format string)");
+			argIndex++;  // consume the argument
+			continue;
+		}
+
+		// Read the argument value
+		uint64_t argVal = ReadArgument(argIndex++);
```
bdash, line 3289: This assumes arguments are always integers, and will give incorrect behavior for subsequent arguments for format strings containing `%f` or similar.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+			for (size_t i = 0; i < maxRead; i++)
+			{
+				uint8_t byte = 0;
+				if (m_memory.Read(&byte, argVal + i, 1) != 1 || byte == 0)
+					break;
+				strVal.push_back((char)byte);
+			}
+
+			// Format with %.*s using known length for safety
+			std::string finalFmt = specFmt + ".*s";
```
bdash, line 3339: This will give incorrect behavior if the format specifier already includes a precision specifier.

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+}
+
+
+std::string LLILEmulator::NormalizeLibcName(const std::string& name)
+{
+	std::string result = name;
+
+	// Strip leading underscore (macOS/Mach-O convention)
+	if (result.size() > 1 && result[0] == '_')
+		result = result.substr(1);
```
bdash, line 2296: Won't this cause stub entries like `__printf_chk` to be unreachable?

plugins/emulator/core/llilemulator.cpp
```diff
@@ -0,0 +1,3893 @@
...
+
+	// Strip j_ prefix (PLT/thunk convention)
+	if (result.size() > 2 && result.substr(0, 2) == "j_")
+		result = result.substr(2);
+
+	return result;
+}
+
+
+const std::unordered_map<std::string, LLILEmulator::StubFn>& LLILEmulator::GetStubTable()
```
bdash, line 2306: I don't see any reason why these stubs should live in the LLIL emulator. AFAICT all they need is the ability to read arguments, read/write memory, and return a value, which could be accomplished via an abstract interface that `LLILEmulator` implements.

If these didn't live on `LLILEmulator` then it'd be possible to write unit tests for them, which is currently quite challenging to do.

plugins/emulator/core/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+set_target_properties(emulatorcore PROPERTIES
+	CXX_STANDARD 20
+	CXX_VISIBILITY_PRESET hidden
+	CXX_STANDARD_REQUIRED ON
+	VISIBILITY_INLINES_HIDDEN ON
+	POSITION_INDEPENDENT_CODE ON
+)
+
+if(BN_INTERNAL_BUILD)
+	set(LIBRARY_OUTPUT_DIRECTORY_PATH "${BN_CORE_PLUGIN_DIR}")
```
bdash, line 32: This needs `plugin_rpath` / `bn_install_plugin` calls like other plug-ins have.

plugins/emulator/api/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+set_target_properties(emulatorapi PROPERTIES
+		CXX_STANDARD 20
+		CXX_VISIBILITY_PRESET hidden
+		CXX_STANDARD_REQUIRED ON
+		VISIBILITY_INLINES_HIDDEN ON
+		POSITION_INDEPENDENT_CODE ON
+		ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/out)
+
+# Python bindings retarget (load the emulatorcore plugin dylib) is pending; wire the
```
bdash, line 21: This comment and file existence check seem outdated.

docs/guide/emulator.md
emesare: Right? Is that not what the developer section is for?

plugins/emulator/core/ilemulator.cpp
```diff
@@ -0,0 +1,964 @@
...
+		if (m_memoryReadHook(this, addr, size, value))
+			return value;
+	}
+
+	intx::uint512 result = 0;
+	if (!m_memory.ReadValue(addr, size, GetEndianness(), result))
+	{
+		SetStopReason(ILEmulatorStopReason::Error,
+			fmt::format("failed to read {} bytes at 0x{:x}", size, addr));
+		return 0;
```
emesare, line 464: I would expect a failure to read memory to not pass back a constant value and instead signal some error state, im not certain an out of band "stop reason" suffices here, if that is fine, then the function should be documented that 0 is passed back on failure.

plugins/emulator/api/python/ilemulator.py
```diff
@@ -0,0 +1,637 @@
...
+        if entries:
+            for i in range(count.value):
+                result.append({
+                    'function_address': entries[i].functionAddress,
+                    'return_address': entries[i].returnAddress,
+                })
+            core.BNLLILEmulatorFreeCallStack(entries)
+        return result
+
+    def get_mapped_regions(self) -> List[Dict]:
```
emesare, line 167: Kind of suprised we are using dict here instead of a something typed

plugins/emulator/api/python/ilemulator.py
```diff
@@ -0,0 +1,637 @@
...
+        count = len(values)
+        arr = (ctypes.c_uint64 * count)(*values)
+        core.BNLLILEmulatorSetArguments(self.handle, arr, count)
+
+    # ── Memory ────────────────────────────────────────────────────────────
+
+    def read_memory(self, addr: int, length: int) -> bytes:
+        """Read *length* bytes from emulator memory at *addr*."""
+        buf = (ctypes.c_ubyte * length)()
+        n = core.BNILEmulatorReadMemory(self._get_base(), buf, addr, length)
```
emesare, line 226: See my concern in the body of the read memory function about error handling

xusheng6:
I am about 1/3 progress going through all latest review comments. I am pushing changes gradually, but only after I am done with all comments will I batch reply to all of them and mark as resolved

Issue #979, opened 2018-03-09
Support for micro-architectures / architecture variants


psifertex:
- [ ] Binary View needs to provide additional information to architectures via callback or other method
- [ ] Ability to create specific extended CPU selection (in UI/API)

PR #8316, opened 2026-07-14
Add support for Apple's vendor-specific aarch64 instructions

### LINKED ISSUE #5933: Add support for `GENTER` and `GEXIT` in AArch64 on Apple platforms
The [GXF](https://blog.svenpeter.dev/posts/m1_sprr_gxf/) feature on newer Apple SoCs uses two custom instructions, `GENTER` and `GEXIT`. These appear in the iOS kernelcache and SPTM on newer devices.

Their opcodes are:

- `20 14 20 00`: GENTER
- `00 14 20 00`: GEXIT

These instructions are not supported by Binary Ninja, and are disassembled as undefined instructions. This is problematic on its own, but also acts as a barrier to disassembling the rest of a function, making it's impact more annoying than just being an undefined instruction.

bdash:
The following instructions are disassembled and lifted to intrinsics:
* `genter #imm5`
* `gexit`
* `sdsb <domain>`
* `wkdmc <Xd>, <Xs>`
* `wkdmd <Xd>, <Xs>`
* `at_as1elx <Xt>`
* `mul53lo <Vd>.2d, <Vm>.2d`
* `mul53hi <Vd>.2d, <Vm>.2d`

Additionally, AMX instructions are disassembled but not yet lifted.

Fixes https://github.com/Vector35/binaryninja-api/issues/5933.

These instructions are in encoding space that is unallocated by ARM. As we do not currently have any mechanism for architecture variants (https://github.com/Vector35/binaryninja-api/issues/979) and these encodings are not currently used by any other vendor, I've opted to have `Arm64Architecture` decode these unconditionally.

In hopes of a better future, I've structured the code such that the vendor-specific instruction handling is entirely separate from the primary disassembler / lifting logic. When support for architecture variants is added in the future, this logic can all be moved into an Apple-specific variant.

### DIFF
--- a/arch/arm64/apple_vendor.cpp
+++ b/arch/arm64/apple_vendor.cpp
@@ -1,0 +1,405 @@
+// Copyright (c) 2026 Vector 35 Inc
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+
+#include "apple_vendor.h"
+
+#include <string.h>
+
+#include "disassembler/regs.h"
+
+using namespace BinaryNinja;
+using namespace std;
+
+namespace {
+
+enum AppleVendorOp
+{
+	AV_NONE,
+	AV_GENTER,
+	AV_GEXIT,
+	AV_SDSB,
+	AV_WKDMC,
+	AV_WKDMD,
+	AV_AT_AS1ELX,
+	AV_MUL53LO,
+	AV_MUL53HI,
+	AV_AMX,
+	AV_APPLE_UNKNOWN_SPTM,
+};
+
+struct AppleVendorInsn
+{
+	AppleVendorOp op = AV_NONE;
+	uint32_t imm = 0;  // genter selector, sdsb domain
+	uint32_t rd = 0;   // destination register field, or the single operand of at
+	uint32_t rs = 0;   // source register field
+};
+
+const char* const SDSB_DOMAIN[4] = {"osh", "nsh", "ish", "sy"};
+
+// AMX operation names indexed by the op field (bits [9:5])
+// Op 17 is the enable/disable pair (set/clr) and is rendered specially.
+// https://github.com/corsix/amx
+const char* const AMX_OPS[23] = {
+	"ldx", "ldy", "stx", "sty", "ldz", "stz", "ldzi", "stzi", "extrx", "extry", "fma64", "fms64",
+	"fma32", "fms32", "mac16", "fma16", "fms16", "set", "vecint", "vecfp", "matint", "matfp", "genlut",
+};
+
+bool Decode(uint32_t insn, AppleVendorInsn& out)
+{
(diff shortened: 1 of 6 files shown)


arch/arm64/apple_vendor.cpp
```diff
@@ -0,0 +1,385 @@
...
+}
+
+bool AppleVendorIsIntrinsic(uint32_t intrinsic)
+{
+	return intrinsic >= APPLE_INTRIN_GENTER && intrinsic < APPLE_INTRIN_END;
+}
+
+void AppleVendorGetAllIntrinsics(vector<uint32_t>& result)
+{
+	for (uint32_t id = APPLE_INTRIN_GENTER; id < APPLE_INTRIN_END; id++)
```
plafosse, line 328: Can we insert static vector instead here? Suggesting:

static constexpr array<uint32_t, 8> avIntrinsics = {
    APPLE_INTRIN_GENTER,
    APPLE_INTRIN_GEXIT,
    APPLE_INTRIN_SDSB,
    APPLE_INTRIN_WKDMC,
    APPLE_INTRIN_WKDMD,
    APPLE_INTRIN_AT_AS1ELX,
    APPLE_INTRIN_MUL53LO,
    APPLE_INTRIN_MUL53HI,
};

result.insert(result.end(), avIntrinsics.begin(), avIntrinsics.end());

arch/arm64/apple_vendor.cpp
```diff
@@ -0,0 +1,385 @@
...
+	case APPLE_INTRIN_MUL53HI:
+		return {NameAndType(Type::IntegerType(16, false)), NameAndType(Type::IntegerType(16, false))};
+	default:
+		return {};
+	}
+}
+
+vector<Confidence<Ref<Type>>> AppleVendorGetIntrinsicOutputs(uint32_t intrinsic)
+{
+	switch (intrinsic)
```
plafosse, line 393: non-gating comment. I've been meaning to do a pass over all these Instrinic return types converting them to static allocations as its one source of these APIs be a lot slower than they need to be.

arch/arm64/apple_vendor.cpp
bdash: This is called by `Arm64Architecture::GetAllIntrinsics` which itself returns a vector. The intrinsics need to make it into that vector. I'm not sure how creating a vector here would help since the elements would then need to be copied from one vector into another.

arch/arm64/apple_vendor.cpp
bdash: I'm not sure what you mean by "static allocations", but I doubt that using static vectors like some other functions in `Arm64Architecture` do would help much. The API still requires that a vector be returned by value, which means copying the static vector. I suspect the memory allocations are the expensive part of this, not whether the resulting memory is populated by a copy or by initialization.

I think we'd be better off caching information in core rather than calling out to the architecture repeatedly. That would also help with architectures implemented in languages like Python where it's going to be expensive due to FFI no matter how the architecture implements these methods.

arch/arm64/apple_vendor.cpp
```diff
@@ -0,0 +1,385 @@
...
+		il.AddInstruction(il.Intrinsic({}, APPLE_INTRIN_GEXIT, {}));
+		il.AddInstruction(il.Trap(0));
+		return false;
+	case AV_SDSB:
+		il.AddInstruction(il.Intrinsic({}, APPLE_INTRIN_SDSB, {il.Const(4, decoded.imm)}));
+		break;
+	case AV_WKDMC:
+	case AV_WKDMD:
+		// Both write the destination and a status back into the source register, and both read the source.
+		il.AddInstruction(il.Intrinsic(
```
plafosse: I think you need to implement GetIntrinsicClass and ensure this is lifted as a MemoryIntrinisc see INTRINSIC_XED_IFORM_REP_MOVSD for an example of how this is done.

plafosse: Are you sure the dest register is actually written or is just the data at the dest register written

arch/arm64/apple_vendor.cpp
plafosse: It is far faster to do reserve+memcpy (essentially what std::insert does) rather than push_back in a loop. But honestly it doesn't really matter. This code will not cause performance problems and is fine as is.

arch/arm64/apple_vendor.cpp
plafosse: Yeah thats a great idea. We already cache a bunch of other stuff but looking into it a bit we don't do it for Intrinsics and we absolutely should! Caching is done in `Architecture::Init()` FWIW.

plafosse (changes requested):
The only blocking thing here is addressing the AV_WKDMD/AV_WKDMC issue and adding a unit test containing these instructions in the core repo.

arch/arm64/apple_vendor.cpp
bdash: It's a little frustrating that you're editing comments that I've already responded to so that they retroactively say something different.

plafosse: sorry wasn't my intention. I think I had it up in two different windows or something.

arch/arm64/apple_vendor.cpp
bdash: I think you're right about MemoryIntrinsic here. I've updated this.

plafosse: Yeah that one is subtle and easy to miss.

plafosse (approved):
Looks good now.

PR #8325, opened 2026-07-15
Integrated template simplifier

plafosse:
## Summary

This PR integrates template simplification directly into the structured GNU3 and MSVC demanglers, introduces a unified configuration/result API, and migrates the C++, Python, and Rust bindings and consumers to that API.

The related [core PR #1575](https://github.com/Vector35/binaryninja/pull/1575) has already been reviewed and deemed ready. The remaining review focus is the API and consumer side represented by this PR.

## Review strategy

Although the aggregate diff is large, this PR has been reorganized specifically to make it easier to review **one commit at a time**.

Each commit is intended to be a coherent review unit, with later commits building on the interfaces introduced by earlier ones. Fixups and review changes have been folded back into the commit where they logically belong. The resulting tree is unchanged from the pre-reorganization snapshot; only the history and presentation were improved.

The recommended review order is:

1. **Commonize demangler infrastructure**

   Shares the reader, logging, structured type-node, traversal, mutation, and rendering infrastructure used by the GNU3 and MSVC demanglers.

2. **Integrate template simplification into structured demanglers**

   Performs template simplification on the structured demangler AST before final rendering. This contains the main simplification behavior, graph-safety handling, and correctness tests.

3. **Introduce config-based demangler APIs**

   Introduces `BNDemanglerConfig` and `BNDemanglerResult`, unified demangling entry points, C++ wrappers, ownership conversions, and the associated ABI change.

4. **Migrate Python demangler bindings**

   Adds the Python `DemanglerConfig` and `DemangleResult` APIs and migrates the built-in and custom demangler interfaces.

   Notable intentional API changes include:

   - Named demanglers return `Optional[DemangleResult]`.
   - `DemangleResult.name` is always a `QualifiedName`.
   - `demangle_any` accepts an optional configuration.
   - `demangle_generic` is replaced by `demangle_any`.
   - Named helpers accept `simplify: bool = True`.

5. **Migrate Rust demangler bindings and consumers**

   Adds the equivalent Rust configuration/result types, updates custom-demangler callbacks, migrates Rust consumers and examples, and removes the obsolete standalone simplifier wrapper.

6. **Migrate remaining demangler consumers**

   Updates the remaining views, cache workflows, RTTI users, UI declarations, and plugin wiring. Frequently used consumers retain a `DemanglerConfig` rather than reconstructing one in inner loops.

Reviewing the commits independently should be substantially easier than attempting to understand the complete GitHub diff at once.

## Why this approach

Previously, template simplification was handled separately from structured demangling. Performing simplification directly on the demangler’s structured representation:

- Avoids reparsing or inferring structure from rendered names.
- Keeps name and type information intact until final rendering.
- Gives the GNU3 and MSVC implementations a shared simplification path.
- Provides a consistent configuration and result model across C++, Python, and Rust.
- Improves performance; evaluation showed approximately a 200% improvement compared with the previous separation between core demangling and simplification.

## Scope

This PR includes:

- Shared structured-demangler infrastructure.
- Structured template simplification for GNU3 and MSVC.
- The new C and C++ demangler configuration/result APIs.
- Python and Rust binding migrations.
- Migration of existing in-tree consumers.
- Tests covering the new demangler API and simplification behavior.

This PR does not attempt to redesign unrelated demanglers or perform broader cleanup outside the affected consumers.

## Reviewer focus

The core implementation has already been reviewed in the companion core PR. For this PR, the most useful areas of review are:

- Whether the public C/C++ configuration and result contracts are clear and correctly owned.
- Whether the Python API presents a coherent transition from the previous helpers.
- Whether the Rust API follows the same semantics idiomatically.
- Whether migrated consumers construct and reuse configurations appropriately.
- Whether the six commits form understandable and independently reviewable stages.


demangler/msvc/demangle_msvc.cpp
```diff
@@ -2606,166 +2586,65 @@ Demangle::DemangleContext Demangle::DemangleSymbol(BackrefList& backrefList)
...
-	if (!view)
-		return false;
-	return DemangleMS(mangledName, outType, outVarName, Ref<BinaryView>(view));
-}
-
-
 class MSDemangler: public Demangler
 {
 public:
-	MSDemangler(): Demangler("MS")
```
CouleeApps, line 2748: Changing the name of the demangler here seems unnecessary and prone to breaking compatibility

demangler/gnu3/demangle_gnu3.cpp
```diff
@@ -28,341 +30,391 @@ using namespace BinaryNinja;
 using namespace std;
 #endif
 
+namespace
+{
+	void AppendStringView(string& out, std::string_view s)
```
CouleeApps: Surely this is just std::string::append, no?

demangler/gnu3/demangle_gnu3.cpp
```diff
@@ -28,341 +30,391 @@ using namespace BinaryNinja;
...
-		return to_string(u.f);
-	}
-	else if (byteCount == 8)
+		switch (hash(elm1, elm2))
+		{
+			case hash('d','c'): return "dynamic_cast";
+			case hash('s','c'): return "static_cast";
+			case hash('c','c'): return "const_cast";
+			case hash('r','c'): return "reinterpret_cast";
+			case hash('t','i'): // fall through
```
CouleeApps: Replacing these identical returns with `// fall through` seems unnecessary, but also why not just use the standard `[[fallthrough]]` (C++17) instead if you were going to annotate it anyway

demangler/gnu3/demangle_gnu3.cpp
```diff
@@ -3065,23 +3128,21 @@ DemangledTypeNode DemangleGNU3::DemangleSymbol(StringList& varName)
 	type.SetPointerSuffixBits(suffix);
 	type.SetConst(cnst);
 	type.SetVolatile(vltl);
-	if (rstrct)
```
CouleeApps, line 3068: Are we just removing support for demangling restrict pointers? It was entirely presentation-only already but I'm curious why this is here.

demangler/gnu3/demangle_gnu3.cpp
```diff
@@ -3201,108 +3268,77 @@ namespace
...
-// ===== Explicit template instantiation =====
-
-
-// ===== Demangler plugin registration =====
-
 class GNU3Demangler: public Demangler
 {
 public:
-	GNU3Demangler(): Demangler("GNU3")
+	GNU3Demangler(): Demangler("gnu3")
```
CouleeApps: Similar comment to the ms demangler, why the name change?

binaryninjaapi.h
```diff
@@ -2434,106 +2433,13 @@ namespace BinaryNinja {
...
-	    \param[in] arch Architecture for the symbol. Required for pointer and integer sizes.
-	    \param[in] mangledName a mangled GNU3 name
-	    \param[out] outType Reference to Type to output
-	    \param[out] outVarName QualifiedName reference to write the output name to.
-	    \param[in] view View to check the analysis.types.templateSimplifier for
-	    \return True if the name was demangled and written to the out* parameters
+	    \param[in] mangledName a potentially mangled name
 
 	    \ingroup demangle
 	*/
```
CouleeApps, line 2534: Since you're moving these defs, you should move the IsXXXMangledString ones too.

binaryninjaapi.h
```diff
@@ -4833,6 +4720,72 @@ namespace BinaryNinja {
 		static QualifiedName FromAPIObject(const BNQualifiedName* name);
 	};
 
+	struct DemanglerConfig
+	{
+		Ref<Platform> platform;
+		// Borrowed reference; the caller must keep the view alive while this config is in use.
+		BinaryView* view = nullptr;
```
CouleeApps: What is up with this? You have a private `Ref<BinaryView> m_viewOwner` later down in the structure, so what is the purpose of having an extra pointer to the view here? Also the weird lifetime condition this comment tries to impose is both wrong (because of the strong ref below) and confusing to users of this api...

binaryninjaapi.h
```diff
@@ -4833,6 +4720,72 @@ namespace BinaryNinja {
...
+		Ref<BinaryView> m_viewOwner;
+	};
+
+	struct DemanglerResult
+	{
+		QualifiedName name;
+		Ref<Type> type;
+
+		static DemanglerResult FromAPIObject(const BNDemanglerResult* result);
+		static DemanglerResult FromAPIObjectAndFree(BNDemanglerResult* result);
```
CouleeApps: Same comment about the ToAPIObject/FromAPIObject/FreeAPIObject as the core repo, these should be standardized

binaryninjaapi.h
```diff
@@ -4833,6 +4720,72 @@ namespace BinaryNinja {
...
+
+	/*! Demangles using LLVM's demangler.
+
+		\param[in] mangledName A mangled MSVC/GNU3/Rust/D name.
+		\param[in] simplify Whether to simplify demangled names.
+		\return Demangled type/name if successful.
+
+		\ingroup demangle
+	*/
+	std::optional<DemanglerResult> DemangleLLVM(
```
CouleeApps, line 4744: :+1: for using std::optional instead of the previous return-bool-and-out-param.

binaryninjaapi.h
```diff
@@ -22547,77 +22467,85 @@ namespace BinaryNinja {
...
 		static std::vector<Ref<Demangler>> GetList();
-		static Ref<Demangler> GetByName(const std::string& name);
+		static Ref<Demangler> GetByName(const std::string& demanglerName);
 
 		/*! Promote a demangler to the highest-priority position.
 
 			\param demangler Demangler to promote
+			\return True if promotion succeeded; false if the demangler was invalid or not registered.
+		 */
+		static bool Promote(const Ref<Demangler>& demangler);
```
CouleeApps, line 22548: Nit: should put this function down next to `Demangle`

binaryninjacore.h
CouleeApps, line 1: Don't forget to bump ABI version (it probably got eaten by a rebase)

binaryninjacore.h
```diff
@@ -3903,13 +3898,27 @@ extern "C"
...
+
+	typedef struct BNDemanglerResult
+	{
+		BNQualifiedName name;
+		BNType* type;
+	} BNDemanglerResult;
+
 	typedef struct BNDemanglerCallbacks
 	{
+		size_t size;
```
CouleeApps: This got missed in a revert of the core change that was trying to add this

binaryninjacore.h
```diff
@@ -8259,37 +8260,32 @@ extern "C"
...
-	    char*** outVarName, size_t* outVarNameElements, const BNBinaryView* const view);
-	BINARYNINJACOREAPI void BNFreeDemangledName(char*** name, size_t nameElements);
-
-	BINARYNINJACOREAPI bool BNDemangleLLVM(const char* mangledName,
-	    char*** outVarName, size_t* outVarNameElements, const bool simplify);
-	BINARYNINJACOREAPI bool BNDemangleLLVMWithOptions(const char* mangledName,
-	    char*** outVarName, size_t* outVarNameElements, const BNBinaryView* const view);
-
-	BINARYNINJACOREAPI BNDemangler* BNRegisterDemangler(const char* name, BNDemanglerCallbacks* callbacks);
+#define BN_DEMANGLER_MSVC "msvc"
```
CouleeApps: Why are these here? If it is really necessary to have macros for a couple of built-in plugin names, at least put them at the top of the file where the other macros go

type.cpp
```diff
@@ -349,21 +350,21 @@ QualifiedName QualifiedName::operator+(const QualifiedName& other) const
 BNQualifiedName QualifiedName::GetAPIObject() const
 {
 	BNQualifiedName result;
-	result.nameCount = m_name.size();
 	result.join = BNAllocString(m_join.c_str());
-	result.name = new char*[m_name.size()];
```
CouleeApps, line 354: Use `AllocApiStringList`? If you were going to change the allocation on this anyway, you might as well use the helper function.

demangle.cpp
```diff
@@ -1,145 +1,174 @@
...
-			return false;
-		outType = localType ? new Type(localType) : nullptr;
-		for (size_t i = 0; i < localSize; i++)
-		{
-			outVarName.push_back(localVarName[i]);
-		}
-		BNFreeDemangledName(&localVarName, localSize);
-		return true;
+		BNDemanglerConfig config = BNGetDefaultDemanglerConfig();
+		static auto cfg = FromAPIObject(&config);
```
CouleeApps: Static here is a good idea but the input for the constructor to the static is not itself static, so actually this is leaking memory every time it gets used

python/demangle.py
```diff
@@ -19,267 +19,279 @@
...
-	:Example:
+	type: Optional['types.Type']
+	name: 'types.QualifiedName'
 
-		>>> type, name = demangle_ms(Architecture["x86_64"], "?testf@Foobar@@SA?AW4foo@1@W421@@Z")
-		>>> get_qualified_name(name)
-		'Foobar::testf'
-		>>>
+
+def _demangle_result_from_core_and_free(
```
CouleeApps: Should be `DemangleResult._from_core_struct` to mirror the rest of the FFI bindings, and then handle the free-operation in the caller

python/demangle.py
```diff
@@ -19,267 +19,279 @@
...
+def demangle_any(
 		mangled_name: str,
-		view: Optional['binaryview.BinaryView'] = None,
-		simplify: bool = False
-) -> Optional[Tuple[Optional['types.Type'], List[str]]]:
+		config: Optional[DemanglerConfig] = None
+) -> Optional[DemangleResult]:
 	"""
-	``demangle_generic`` demangles a mangled symbol name to a Type object.
+	``demangle_any`` demangles a mangled symbol name using an optional prebuilt demangler config.
```
CouleeApps: Better doc for this would be "Attempt to demangle a mangled name, trying all relevant demanglers and using whichever one accepts it." which is what you used in the C++ docs

python/demangle.py
```diff
@@ -19,267 +19,279 @@
...
 	"""
-	result = None
-	if isinstance(input_name, str):
-		result = core.BNRustSimplifyStrToStr(input_name)
-	elif isinstance(input_name, types.QualifiedName):
-		result = core.BNRustSimplifyStrToStr(str(input_name))
-	else:
-		raise TypeError("Parameter must be of type `str` or `types.QualifiedName`")
-	return result
+	return _demangle_type_and_name(arch_or_platform, mangled_name, simplify, core.BNGetMSVCDemangler)
```
CouleeApps: Not sure why `demangle_ms` and `demangle_gnu3` get a special helper but `demangle_llvm` doesn't? If you're going to change these signatures, make all 3 consistent.

python/demangle.py
```diff
@@ -19,267 +19,279 @@
...
 
 
-def simplify_name_to_qualified_name(input_name: Union[str, types.QualifiedName], simplify: bool = True):
```
CouleeApps, line 253: Straight up removing public APIs like this is not in line with our deprecation policy... At least mark it as deprecated and provide a shim to the new implementation for a release.

python/demangle.py
```diff
@@ -345,34 +359,38 @@ def register(cls):
...
+
 		cls._registered_demanglers.append(demangler)
+		return True
 
 	@classmethod
 	def promote(cls, demangler):
 		"""
 		Promote a demangler to the highest-priority position.
 
-			>>> list(Demangler)
```
CouleeApps, line 367: Don't remove this documentation example

python/demangle.py
```diff
@@ -19,267 +19,279 @@
...
-		if not handle:
-			return (None, names)
-		return (types.Type.create(handle), names)
-	return (None, mangled_name)
+	return _demangle_result_from_core_and_free(result)
 
 
-def simplify_name_to_string(input_name: Union[str, types.QualifiedName]):
+def demangle_ms(
+		arch_or_platform: Union[Architecture, Platform],
```
CouleeApps: Would be nice if you could pass BinaryView here too and make use of the new DemangleConfig with BinaryView feature that you added.

python/demangle.py
```diff
@@ -441,15 +459,15 @@ def is_mangled_string(self, name: str) -> bool:
 
 	def demangle(
 			self,
-			arch: Architecture,
```
CouleeApps, line 444: Changing these method signatures is also a breaking change with no deprecation path given... Not sure how to do a transition for this without turning the API into spaghetti. It's probably not going to break too many users' code for this demangler instance method, but it's still against policy.

rust/src/rc.rs
```diff
@@ -318,14 +318,6 @@ where
 
 #[allow(private_bounds)]
 impl<P: CoreArrayProviderInner> ArrayGuard<P> {
-    pub(crate) unsafe fn new(raw: *mut P::Raw, count: usize, context: P::Context) -> Self {
```
CouleeApps, line 321: Why is this change here? This is entirely irrelevant to this PR

rust/src/demangle.rs
```diff
@@ -16,192 +16,254 @@
...
+            unsafe { BNGetDemanglerConfigForPlatform(platform.handle, simplify_templates) };
+        Self::from_api_object(&config)
+    }
+
+    pub fn for_binary_view(view: &BinaryView) -> Self {
+        let config = unsafe { BNGetDemanglerConfigForBinaryView(view.handle) };
+        Self::from_api_object(&config)
+    }
+
+    pub fn from_api_object(config: &BNDemanglerConfig) -> Self {
```
CouleeApps: Should be pub(crate) and not exposed

rust/src/demangle.rs
```diff
@@ -16,192 +16,254 @@
...
+        let platform = match config.platform.is_null() {
+            true => Self::default().platform,
```
CouleeApps: Also clippy has a point here that using `default()` inside `from_api_object()` is a risk of infinite recursion since `default()` calls `from_api_object()`... Not sure if that happens in practice but it looks bad on static analysis

rust/src/demangle.rs
```diff
@@ -16,192 +16,254 @@
...
+    }
+
+    pub fn get_platform(&self) -> Ref<Platform> {
+        self.platform
+            .clone()
+            .or_else(|| Self::default().platform)
+            .expect("default demangler config has no platform")
+    }
+
+    pub fn to_api_object(&self) -> BNDemanglerConfig {
```
CouleeApps: Also needs pub(crate)

rust/src/demangle.rs
```diff
@@ -16,192 +16,254 @@
...
+
+impl DemanglerResult {
+    pub fn new(name: impl Into<QualifiedName>, ty: Option<Ref<Type>>) -> Self {
+        Self {
+            name: name.into(),
+            ty,
+        }
+    }
+
+    pub fn from_api_object(result: &BNDemanglerResult) -> Self {
```
CouleeApps: Also needs pub(crate)

rust/src/demangle.rs
```diff
@@ -16,192 +16,254 @@
...
+            false => Some(unsafe { Type::ref_from_raw(BNNewTypeReference(result.type_)) }),
+        };
+        Self {
+            name: QualifiedName::from_raw(&result.name),
+            ty,
         }
-        false => None,
+    }
+
+    pub fn from_api_object_and_free(result: &mut BNDemanglerResult) -> Self {
```
CouleeApps: Also needs pub(crate)

rust/src/demangle.rs
```diff
@@ -16,192 +16,254 @@
...
+        let demangler_result = Self::from_api_object(result);
+        unsafe { BNFreeDemanglerResult(result) };
+        demangler_result
+    }
+
+    pub fn to_api_object(&self) -> BNDemanglerResult {
```
CouleeApps: Also needs pub(crate)

rust/src/demangle.rs
```diff
@@ -16,192 +16,254 @@
...
+        Self {
+            name: name.into(),
+            ty,
+        }
+    }
+
+    pub fn from_api_object(result: &BNDemanglerResult) -> Self {
+        let ty = match result.type_.is_null() {
+            true => None,
+            false => Some(unsafe { Type::ref_from_raw(BNNewTypeReference(result.type_)) }),
```
CouleeApps, line 131: Not sure of the semantics of adding a ref here versus not (as previous). I figure your test probably encountered a UAF from the previous behavior but just pointing this out as notable

rust/src/demangle.rs
```diff
@@ -16,192 +16,254 @@
...
-        BNDemangleGNU3(
-            arch.handle,
-            mangled_name.as_ptr(),
-            &mut out_type,
-            &mut out_name,
-            &mut out_size,
-            simplify,
-        )
-    };
+impl From<(QualifiedName, Option<Ref<Type>>)> for DemanglerResult {
```
CouleeApps, line 161: Nice

rust/src/demangle.rs
```diff
@@ -16,192 +16,254 @@
...
+pub fn demangle_any(mangled_name: &str, config: &DemanglerConfig) -> Option<DemanglerResult> {
+    let mangled_name = mangled_name.to_cstr();
+    let api_config = config.to_api_object();
+    let mut result = BNDemanglerResult::default();
+    let res = unsafe { BNDemangle(mangled_name.as_ptr(), &api_config, &mut result) };
+    collect_demangler_result(res, &mut result)
 }
 
-pub fn demangle_ms_with_view(
+pub fn demangle_generic(
```
CouleeApps, line 209: You removed `demangle_generic` from the Python and C++ APIs but kept it for Rust... These need to be consistent. I would say, add it back to Python and C++ but mark as deprecated (if removing it is your desire).

rust/src/demangle.rs
```diff
@@ -16,192 +16,254 @@
...
-            Some((names.into(), out_type))
-        }
-        false => None,
+pub fn demangle_llvm(mangled_name: &str, simplify: bool) -> Option<DemanglerResult> {
+    let mut config = DemanglerConfig::default();
+    config.simplify_templates = simplify;
+    demangle_with_demangler(llvm_demangler(), mangled_name, &config)
+}
+
+pub fn demangle_gnu3(
```
CouleeApps, line 241: Same comment here as in python about the inconsistency of these 3 sibling functions. If you're going to change them, make them consistent.

rust/src/demangle.rs
```diff
@@ -16,192 +16,254 @@
...
-    }
+pub fn demangle_any(mangled_name: &str, config: &DemanglerConfig) -> Option<DemanglerResult> {
+    let mangled_name = mangled_name.to_cstr();
+    let api_config = config.to_api_object();
+    let mut result = BNDemanglerResult::default();
+    let res = unsafe { BNDemangle(mangled_name.as_ptr(), &api_config, &mut result) };
+    collect_demangler_result(res, &mut result)
 }
 
-pub fn demangle_ms_with_view(
```
CouleeApps, line 168: You're removing the `_with_view` variants for the demangler-specific functions without a replacement. Either provide that or a `_with_config` variant maybe so that consumers can still get that behavior. The API break with no deprecation in the rust API is acceptable, but removing the functionality when your change set actually tries to expand upon it should be changed.

plugins/pdb-ng/src/symbol_parser.rs
```diff
@@ -1813,9 +1813,15 @@ impl<'a, S: Source<'a> + 'a> PDBParserInstance<'a, S> {
...
         rva: Rva,
     ) -> Result<(Option<Conf<Ref<Type>>>, Option<QualifiedName>)> {
-        let (mut t, mut name) = match demangle_ms_with_view(&self.arch, raw_name, Some(self.bv)) {
-            Some((name, Some(t))) => (Some(Conf::new(t, DEMANGLE_CONFIDENCE)), name),
-            Some((name, _)) => (None, name),
+        let simplify_templates = self.settings.get_bool_with_opts(
+            "analysis.types.templateSimplifier",
+            &mut self.settings_query_opts.clone(),
+        );
+        let (mut t, mut name) = match demangle_ms(&self.arch, raw_name, simplify_templates) {
```
CouleeApps: Losing the `_with_view` variant here? Not sure what impact that has, but seems like a downgrade

view/sharedcache/workflow/SharedCacheWorkflow.cpp
```diff
@@ -85,8 +96,9 @@ void IdentifyStub(BinaryView& view, const SharedCacheController& controller, uin
 	if (!symbol.has_value())
 		return;
 
-	// TODO: The demangled type here is almost always wrong so we omit it for now.
```
CouleeApps, line 88: I haven't confirmed locally but would like to know: are these TODOs actually fixed by this change? I can't imagine the type would go from "almost always wrong" to "correct enough" to remove this TODO and still throw away the type.

view/sharedcache/workflow/SharedCacheWorkflow.cpp
```diff
@@ -95,8 +107,6 @@ void IdentifyStub(BinaryView& view, const SharedCacheController& controller, uin
 	{
 		// NOTE: The type library name is expected to be the image name currently.
 		// Try and pull the type from the associated type library (if there is one)
-		// TODO: The demangled type here is missing a param
```
CouleeApps, line 98: This one too

CouleeApps (changes requested):
Overall, I like the general shape of the changes proposed here. Moving the template simplifier into the demangler where it can work with the name AST is a major improvement in functionality and likely in performance too. Providing a unified Demangler Config structure for operation is also really smart, and helps clean up one of our older, more tech debt-filled APIs.

That said, I have quite a few things I noticed in this PR. See the pile of line comments and also the following:
* Putting the template simplifier in the API like this excludes it from being used in AST-mode for Python and Rust plugins. I'm not sure of a good way to do that without moving the entire thing into a core API and doing FFI for everything, which it seems like you didn't want to do. I would still recommend it though.
* Rebasing this branch caused the core PR branch to desync commits. Be sure to squash when you rebase or otherwise update the submodule pointer in the intermediate commits so that it always points to a commit on the rebased branch.

binaryninjaapi.h
```diff
@@ -2543,25 +2449,7 @@ namespace BinaryNinja {
 	*/
 	bool IsGNU3MangledString(const std::string& mangledName);
 
-	/*!
-		\ingroup demangle
-	*/
```
CouleeApps, line 2548: Not sure of our policy with regards to deprecation and API-stability in C++ is, but these are being removed without deprecation notice either.

demangler/gnu3/demangle_gnu3.cpp
plafosse: TIL

demangle.cpp
plafosse: yeah good catch

demangler/gnu3/demangle_gnu3.cpp
plafosse: No restrict support was removed here. The old local `rstrct` was never populated; `type.GetPointerSuffixBits()` already carries `RestrictSuffix` into the rebuilt function type.

rust/src/rc.rs
plafosse: This constructor became unused when the demangle helpers stopped consuming raw string arrays; retaining it produces a `dead_code` warning.

plafosse:
> Putting the template simplifier in the API like this excludes it from being used in AST-mode for Python and Rust plugins. I'm not sure of a good way to do that without moving the entire thing into a core API and doing FFI for everything, which it seems like you didn't want to do. I would still recommend it though.

Yes this PR was big enough that I didn't want to add a whole new API thats really only helpful for C++ mangled name analysis. I do think it would eventually be a very nice API to add. This would also make it possible to make template simplifcations extensible which we'll save for the future.

rust/src/rc.rs
CouleeApps: Still strange that these changes were load bearing but makes sense. This feels like it should keep existing though, maybe a `#[allow(dead_code)]` would be better suited.

binaryninjaapi.h
plafosse: Yeah not sure what happened there. Removing it completely.

demangler/msvc/demangle_msvc.cpp
plafosse: yeah I was trying to standarize casing its currently "MS", "GNU3" and "llvm". So what are your thoughts leave the inconsistency or potentially break compatibility? My thought was that we won't have a better time to standarize than now.

demangler/gnu3/demangle_gnu3.cpp
plafosse: Same comment as above

python/demangle.py
plafosse: I'll add them back with compatibility shims and unit tests in the core.

python/demangle.py
plafosse: Ah yeah good catch. That documentation got removed when I had the removed the ability to promote after initial plugin load. Since that particular change was removed this documentation should be re-added.

python/demangle.py
plafosse: Yeah, I'm not 100% sure on this idea. In a previous iteration passing in a BinaryView as the 3rd parameter meant that it would query the "simplifyTemplates" setting on the BinaryView. This was a HORRIBLE design decision and lead to 10x slow down on demangling. The goal here was to clean up these APIs. So this is why its done this way. I'm open for alternate suggestions but I really don't like the idea of adding back in the BinaryView. 

My view of the current change is that demangle_ms/gnu3/llvm are legacy apis that I've taken away a foot gun from.  Going forward the recommendation is to use `demangle_any` for pretty much everything. It takes an optional DemangleConfig so its easier to do use for quick console-based checks AND it is performant and uses the specified config when one is passed in. 

I am even considering adding in _prefered demangler_ and _fallback behavior_  into the Config.

demangler/msvc/demangle_msvc.cpp
CouleeApps: The LLVM was capitalized before? `LLVMDemangler(): Demangler("LLVM")`... Not sure why we would change all 3 of them given they were already consistent

python/demangle.py
CouleeApps: You have a good point about performance. Maybe you could allow passing the DemangleConfig instead of the view, and then users of this could get the improved config options for these. Also of note, this _is_ the Python API, where the performance of an extra setting lookup in the view is likely irrelevant compared to the overhead of the python interpreter.

demangler/msvc/demangle_msvc.cpp
plafosse: ok yeah sorry I'm just confused

view/sharedcache/workflow/SharedCacheWorkflow.cpp
plafosse: Fixed in 36a54eb9

view/sharedcache/workflow/SharedCacheWorkflow.cpp
plafosse: Fixed in 36a54eb9

plugins/pdb-ng/src/symbol_parser.rs
plafosse: Fixed in 603fe857

rust/src/demangle.rs
plafosse: Fixed in 603fe857

demangler/gnu3/demangle_gnu3.cpp
plafosse: They're all using the defined macros now and the name change is reverted in fd1edbc8 and 83151e4f

demangler/gnu3/demangle_gnu3.cpp
plafosse: Fixed in 01f57c65

demangler/gnu3/demangle_gnu3.cpp
plafosse: Fixed in 01f57c65

demangler/msvc/demangle_msvc.cpp
plafosse: Fixed in 01f57c65

binaryninjaapi.h
plafosse: Fixed in 83151e4f

binaryninjaapi.h
plafosse: fixed in 83151e4f

binaryninjaapi.h
plafosse: Fixed in 83151e4f

binaryninjaapi.h
plafosse: Fixed in 83151e4f

binaryninjacore.h
plafosse: Fixed in 83151e4f

binaryninjacore.h
plafosse: Fixed in 83151e4f

demangle.cpp
plafosse: Fixed in 83151e4f

python/demangle.py
plafosse: Fixed in 2883dd9d

python/demangle.py
plafosse: Fixed in 2883dd9d

python/demangle.py
plafosse: The purpose was because demangle_llvm is incapable of returning a type so t doesn't need that helper. The newly re-worked version completely changes how this is done as we're accept the old parameters to maintain backward compatibility but also allow the new DemangleConfig/DemangleResult to be used. This if I had to guess is the most contentious change. My other option was to just create a new API but I'm not sure that would be better. Fixed in 2883dd9d

python/demangle.py
plafosse: Fixed in 2883dd9d

python/demangle.py
plafosse: Fixed in 2883dd9d

python/demangle.py
plafosse: See above comment about the backward compat stuff I did here. I also added a big warning in the documentation to hopefully reduce api misuse. Fixed in 2883dd9d

rust/src/demangle.rs
plafosse: Fixed in 603fe857

rust/src/demangle.rs
plafosse: Fixed in 603fe857

rust/src/demangle.rs
plafosse: fixed in 603fe857

rust/src/demangle.rs
plafosse: Fixed in 603fe857

plafosse:
@CouleeApps Thanks for the extensive review! I _believe_ I have addressed all outstanding commentary. The modifications are all in fixup commits so they can be viewed independently. These I obviously plan to squash into the commit they reference before the PR is merged. Let me know what you think!

CouleeApps (approved):
Great changes, thanks for addressing all of the things I mentioned. Looking forward to seeing the improved results and perf from this branch.

PR #8336, opened 2026-07-20
updates to docs

ashvika-creator:


docs/dev/plugins.md
```diff
@@ -14,6 +14,23 @@ To start, we suggest you download the [sample plugin](https://github.com/Vector3
...
+
+### Images in README Files
+
+When you include images for preview in the Extension Manager, be aware that animated gifs are not fully supported. They will be rendered as static images of the first frame.
+
+Additionally, all images should be included using FULL github URLs as they will not render in the Extension Manager without that.
+
+For example:
+
+![](https://raw.githubusercontent.com/Vector35/6502/refs/heads/master/media/nes.png)
```
psifertex, line 35: we should show this not as an actual image but inside triple backticks so it's escaped.

docs/dev/plugins.md
```diff
@@ -25,24 +25,24 @@ Binary Ninja displays plugin descriptions differently depending on the contents
...
 
 ![](https://raw.githubusercontent.com/Vector35/6502/refs/heads/master/media/nes.png)
   
-### Submitting to the Plugin Manager
+### Submitting to the Extension Manager
 
-If your plugin was created as described above, there's only two steps to get it submitted to the plugin manager!
+If your plugin was created as described above, there's only two steps to get it submitted to the Extension Manager!
 
 1. First, create a release either [manually](https://binary.ninja/2019/07/04/plugin-manager-2.0.html#5-create-a-release) or using our [release helper](https://github.com/Vector35/release_helper).
```
psifertex: I don't think we should recommend this anymore. Instead, we can point to: https://github.com/vector35/sample_plugin#cutting-a-release

which automates the whole process! We should adjust the steps accordingly probably.

psifertex (changes requested):

PR #8338, opened 2026-07-20
refactor: migrate rust crates to edition 2024

seuros:
Run cargo fix --edition on binaryninja and binaryninjacore-sys, bump both manifests to edition 2024

Most of the change were automated by the fix.  There is more cosmetic changes possible (like order of the imports), but i reverted those.


emesare:
Make an issue please with specific instances of UB, or why you need 2024, otherwise we will bump to 2024 ourselves when we get the chance. Thank you.

PR #8355, opened 2026-07-22
[Python] Fix exceptions from __del__ during interpreter shutdown

bdash:
`StructureType`, `EnumerationType`, and `NamedTypeReferenceType` called `super(Class, self)` in `__del__`. The interpreter nulls that module-global class name during teardown, so finalizers running at shutdown raised a `TypeError`. Switch to zero-arg `super()` here and across the rest of the Python API.


plafosse (approved):
The conversions are mechanical and semantically equivalent for these methods and fix the interpreter-shutdown dependency on module-global class names. LGTM

PR #8356, opened 2026-07-22
Extend rbran's parser robustness fixes with file-backed bounds and configurable limits (ELF)

Weitao-Sun:
## Summary

Split out from #8303 per review feedback, containing the ELF portion only. (Mach-O is being merged in #8303; PE is a separate PR.)

Builds on rbran's parser robustness fixes with the following improvements:

- `ReadStringTable`: rejects sections whose declared size exceeds the actual file length, and fixes an off-by-one in the offset check, preventing reads past the end of the buffer.
- `Init`: fixes an integer underflow when `localMipsSyms` is zero.
- `ApplyTypesToParentStringTable`: bounds `section.size` to a configurable setting (`files.elf.maxStringTableTypeSizeMB`, default 4 MB) rather than trusting the declared section size directly.
- Adds `IsRangeBackedByFile`, which validates every byte in a range is file-backed (rather than only the first/last byte), and uses it to guard the MIPS GOT entry count against a `PT_LOAD` segment's declared file extent exceeding the actual file length (with overflow checking).
- Changes the MIPS GOT budget setting from an entry-count cap to `files.elf.maxMipsGotMB` (default 4 MB, max 64 MB), mirroring `fs/binfmt_elf.c`'s byte-size-based approach so the same limit applies equally regardless of 32/64-bit entry size.


view/elf/elfview.cpp
```diff
@@ -2753,7 +2804,27 @@ string ElfView::ReadStringTable(BinaryReader& reader, const Elf64SectionHeader&
...
+
+
+// Returns true only if the entire range [start, start+size) is backed by file data.
+bool ElfView::IsRangeBackedByFile(uint64_t start, uint64_t size) const
+{
+	if (size == 0)
+		return false;
+	if (size > UINT64_MAX - start)
+		return false;
+	for (uint64_t offset = start; offset < start + size; offset++)
```
emesare: This seems quite expensive, would it not be okay to just check the start and end? I see you mention it in the PR description but what is this guarding against?

emesare (changes requested):

view/elf/elfview.cpp
Weitao-Sun: Thanks for the feedback — I've adopted your suggested fix, pushed in d186a7569.

PR #8362, opened 2026-07-25
Fix setter decorator for project property

owah:
`PluginCommandContext.project`'s setter is decorated with `@function.setter` instead of `@project.setter`. Because `@x.setter` returns a new property built from `x`, this attaches the new setter to the *function* property while keeping its original getter, and binds the whole thing to the name `project`. The result: `context.project` reads back `self._function` instead of `self._project`.


plafosse:
Thanks for the fix!

PR #8382, opened 2026-08-03
Add Binary Similarity Documentation

emesare:

This change introduces another page in the user guide aimed at explaining and going over the features of the binary similarity system. Might still need to change / update the photos depending on how the rest goes.


docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+
+Binary Ninja includes two similarity providers:
+
+- **Google BinDiff** finds structurally similar functions, including functions that are not byte-for-byte identical.
+- **WARP** finds exact function matches using [WARP][WARP] signatures.
+
+Additional providers may be supplied by plugins. You can use more than one provider in the same session and compare
+their results together.
+
+![Similarity Sidebar](../img/similarity/sidebar.png "Similarity Sidebar")
```
psifertex, line 19: I wonder if just showing the sidebar itself and not the linear view above would be better, I'm not sure there's context above that matters? Easy enough to crop without re-taking the screenshot.

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+- **Start Session Processing** starts the comparison (only shown if stopped).
+- **Stop Session Processing** requests that the active run stop (only shown if running).
+
+### Quick Setup
+
+For a two-binary comparison, select **Quick Setup** in an empty session. The wizard has two pages:
+
+1. **Choose Binaries** selects a reference binary and a target binary. Each can be an open Binary Ninja view or a file
+   selected with **Browse files...**. Quick Setup creates the edge `reference -> target`.
+2. **Select Matching Components** lists providers and optional resolvers in a table. Select at least one provider.
```
psifertex, line 77: Could use a screenshot here too. Doesn't have to hold up acceptance though.

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+    For a simple before-and-after comparison, use the [Quick Setup](#quick-setup) action when first opening the sidebar.
+
+### Providers and Resolvers
+
+A provider reports possible function matches along with two measurements:
+
+- **Similarity** describes how closely the functions match.
+- **Confidence** describes how strongly the provider supports that match.
+
+The **Metric Similarity Resolver** combines available provider results, ignores results below its configured thresholds,
```
psifertex, line 46: What is the metric similarity resolver? A menu? A button? A tab? Like I get that it's. "resolver" as mentioned above, but I feel like some context is missing here as to where I find it / use it.

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+
+By default, the functions discovered in a binary are scheduled for processing. Right-click a node and select
+**Node Configuration...** to customize it:
+
+- Use the **Entities** tab to include or exclude functions from the next run.
+- Use the **Load Options** tab to control how Binary Ninja opens and analyzes that file. Useful for large binaries or raw firmware.
+
+Restricting a session to the functions you care about can substantially reduce processing time for large binaries.
+
+![Node Configuration 1](../img/similarity/nodeconfig1.png "Node Configuration 1")
```
psifertex, line 116: consider using tabs:

https://dev-docs.binary.ninja/guide/sharedcache.html#interacting-with-a-shared-cache
(aka https://github.com/Vector35/binaryninja-api/blame/dev/docs/guide/sharedcache.md#L100-L114)

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+
+![Node Configuration 2](../img/similarity/nodeconfig2.png "Node Configuration 2")
+
+### Configuring Matching
+
+Select **Session Configuration** on the session rail to open the configuration dialog. Providers and resolvers are
```
psifertex, line 122: "session rail"? This definitely could use a screenshot to clarify (or, if following the suggestion above to number the visual elements, just refer back to the top image with a number)

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+- Sort by similarity or confidence.
+- Color rows by similarity.
+- Show only resolved or unresolved functions.
+
+The check mark identifies the result selected by a resolver. Clicking a source or target address navigates to that
+function when its binary is open.
+
+If you wish to navigate with the main binary view, toggle the sidebar option **Sync with Navigation**.
+
+![Results](../img/similarity/results.png "Results")
```
psifertex, line 165: also could use a crop unless the content above is important? Or, if the main view is important too, shrinking the height of the window would make the important stuff easier to pick out.

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+
+### Applying a Result
+
+Right-click one or more result rows and select **Apply** to transfer the selected matches' available analysis
+information. The exact information depends on the provider and the compatibility of the functions.
+
+Applying requires active views for the affected binaries. If a required view is not available, Binary Ninja lists the
+binaries that are not open and asks whether to load and open them.
+
+After applying information make sure to save the database like you normally would, as the information will not be persisted otherwise.
```
psifertex, line 175: s/be persisted/persist/

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+
+### Rendering a Comparison
+
+Rendering is done via clicking on a result (or via the right click menu), the views will be shown with whatever information
+the provider deems relevant to show the user, typically this is done by highlighting changes.
```
psifertex, line 180: remove "to show the user"

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+- **Graph** and **Linear** tabs when the provider supplies those representations.
+- A view selector for common view levels (disassembly and ILs), this is a "preference" and the provider may ignore it.
+- **Synchronize Similarity Views**, synchronizes scrolling and zooming between comparison panes. It is disabled by default.
+- **Open Result in Main Tab**, which opens a copy of the rendered comparison in the main tab area.
+
+![Rendering Result](../img/similarity/render.png "Rendering Result")
+
+## Multiple Sessions
+
+Use the **New Similarity Session** command to create another independent session. When more than one session exists, a
```
psifertex, line 193: command-palette command? Is there no other UI or menu for it? Stylistically I believe we escape commands `as code` instead of **with emphasis**.

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+`session.run()` starts in the background, so headless scripts should wait for `completion.is_finished` and may call
+`completion.request_stop()` to cancel.
+
+When the [sidebar](#using-the-binary-similarity-sidebar) is open, the python console also exposes the active session as
+`current_similarity_session`. This is useful for inspecting or extending the graph programmatically without rebuilding the
+session in a separate script.
+
+## Troubleshooting
+
+Common troubleshooting scenarios, if you run into something you think we should add here let us know on our [Slack][Slack].
```
psifertex, line 248: easier to link to https://dev-docs.binary.ninja/guide/index.html?h=contact#support even though it's one layer of indirection so we have less things to update if we want to change things around (plus it points out the other support mechanisms)

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+the required binaries. Address navigation does not open a binary automatically; open it in Binary Ninja and the view will
+be attached to the sessions node so that navigation may occur.
+
+### Choosing Between WARP and BinDiff
+
+Use WARP when you expect functions to match exactly aside from relocatable or other variant instructions. Use BinDiff
+when you want to find functions that remain structurally related after recompilation or source changes. Using both
+providers can give a resolver exact evidence where available and broader candidates everywhere else.
+
+## Glossary
```
psifertex, line 276: I'm not sure about the glossary. If it's just to keyword stuff to make it easier to find the docs that might make sense, but I don't know if the definitions below are as useful as just using the phrases in the appropriate context where they are described above. Related: I was thinking about removing them from the other places we use them. Typically, a global glossary that links back to individual pages might make sense but I'm not sure about separate ones spread throughout.

No action item here, I'm just brainstorming out loud, we can discuss it more broadly later.

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
...
+providers when choosing between possible matches.
+
+### Resolved Result
+
+A **Resolved Result** is the result selected for an entity by a resolver. It is shown with a check mark in the results
+view.
+
+[WARP]: warp.md
+[Python API]: https://api.binary.ninja/binaryninja.similarity-module.html
+[Binary Ninja API repository]: https://github.com/Vector35/binaryninja-api/blob/dev/python/examples/bindiff.py
```
psifertex, line 341: is this from something that used to be linked and isn't anymore? I don't see a reference above currently. Also, which commit is the file itself in?

docs/guide/similarity.md
```diff
@@ -0,0 +1,342 @@
+# Binary Similarity
```
psifertex: General feedback on the docs:
 - should have a view that explains the functions of the sidebar buttons (something like https://dev-docs.binary.ninja/guide/index.html#edit-function-properties-dialog that is, the image itself)
 - are the images compressed with ect? some good (big) screenshots so if not probably some file savings

psifertex (changes requested):
Most of these should be quick changes. Only bigger issue is putting together an image with numbered overlays like other UI elements so that it's more clear which things you're talking about in which sections, it's easy to get lost in the text descriptions. @ashvika-creator can probably help with that if you'd like, she's got the screenshot setup at hand if you want to let her know the numbers/orders.

psifertex:
Weird -- my first review comment got eaten somehow? Specifically suggested a better overview image with numbered elements (alluded to in the first comment above). It also mentioned using [ect](https://github.com/Inc44/png_ect) on all the PNGs as from the size I think they might not have been compressed yet?

fuzyll (commented):
At least half of these images will need replaced once you tell me what icons you need and we get them properly placed into the product.

This doesn't necessarily need to block having _any_ documentation in the product, but we also shouldn't lose this as a release task.

emesare:
Added with `6.0.10489`, feedback was addressed in `bf4fc51a1315803f53e43859d369175225987b96`.

PR #8384, opened 2026-08-03
Add Binary Similarity Rust Bindings

emesare:

Adds the Rust bindings we depend on for WARP integration.

Depends on:
 - https://github.com/Vector35/binaryninja-api/pull/8386


emesare:
Added with `6.0.10489`

PR #8385, opened 2026-08-03
Add Binary Similarity Python Bindings

emesare:

Adds the python bindings we depend on for user examples and python core unit tests

Depends on:
 - https://github.com/Vector35/binaryninja-api/pull/8386

### DIFF
--- a/python/similarity.py
+++ b/python/similarity.py
@@ -1,0 +1,1765 @@
+# Copyright (c) 2015-2026 Vector 35 Inc
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to
+# deal in the Software without restriction, including without limitation the
+# rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+# sell copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+
+"""Binary similarity providers, session graphs, result resolution, and rendering."""
+
+import ctypes
+from dataclasses import dataclass, replace
+from datetime import timedelta
+from typing import Any, List, Optional
+
+import binaryninja
+
+from . import _binaryninjacore as core
+from . import binaryview, enums, filemetadata, flowgraph, function, lineardisassembly, settings
+from .log import log_error_for_exception
+
+SimilarityEntityId = int
+#: Identifies a provider instance within a session.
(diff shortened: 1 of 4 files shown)


python/similarity.py
```diff
@@ -0,0 +1,1765 @@
...
+        if self.provider_id is not None and self.resolver_id is not None:
+            raise ValueError("provider_id and resolver_id are mutually exclusive")
+
+    @classmethod
+    def for_session(cls) -> 'SimilaritySessionCompletionQuery':
+        """Select the whole session."""
+        return cls()
+
+    @classmethod
+    def for_node(cls, node_id: int) -> 'SimilaritySessionCompletionQuery':
```
plafosse, line 66: Type hint should be a defined type rather than an `int`

python/similarity.py
```diff
@@ -0,0 +1,1765 @@
...
+
+    def __repr__(self):
+        return f"<SimilarityEntityInfo type={self.type.name}, address={self.address:#x}, name={self.name!r}>"
+
+
+@dataclass(frozen=True)
+class SimilarityEntityRef:
+    """Identifies an entity within a session node."""
+
+    node_id: int
```
plafosse, line 115: use a defined type instead of an `int`

python/similarity.py
```diff
@@ -0,0 +1,1765 @@
...
+
+    def __repr__(self):
+        return f"<SimilarityEntityRef node_id={self.node_id}, entity_id={self.entity_id}>"
+
+
+@dataclass(frozen=True)
+class SimilarityView:
+    """A graph or linear view produced for a similarity result."""
+
+    group: str
```
plafosse, line 126: Use a defined type instead of `str`

python/similarity.py
```diff
@@ -0,0 +1,1765 @@
...
+
+    start: int
+    end: int
+    type: SimilarityAnnotationType
+
+
+class SimilarityRenderContext:
+    """Holds views used to display a similarity result."""
+
+    def __init__(self, handle=None):
```
plafosse, line 146: I know were inconsistent here but define a type hint for the handle. Its super helpful and can help avoid crashes

python/similarity.py
```diff
@@ -0,0 +1,1765 @@
...
+            core.BNFreeSimilaritySessionResolverTypeList(types)
+
+    def __getitem__(cls, value):
+        binaryninja._init_plugins()
+        resolver_type = core.BNGetSimilaritySessionResolverTypeByName(str(value))
+        if resolver_type is None:
+            raise KeyError(f"'{value}' is not a valid similarity session resolver type")
+        return CoreSimilaritySessionResolverType(handle=resolver_type)
+
+    def __contains__(cls: '_SimilaritySessionResolverTypeMetaClass', name: object) -> bool:
```
plafosse, line 830: shouldn't this be `self`

python/similarity.py
```diff
@@ -0,0 +1,1765 @@
...
+        if not isinstance(name, str):
+            return False
+        try:
+            cls[name]
+            return True
+        except KeyError:
+            return False
+
+    def get(
```
plafosse, line 839: should this first param be `self` as its not a classmethod

python/similarity.py
```diff
@@ -0,0 +1,1765 @@
...
+                )
+        finally:
+            core.BNFreeSimilarityViewList(views, count.value)
+        return result
+
+
+class DiffRenderer:
+    """Renders functions with similarity range annotations."""
+
+    def __init__(self, handle=None):
```
plafosse, line 234: handle type hint

plafosse:
Did a quick review, nothing stood out to me as wrong. All of my comments are non-blocking nits:

* Prefer named types rather than int/str
* Ensure your handles have type hints. 
* You incremented the BN_MINIMUM_CORE_ABI_VERSION instead of the BN_CURRENT_CORE_ABI_VERSION

I didn't check the core PR and you may have already done this but you should add these objects to the ActiveObjectCounter if you haven't already and add a test which validate that we're not leaking these objects. 

Actually one that confuses me and I don't know if its blocking or not

```python
SimilarityProviderType.create returns Optional['SimilarityProvider']:
```

but `SimilarityProviderType._create` raises `ValueError` if it returns None that seems odd?

plafosse (approved):

emesare:
Added with `6.0.10489` feedback was addressed in `eaacd6d26ee5ec55aa16080b6afe7947c2629e9c`.

PR #8386, opened 2026-08-04
[API] Add FFI wrapper modifier to strong typedefs

emesare:

This modifier is used for wrapping and unwrapping FFI structures representing the raw value type.

In particular we are using it in the binary similarity like:

```cpp
namespace BinaryNinjaCore {
	namespace st = ::bn::base::strong_typedef;
	using SimilarityEntityId = ::bn::base::StrongTypedef<uint32_t, struct SimilarityEntityIdTag,
		st::FfiWrapper<BNSimilarityEntityId, &BNSimilarityEntityId::value>, st::Ordered, st::Hashable,
		st::Incrementable>;
}  // namespace BinaryNinjaCore
```

Where we want to use the strong_typedef modifiers to derive functionality for use in containers (hashable, ordered) and still preserve the relationship between the C++ and C (via FfiWrapper).

Also see https://github.com/Vector35/binaryninja/pull/1737 for unit tests.

### DIFF
--- a/base/strong_typedef.h
+++ b/base/strong_typedef.h
@@ -22,6 +22,7 @@
 
 #include <compare>  // IWYU pragma: keep
 #include <stddef.h>
+#include <concepts>
 #include <functional>
 #include <type_traits>
 #include <utility>
@@ -89,6 +90,8 @@
 //     Hashable        usable as a key in std and absl hash containers
 //     Formattable     formattable with fmt, forwarding format specs such as {:#x} to the
 //                     underlying type's formatter
+//     FfiWrapper      implicit conversion to and from a C-style wrapper structure through
+//                     the specified member, without making the underlying type implicit
 //     NonExtractable  removes the explicit operator T() and the Value() accessor, so the
 //                     underlying value can be constructed but never read back out. Every
 //                     other modifier continues to function as normal.
@@ -114,6 +117,25 @@ namespace detail {
 // True when Modifier appears in the Mods pack.
 template <typename Modifier, typename... Mods>
 inline constexpr bool HasModifier = (std::is_same_v<Modifier, Mods> || ...);
+
+template <typename Mod, typename Ffi, typename T>
+concept FfiWrapperFor = requires(const Ffi& value) {
+	typename Mod::FfiType;
+	requires std::same_as<typename Mod::FfiType, Ffi>;
+	{ Mod::template FromFfi<T>(value) } -> std::same_as<T>;
+};
+
+template <typename Ffi, typename T, typename... Mods>
+inline constexpr bool HasFfiWrapper = (FfiWrapperFor<Mods, Ffi, T> || ...);
+
+template <typename T, typename Ffi, typename First, typename... Rest>
+constexpr T FromFfi(const Ffi& value)
+{
+	if constexpr (FfiWrapperFor<First, Ffi, T>)
+		return First::template FromFfi<T>(value);
+	else
+		return FromFfi<T, Ffi, Rest...>(value);
+}
 
 // Internal access to a StrongTypedef's underlying value so that modifiers have
 // access to it even when NonExtractable is in use.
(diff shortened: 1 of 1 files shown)


base/strong_typedef.h
```diff
@@ -89,6 +90,8 @@
 //     Hashable        usable as a key in std and absl hash containers
 //     Formattable     formattable with fmt, forwarding format specs such as {:#x} to the
 //                     underlying type's formatter
+//     FfiWrapper      implicit conversion to and from a C-style wrapper structure through
```
bdash: AFAIK we don't use "FFI" in the naming of types or methods related to the C++ <-> C API conversions. The methods in ffi.h are all framed in terms of "API objects".

A name like `APIStruct` could be a better fit.

base/strong_typedef.h
```diff
@@ -115,6 +118,25 @@ namespace detail {
 template <typename Modifier, typename... Mods>
 inline constexpr bool HasModifier = (std::is_same_v<Modifier, Mods> || ...);
 
+template <typename Mod, typename Ffi, typename T>
```
bdash: I'd suggest moving all of this new stuff to the end of the detail namespace, after `AffineOps`, so it's easier to find the operations like `Access` that are more widely used.

base/strong_typedef.h
```diff
@@ -409,6 +431,33 @@ struct Formattable
 	};
 };
 
+// Enables implicit conversion to and from a C-style wrapper structure whose selected
+// member stores the StrongTypedef's underlying value. Conversion to the underlying type
+// itself remains explicit.
+template <typename Ffi, auto ValueMember>
+struct FfiWrapper
```
bdash: I'd suggest aligning the members that this modifier provides to match the `APIAble` "concept" core has. That should allow helpers like `AllocAPIObjectList` / `ParseAPIObjectList` / `FreeAPIObjectList` to be used with these objects the core side.

Something like:
```cpp
template <auto ValueMember>
struct APIStruct
{
	static_assert(std::is_member_object_pointer_v<decltype(ValueMember)>);

	using TAPI = decltype(detail::MemberPointerClass(ValueMember));

	template <typename Self, typename T>
		requires requires(TAPI object, const T& value) { object.*ValueMember = value; }
	struct Apply
	{
			[[nodiscard]] constexpr TAPI GetAPIObject() const
			{
					TAPI result{};
					result.*ValueMember = detail::Access::Get(static_cast<const Self&>(*this));
					return result;
			}
			static constexpr Self FromAPIObject(const TAPI* object) { return Self(object->*ValueMember); }
			static constexpr void FreeAPIObject(TAPI*) {}
			constexpr operator TAPI() const { return GetAPIObject(); }
	};
};
```

`detail::HasAPIObject` would be your `HasFfiWrapper` without the `FromFfi` requirement (and so without the `T` template parameter).

This also shows how to have the member pointer as the only template parameter, avoiding restating the type. It relies on sneaky `detail::MemberPointerClass` that would look like this:
```cpp
// Recovers the class type from a pointer-to-member. Declared and never defined:
// it is only ever named in an unevaluated context.
template <typename Class, typename Value> Class MemberPointerClass(Value Class::*);
```

Additionally, since the conversion _from_ the API struct now lives within `Apply`, the constructor implementation can call it directly, and the helper function that walks the modifier pack can be removed:

```cpp
template <typename API>
	requires detail::HasAPIObject<std::remove_cvref_t<API>, Mods...>
constexpr StrongTypedef(API&& object) : StrongTypedef(StrongTypedef::FromAPIObject(&object))
{
}
```

bdash (changes requested):
I really like the idea, but there are a couple of changes that would help this fit better with our existing C++ <-> C API conversions.

base/strong_typedef.h
bdash: A slight improvement that just occurred to me: we can take the API struct type as the only template parameter.

We can do `return TAPI{detail::Access::Get(...)};` to construct the API type without knowing the field name. We can use structured binding to extract the only member of a struct via `const auto& [value] = *object;` without knowing its name.

This simplifies the use site, rejects wrapper structs that have more than one member, rejects narrowing conversions (if the struct member has the wrong type), and eliminates the need for the `detail::MemberPointerClass` helper.

emesare:
Ah yes, the initializer change really does improve the usability, made the requested changes, thanks!

bdash (approved):

PR #8387, opened 2026-08-04
Add BinDiff Similarity Provider (API)

emesare:

Adds a first party integration with https://github.com/google/bindiff meant for simple structural (cfg) diffing and matching, supersedes the binexport plugin.

Screenshot 2026-08-02 at 10 05 46 PM

An example of the rendering performed giving the bindiff annotations.

The sister PR is found here: https://github.com/Vector35/binaryninja/pull/1743

Depends on:
 - https://github.com/Vector35/binaryninja/pull/1743
 - https://github.com/Vector35/binaryninja-api/pull/8383


plugins/bindiff_similarity/diffview.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include <third_party/zynamics/binexport/virtual_memory.h>
+#include <functional>
+#include <string>
+
+class BinDiffView
+{
+public:
+	security::bindiff::CallGraph m_callGraph;
+	security::bindiff::FlowGraphs m_flowGraphs;
```
plafosse, line 17: I believe you're leaking memory here. BinDiffView needs a destructor.

```c++
~BinDiffView()
{
    security::bindiff::DeleteFlowGraphs(&m_flowGraphs);
}
```

FlowGraphs is defined as

```c++
using FlowGraphs = std::set<FlowGraph*, SortByAddress>;
```

Additionally copying should be disabled too or given explicit ownership semantics

```
BinDiffView(const BinDiffView&) = delete;
BinDiffView& operator=(const BinDiffView&) = delete;
```

plugins/bindiff_similarity/provider.cpp
```diff
@@ -0,0 +1,344 @@
...
+				std::move(primaryRender), std::move(secondaryRender)});
+		}
+		if (fixedPoints.empty())
+			completion.SetProgress(progressQuery, edgeProgress(0.95));
+	}
+
+	for (const auto& match : pendingMatches)
+	{
+		const SimilarityResultId primaryResult =
+			results.AddResult(match.key.primary, match.key.secondary, match.similarity, 255);
```
plafosse, line 197: Couple of questions here:

* Whats the reason these are getting Max confidence this seems wrong. Could you just store confidence in pendingMatches?
* Please don't use magic values like this prefer named constants.

plugins/bindiff_similarity/processor.cpp
```diff
@@ -0,0 +1,426 @@
...
+		m_impl->m_functionStarts.insert(function->GetStart());
+
+	for (const auto& segment : view.GetSegments())
+	{
+		const uint64_t segmentAddr = segment->GetStart();
+		const uint64_t segmentLen = segment->GetLength();
+		const int permissions = GetSegmentPermissions(*segment);
+
+		// Map the segment without reading its contents.
+		m_impl->m_flagSpace.AddMemoryBlock(segmentAddr, AddressSpace::MemoryBlock(segmentLen), permissions);
```
plafosse, line 106: We should not be doing this for non-file backed segments which can be arbitrarily large. MemoryBlock will allocate a byte for every segmentLen.

plugins/bindiff_similarity/provider.cpp
```diff
@@ -0,0 +1,344 @@
...
+				security::bindiff::ResetMatches(&secondary);
+			}
+		} reset {primaryView.m_flowGraphs, secondaryView.m_flowGraphs};
+
+		security::bindiff::FixedPoints fixedPoints;
+		security::bindiff::MatchingContext context(primaryView.m_callGraph, secondaryView.m_callGraph,
+			primaryView.m_flowGraphs, secondaryView.m_flowGraphs, fixedPoints);
+
+		// TODO: Expose options for diffing.
+		security::bindiff::Diff(&context, security::bindiff::GetDefaultMatchingSteps(),
```
plafosse, line 110: Non-blocking but it would be great if we could plumb our `completion.IsStopRequested()` into `Diff` as this is where I think the bulk of the time is going to be spent.  Maybe its worth keeping track of this as a separate issue?

plugins/bindiff_similarity/plugin.cpp
```diff
@@ -0,0 +1,66 @@
...
+		}
+		else if (inputFilename.empty())
+		{
+			LogError("Cannot export an unsaved view to BinExport in headless mode");
+			return;
+		}
+
+		bool success;
+		{
+			BinDiffProcessor processor(*view);
```
plafosse, line 38: Non-blocking but I feel like this should really be done off the main thread with a progress callback. Perhaps another issue we should create.

docs/guide/binexport.md
```diff
@@ -1,55 +1,43 @@
...
+PluginCommand.get_valid_list(context)["BinExport"].execute(context)
 ```
 
-### Disabling
+The source view must have a filename when the command is run headlessly.
+
+## Built-in similarity
+
+The **Google BinDiff** provider can also compare binaries directly in a
+[similarity session](./similarity.md), without first creating intermediate BinExport files.
```
plafosse, line 35: I trust `similarity.md` is in another PR?

plafosse (changes requested):

plafosse:
I think the only two blocking requests are the memory exhaustion and the memory leak issues.

emesare:
Added with `6.0.10489`, feedback was addressed in `8e0dfa5f12c9667a6271d3541ff6c557663c874b`

PR #8388, opened 2026-08-04
Add metric resolver for binary similarity (API)

emesare:

This adds a plugin for the metric resolver, which is meant to automatically apply annotations and resolve matches in a binary similarity session, the metric resolver utilizes the expose **similarity** and **confidence** scores to resolve results above given thresholds. This acts as our first resolver and easiest to understand, in the future we may instead have resolvers which actually look at the rendered diffs for example.

The sister PR is located here: https://github.com/Vector35/binaryninja/pull/1745

Depends on:
 - https://github.com/Vector35/binaryninja/pull/1745
 - https://github.com/Vector35/binaryninja-api/pull/8383

### DIFF
--- a/base/strong_typedef.h
+++ b/base/strong_typedef.h
@@ -22,6 +22,7 @@
 
 #include <compare>  // IWYU pragma: keep
 #include <stddef.h>
+#include <concepts>
 #include <functional>
 #include <type_traits>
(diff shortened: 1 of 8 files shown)


emesare:
Added with `6.0.10489`

PR #8389, opened 2026-08-04
Add WARP similarity provider

emesare:

Adds a first party integration with WARP meant for exact function matching.

See https://github.com/Vector35/binaryninja-api/pull/8387 for another similarity provider instead written in C++.

Depends on:
 - https://github.com/Vector35/binaryninja-api/pull/8384

### DIFF
--- a/base/strong_typedef.h
+++ b/base/strong_typedef.h
@@ -22,6 +22,7 @@
 
 #include <compare>  // IWYU pragma: keep
 #include <stddef.h>
+#include <concepts>
 #include <functional>
 #include <type_traits>
 #include <utility>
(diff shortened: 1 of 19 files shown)


psifertex:
One note -- it doesn't look like the quick setup supports enterprise paths very well.
image


Likewise, the "browse" should be using the enterprise picker by default when available but it only supports the local file browser.

Also, the "add file" button in the main view.

psifertex:
once I've done an analysis there's no "reset" but somehow the matching 
Screenshot 2026-08-07 at 1 33 45 AM
is stuck and I can't change them even after closing the previous files?

emesare:
Added with `6.0.10489`

PR #8391, opened 2026-08-05
Fix Python 3.14 custom demangler dispatch

plafosse:
This does two things. Most importantly it fixes a run time error which occurred in the legacy demangler plugin detection code when run under python 3.14. Python 3.13 and below were fine. Second it moves the legacy detection code up to registration improving perf.

### DIFF
--- a/python/__init__.py
+++ b/python/__init__.py
@@ -21,6 +21,7 @@
 import atexit
 import sys
 import ctypes
+import inspect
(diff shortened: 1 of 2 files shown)


fuzyll (approved):
Thanks for fixing this up! I've confirmed that, with the core change to make this work with a unit test, this no longer breaks on 3.14. Approved. 🙂

PR #8393, opened 2026-08-06
Recover implicit this parameters in GNU3 symbols

plafosse:
## Summary

- make GNU3 demangling instance-biased for ambiguous scoped functions
- recover an implicit `this` parameter pointing to the structured enclosing type
- omit `this` for symbols known to be non-members
- preserve instance recovery for class methods nested inside anonymous namespaces
- update direct and generated GNU3 demangler expectations

## Rationale

GNU3 symbols generally do not encode whether an ambiguous scoped function is static. The previous implementation only emitted `this` when it could prove one existed, which omitted it from ordinary member functions. This changes the fallback to assume an instance method unless the symbol provides evidence that it cannot be one. Ambiguous static members and namespace-scoped functions intentionally follow that fallback.

## Validation

- compiled as part of a full `RelWithDebInfo` Binary Ninja standalone application build with UI and plugins
- built the GNU3 direct and generated unit-test targets

### DIFF
--- a/demangler/gnu3/demangle_gnu3.cpp
+++ b/demangler/gnu3/demangle_gnu3.cpp
@@ -288,6 +288,19 @@ namespace
 	}
 
 
+	bool IsStaticOnlyMemberFunction(BNNameType nameType)
+	{
+		return nameType == OperatorNewNameType || nameType == OperatorNewArrayNameType ||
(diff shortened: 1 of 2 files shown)


plafosse:
Merged in 5.4.10332 `cbfc477e880df1deb6bf8ee062b0f9eb0d26ab4d`

PR #8395, opened 2026-08-08
Fix a use-after-free in the GNU3 demangler when expanding a template parameter pack

bdash:
The code was passing a member variable to a function as a const reference. The function reassigned the same member variable within its body, causing its parameter to point to freed memory.

### DIFF
--- a/demangler/gnu3/demangle_gnu3.cpp
+++ b/demangler/gnu3/demangle_gnu3.cpp
@@ -2194,7 +2194,8 @@ bool DemangleGNU3::TryDemangleTemplateParamExpressionPackExpansion(string& expr,
 }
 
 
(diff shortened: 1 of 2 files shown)


plafosse:
I asked codex to repo and came up with this:

```
MallocScribble=1 MallocPreScribble=1 python3 -c "import binaryninja; from binaryninja import demangle; arch=binaryninja.Architecture['x86_64']; demangle.demangle_gnu3(arch, '__Z3fooIJiiiiiiiiiiiiiiiiiiiiiiiiiiEEv1BIJDpT_EE')"
```

bdash:
The internal change adds a unit test that reproduces 100% of the time on Linux.

plafosse:
Oh I totally missed that! Cool.

plafosse (approved):

PR #8401, opened 2026-08-10
Add ARMv7 vabs and fix vcvt lifting

zznop:
This PR adds the following:

1. Lift scalar `vabs` directly with `FloatAbs`
2. Use an intrinsic to lift `vabs` on packed/vector forms that operate on lanes
3. Fix a bug in the existing `vcvt` sizing for instructions such as `vcvt.f64.f32` that were previously performing an incorrect conversion
4. Handle the `vcvt` fixed-point family for lane-wise fixed-point conversions using an intrinsic
5. Continue lifting scalar `vcvt` directly with `fconvert`

All changes have been verified against these two binaries with tests added to `test_lift.py`

[vabs-vcvt-arm.elf.zip](https://github.com/user-attachments/files/30908326/vabs-vcvt-arm.elf.zip)
[vabs-vcvt-thumb.elf.zip](https://github.com/user-attachments/files/30908325/vabs-vcvt-thumb.elf.zip)

image


plafosse (approved):
new code looks perfect

PR #8406, opened 2026-08-11
[DWARF] Avoid reprocessing types repeated in every compilation unit

bdash:
Placeholders are built once per type rather than per declaration, and a definition already committed under its name is not committed again.

This decreases the number of types that the DWARF processor attempts to define on the view by two orders of magnitudes in some cases. Previously the duplicates were discarded by core within `DefineTypes`, but not before expensive work had been performed on each submitted type.

### DIFF
--- a/plugins/dwarf/dwarf_import/src/dwarfdebuginfo.rs
+++ b/plugins/dwarf/dwarf_import/src/dwarfdebuginfo.rs
@@ -692,6 +692,7 @@ impl DebugInfoBuilder {
                 if let Some(target_uid) = debug_type.target_type_uid {
                     if let Some(target_type) = self.get_type(target_uid) {
                         debug_info.add_type(&debug_type_name, &target_type.get_type(), &[]);
+                        type_uids_by_name.insert(debug_type_name, *debug_type_uid);
                     } else {
                         tracing::error!(
                             "Failed to find typedef {} target for uid {}",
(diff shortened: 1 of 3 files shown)


plugins/dwarf/dwarf_import/src/dwarfdebuginfo.rs
```diff
@@ -583,31 +664,43 @@ impl DebugInfoBuilder {
...
+                // We already stored a type with this name and it's a different type, deconflict the name and try again
                 let mut skip_adding_type = false;
-                if stored_debug_type.ty != debug_type.ty {
-                    // We already stored a type with this name and it's a different type, deconflict the name and try again
-                    let mut i = 1;
-                    loop {
-                        if let Some(stored_uid) = type_uids_by_name.get(&debug_type_name) {
-                            if debug_type_uid == stored_uid {
+                let mut i = 1;
+                while stored_debug_type.ty != debug_type.ty {
```
negasora: `stored_debug_type` and `debug_type` aren't modified in the loop body. I'm not sure why this code was changed since afaict it's the same outcome, and the loop condition makes it look like one of those is modified in the loop body and adds an extra type comparison on each iteration.

negasora (changes requested):

plugins/dwarf/dwarf_import/src/dwarfdebuginfo.rs
bdash: Oops. This is remnants of a more complex change I had experimented with that I apparently failed to entirely revert. Fixed now.

negasora (approved):

PR #8407, opened 2026-08-11
[aarch64] Use a more compact representation for instruction decoding state

bdash:
This reduces the amount of time spent zeroing the structs prior to instruction decoding.


arch/arm64/disassembler/decode.h
```diff
@@ -149,347 +149,10 @@ typedef struct context_
...
-	uint64_t width;
-	bool withstatus;
-	uint64_t wmask;
-	bool wnontemporal;
-	uint64_t writeback;
-	uint64_t xs;
-	uint64_t ZAda, ZAd, ZAn, ZAt, Zk, zero_data;
-
+	/* specification scratchpad. The fields are declared in context_fields.h
+	   and packed by gen_ctx_layout.py into context_layout.h so that fields
```
CouleeApps, line 153: Can you commit gen_ctx_layout.py and context_fields.h so that future maintainers can update this struct when necessary?

bdash: It lives in the same repository as the rest of the disassembler's code generation. I sent you a separate PR for that.

CouleeApps (approved):

Issue #7880, opened 2026-01-14
PPC32 and Thumb2 big-endian type libraries have incorrect alternate names


psifertex:
Type libraries for PPC32 and Thumb2 big-endian architectures fail to load automatically because their `alternate_names` don't match what binaries actually import.

  **PPC32**: Binaries import `libc.so.6`, but the type library has `alternate_names: ['libc32.so', 'libc32.so.6']`

  **Thumb2 BE**: Binaries import `libc.so.6`, but the type library has `alternate_names: ['libceb.so', 'libceb.so.6']`

  This causes Binary Ninja to fail to match the type library to imported libraries, so function signatures from libc (printf, malloc, etc.) are not applied.

  ### Affected files

  | Architecture | File | Incorrect | Correct |
  |--------------|------|-----------|---------|
  | ppc | `ld32.so.1.bntl` | `['ld32.so', 'ld32.so.1']` | `['ld.so', 'ld.so.1']` |
  | ppc | `libc32.so.6.bntl` | `['libc32.so', 'libc32.so.6']` | `['libc.so', 'libc.so.6']` |
  | ppc | `libm32.so.6.bntl` | `['libm32.so', 'libm32.so.6']` | `['libm.so', 'libm.so.6']` |
  | ppc | `libpthread32.so.0.bntl` | `['libpthread32.so', 'libpthread32.so.0']` | `['libpthread.so', 'libpthread.so.0']` |
  | thumb2 | `ld-linuxeb.so.3.bntl` | `['ld-linuxeb.so', 'ld-linuxeb.so.3']` | `['ld-linux.so', 'ld-linux.so.3']` |
  | thumb2 | `libceb.so.6.bntl` | `['libceb.so', 'libceb.so.6']` | `['libc.so', 'libc.so.6']` |
  | thumb2 | `libmeb.so.6.bntl` | `['libmeb.so', 'libmeb.so.6']` | `['libm.so', 'libm.so.6']` |
  | thumb2 | `libpthreadeb.so.0.bntl` | `['libpthreadeb.so', 'libpthreadeb.so.0']` | `['libpthread.so', 'libpthread.so.0']` |

  ### Root cause

  The type library generation process incorrectly embedded architecture-specific suffixes ("32" for PPC, "eb" for big-endian) into the alternate names. These suffixes are internal naming conventions and don't reflect what ELF binaries actually reference in their DT_NEEDED entries.

psifertex:
[...]

PR #8409, opened 2026-08-11
Add new API call to remove alternate names from a loaded type library.

UvuvDev:
Added an API call that allows the removal of alternate names from a loaded type library. This is needed if we need to manually patch the type libraries we have to prevent another issue like #7880 .


fuzyll (approved):
I looked at this when I was looking at the other PR. Approved.

Issue #8087, opened 2026-04-14
[DSC] Some Objective-C methods do not have parameter types applied


bdash:
**Version and Platform (required):**
- Binary Ninja Version: 5.3.9434 Ultimate, f7cf98e7
- OS: macos
- OS Version: 26.3.1
- CPU Architecture: arm64

**Bug Description:**
When opening a macOS shared cache I've noticed that some Objective-C methods do not have parameter types applied.

For instance, `-[MLMediaGroup initWithMediaLibrary:attributes:parent:]` from `MediaLibrary.framework` shows up as:
```
id -[MLMediaGroup initWithMediaLibrary:attributes:parent:](int64_t arg1, int64_t arg2, int64_t arg3, int64_t arg4, int64_t arg5)
```

rather than the expected
```
id -[MLMediaGroup initWithMediaLibrary:attributes:parent:](struct MLMediaGroup* self, SEL sel, id mediaLibrary, id attributes, id parent)
```

That breaks type propagation and prevents ivars from being resolved.

That said, methods in `CoreHandwriting.framework`, such as `-[CHMultiScriptTextSynthesizer initWithStyleInventory:]`, do have their argument types applied correctly. I'm not sure exactly what the relevant difference is here.

**Steps To Reproduce:**
1. Load a macOS shared cache from `/System/Volumes/Preboot/Cryptexes/OS/System/Library/dyld/dyld_shared_cache_arm64e`
2. Load MediaLibrary.framework, then navigate to `-[MLMediaGroup initWithMediaLibrary:attributes:parent:]`.

**Additional Information:**

This appears to be a regression in 5.3, due to 1bb00127874e.

psifertex:
Note that the 5.3 R2 release fixed the regression on the stable, however the issue is being left open to track re-applying the previous commit with the fix to avoid this regression.

PR #8411, opened 2026-08-12
[ObjC] Store Objective-C metadata in the database instead of regenerating it on load

bdash:
Processed Obj-C metadata was stored ephemerally on the view and not serialized to the .bndb. On reopen the Obj-C processor had to be re-run on every loaded image to regenerate the metadata, which caused many functions to be reanalyzed due to types and symbols being reapplied. Commit 1bb00127874e tried to suppress that from inside `DefineObjCSymbol` but ended up skipping method type application on fresh shared-cache loads. As a result, methods whose names were already in the symbol table (some methods in macOS shared caches) did not have argument types applied.

We now save the Objective-C metadata to the .bndb and only process Objective-C data in the binary if the metadata is absent or out of date. Additionally, each `ProcessObjCData` call now merges its output into any existing metadata so multiple shared cache images accumulate metadata into a single unified object instead of clobbering each other.

An "Objective-C Literals" metadata key is added with its own version, so future changes to Obj-C literal processing can force reprocessing on existing databases without touching the main Obj-C metadata version.

The changes made to `ObjCProcessor` in 1bb00127874e are reverted as they were incorrect and are no longer needed with this new approach.

Fixes https://github.com/Vector35/binaryninja-api/issues/8087.

### DIFF
--- a/objectivec/objc.cpp
+++ b/objectivec/objc.cpp
@@ -68,6 +68,150 @@ namespace {
 		return Type::NamedType(builder.Finalize());
 	}
 
+	constexpr uint64_t kObjCMetadataVersion = 2;
+	constexpr const char* kObjCMetadataKey = "Objective-C";
+	constexpr uint64_t kObjCLiteralsVersion = 1;
+	constexpr const char* kObjCLiteralsKey = "Objective-C Literals";
+
+	bool HasUpToDateVersion(BinaryView* view, const char* key, uint64_t currentVersion)
+	{
+		auto existing = view->QueryMetadata(key);
+		if (!existing)
+			return false;
+		auto kv = existing->GetKeyValueStore();
+		auto it = kv.find("version");
+		if (it == kv.end())
+			return false;
+		return it->second->GetUnsignedInteger() >= currentVersion;
+	}
+
+	// Classes, categories, and methods are arrays of records keyed on `loc`. Entry locations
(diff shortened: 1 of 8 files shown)


objectivec/objc.cpp
```diff
@@ -1119,11 +1305,6 @@ void ObjCProcessor::GenerateClassTypes()
 {
 	for (auto& [_, cls] : m_classes)
 	{
-		QualifiedName classTypeName = cls.name;
-		std::string classTypeId = Type::GenerateAutoTypeId("objc", classTypeName);
-		if (m_data->GetTypeById(classTypeId))
```
emesare, line 1124: What was the purpose of moving this and removing the check, I assume this means that there is never any duplicate class type names?

bdash: This specific code comes from https://github.com/Vector35/binaryninja-api/commit/1bb00127874e414066fe91943aa938ae414eca2a, and was part of trying to make processing Obj-C runtime metadata twice a no-op.

This new approach uses metadata stored on the view to know that it should entirely skip doing Obj-C processing when loading from a database, rather than trying (and failing) to make it into a no-op to process Obj-C runtime metadata twice.

emesare (approved):

bdash:
GitHub apparently failed to notice that 0dd311b8277d2c17aa62f9f6762b992cbf7628bb was merged to dev.

Issue #32, opened 2015-11-05
python shell


psifertex:
_From @psifertex on May 5, 2015 13:57_

Just like binaryninja-python

_Copied from original issue: Vector35/binaryninja#11_

Manouchehri:
Related: #277.

lucasduffey:
also: ipython shell might be better than standard python shell

plafosse:
Done.

PR #8413, opened 2026-08-14
[ARMv7] Fix full-width bitfield instruction lifting

zznop:
Currently, we construct masks using a 32-bit expression (`1 << width) - 1`). This is valid when `lsb` is zero, but shifting a 32-bit integer by 32 is undefined. This produces a zero mask and causes incorrect LLIL for full-width operations such as `bfi Rd, Rn, #0, #32`, `bfc Rd, #0, #32` and `ubfx Rd, Rn, #0, #32`. 

This change constructs bitfield masks using a 64-bit intermediate before narrowing to 32 bits: `(uint64_t{1} << width) - 1)`.

Before:
Screenshot 2026-08-14 at 12 16 56 PM

After:
Screenshot 2026-08-14 at 12 17 43 PM

I've attached a binary if anyone cares to test.
[thumb2_bitfield_width32.elf.zip](https://github.com/user-attachments/files/31078274/thumb2_bitfield_width32.elf.zip)

### DIFF
--- a/arch/armv7/il.cpp
+++ b/arch/armv7/il.cpp
@@ -2208,12 +2208,12 @@ bool GetLowLevelILForArmInstruction(Architecture* arch, uint64_t addr, LowLevelI
 		case ARMV7_BFC:
 			ConditionExecute(il, instr.cond, SetRegisterOrBranch(il, op1.reg,
 				il.And(get_register_size(op1.reg), ReadRegisterOrPointer(il, op1, addr),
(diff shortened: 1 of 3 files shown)


zznop:
Also credit to @psifertex  for initial triage and finding the issue

plafosse (approved):
Approved but first please use `1ULL` like the rest of the code base instead of `uint64_t{1}`

zznop:
Updated syntax to `1ULL` and merged

PR #8425, opened 2026-08-18
Register ELF and Mach-O loader settings as per-file load settings instead of global

bpotchik:


psifertex:
I'm not a good choice to review this, the original issue was brought up during a code review from an agent I was running but I specifically said at the time I had no real idea if it was even real so probably needs someone else's eyeballs. :-)

PR #8431, opened 2026-08-18
Silence compiler warnings from vendored BinDiff/BinExport code

bdash:
Build the vendored trees and their fetched dependencies with `-w`, and mark the subtree `SYSTEM` so warnings in vendored headers included from the plugin's own sources are suppressed too. The plugin's sources keep the top-level warning flags.


emesare (approved):
Looks good to me so long as it builds

PR #8437, opened 2026-08-19
Fix Python and Rust memory leaks

plafosse:
Fixes the Python and Rust ownership issues reported in Vector35/binaryninja#1506. Ensures owned references and parse results are freed correctly on both success and failure paths.
Verified with Python compilation and Rust formatting/build checks.


zznop (approved):
Your changes look good. However, looks like the C++ `CoreTypeParser::ParseTypeString` still has a failure path that leaks `apiResult`: https://github.com/Vector35/binaryninja-api/blob/test_fix_memory_leaks/typeparser.cpp#L588

plafosse:
Oh good check. Fixed.

plafosse:
Merged in 6.0.10512 `cdfd2aa43196e9ae8a115abf283dc007b4b04046`

PR #8453, opened 2026-08-22
Fix a leak in BinaryReader::ReadVector

saagarjha:
The new behavior requires that T is value initializable but generally you have no business calling this function if that is not the case, so this should be fine.


CouleeApps:
1. Wow good find
2. Any reason to value-initialize instead of reserve() ?

saagarjha:
reserve doesn't change the vector's size which means it mostly just changes the backing buffer without making it available to us to write into. I don't think there is an easy way to uninitialized-fill a vector unfortunately

CouleeApps (approved):

PR #8465, opened 2026-08-25
Eliminate polling latency in BinaryView search generators.

bpotchik:
Remove 100ms polling latency from BinaryView search QueueGenerator

### DIFF
--- a/python/binaryview.py
+++ b/python/binaryview.py
@@ -9976,24 +9976,30 @@ to a the type "tagRECT" found in the typelibrary "winX64common"
 		return result.value
 
 	class QueueGenerator:
+		_done = object()
+
 		def __init__(self, t: threading.Thread, results: queue.Queue):
 			self.thread = t
 			self.results = results
+			self._finished = False
 			t.start()
 
 		def __iter__(self):
 			return self
 
 		def __next__(self):
-			while True:
-				try:
-					return self.results.get(timeout=0.1)
-				except queue.Empty:
-					if not self.thread.is_alive():
-						try:
-							return self.results.get_nowait()
-						except queue.Empty:
-							raise StopIteration
+			if self._finished:
+				raise StopIteration
+			result = self.results.get()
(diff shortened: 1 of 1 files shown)


psifertex:
claude ID'd a hang if you call `next()` after emptying the search results. I don't think it's super likely to trigger in practice with well-formed code but it is technically possible now and wasn't before.

```
┌───────────────────────────────────────────────┬─────────┬────────────┐
│                                               │ with PR │  without   │
├───────────────────────────────────────────────┼─────────┼────────────┤
│ for a, d in bv.search(...) (hit and miss)     │ ok      │ ok         │
├───────────────────────────────────────────────┼─────────┼────────────┤
│ list() / sorted() / any() / zip()             │ ok      │ ok         │
├───────────────────────────────────────────────┼─────────┼────────────┤
│ next(iter(...), None)                         │ ok      │ ok         │
├───────────────────────────────────────────────┼─────────┼────────────┤
│ for ...: break (never exhausted)              │ ok      │ ok         │
├───────────────────────────────────────────────┼─────────┼────────────┤
│ limit=1, take one, then list(g) for leftovers │ ok      │ ok         │
├───────────────────────────────────────────────┼─────────┼────────────┤
│ two for loops over the same generator         │ HANG    │ ok, 210 ms │
├───────────────────────────────────────────────┼─────────┼────────────┤
│ for loop, then a defensive next(g, None)      │ HANG    │ ok, 206 ms │
├───────────────────────────────────────────────┼─────────┼────────────┤
│ two threads draining one generator            │ HANG    │ ok, 105 ms │
└───────────────────────────────────────────────┴─────────┴────────────┘
```

It does fix the performance issue I was running into!

bpotchik:
@psifertex can you re-review. Thanks!

psifertex (approved):
Loos good! Re-ran my tests and while the AI came up with some weird hypothetical edge cases (like trying to drain twice), I think this is good.

PR #8471, opened 2026-08-26
Remove SysV from the Windows kernel platform

plafosse:
Remove the explicit SysV calling-convention registration from windows-kernel-x86_64. SysV remains available at the x86-64 architecture level but is no longer advertised as supported by the Windows-kernel platform.


D0ntPanic (approved):
Looks good

PR #8500, opened 2026-09-01
Improve MIPS floating-point lifting and calling conventions

### LINKED ISSUE #4010: sdc1/ldc1 not lifted on MIPS32
sdc1 does a double precision (8-byte) store to memory from a floating point register, and ldc1 does a double precision load from memory. On MIPS the floating point registers can be accessed from the even number registers ($f0, $f2, $f4, ...) to get an 8-byte double precision value spanning 2 registers (eg: $f0 and $f1). It is also common to use these instructions to save the two registers to the stack at the beginning of a function and to restore them later in a single command. 

Currently these two instructions are unimplemented when lifting to LLIL. It should be possible to patch in my own LLIL with an ArchitectureHook, but I can't find a reasonable way to do this due to Vector35/binaryninja-api#1451 and Vector35/binaryninja-api#1452. 

Setting the IL to do a reg_split between the two registers doesn't work due to Vector35/binaryninja-api#1452. In that example, even though I have $f20 and $f21 set as callee_saved_regs BN doesn't detect that IL instruction as saving both $f20 and $f21. 

Since it doesn't know they are being saved BN incorrectly detects them as input parameters into the function and that gets passed up the call tree.

I could do separate reads and writes to the registers and that does get detected as saving the registers, but when handling an actual double precision number outside of the beginning/end of the function that would break the lifting since the 8-byte value would be broken in half and not shown nicely. 

I thought about modifying the registers in the mips32 architecture to create an 8-byte wide register and to set the existing registers as subregisters of that. I could then lift that 8-byte wide register as being stored/loaded to/from memory. But due to Vector35/binaryninja-api#1451 I can't seem to modify registers in an ArchitectureHook to let me do this.

Is there another way I can try working around these unimplemented instructions? Or can these instructions be implemented in the core architecture?

zznop:
This PR includes the following:
- Add o32 hard-float calling convention support.
- Expand MIPS floating-point lifting for loads/stores, register transfers, comparisons, multiply-add, and conditional moves.
- Correctly handle FR=0 register pairs, FR=1 registers, FCC conditions, endianness, and unpredictable cases.
- Add regression tests (`test_lifting.py`) for fixed and newly lifted instructions

Closes #4010 

After/Before:
Screenshot 2026-09-01 at 10 07 24 AM


arch/mips/arch_mips.cpp
```diff
@@ -2796,6 +2815,217 @@ class MipsO32CallingConvention: public CallingConvention
...
+		return DefaultIsReturnTypeRegisterCompatible(type);
+	}
+
+	virtual ValueLocation GetReturnValueLocation(BinaryView* view, const ReturnValue& returnValue) override
+	{
+		Ref<Type> type = returnValue.type.GetValue();
+		if (!type || type->IsVoid())
+			return ValueLocation();
+
+		if (type->GetClass() == NamedTypeReferenceClass && type->GetWidth() == 0)
```
rssor: do we have DerefNamedType available over the API?

rssor (approved):

arch/mips/arch_mips.cpp
zznop: Yes it's available. I just updated the code to use it.

PR #8508, opened 2026-09-02
Fix cache UI actions without UI context

plafosse:
Fix cache image-loading actions invoked without an active UI context, preventing null dereferences during background execution. Also applies the same protection to dyld shared-cache actions.

Repro:

First select the unloaded green address then run:

```
from binaryninja import execute_on_main_thread_and_wait
from binaryninjaui import UIContext

def repro_bnkc4():
    ui = UIContext.activeContext()
    view = ui.getCurrentViewFrame().getCurrentViewInterface()
    handler = view.actionHandler()
    ctx = view.actionContext()

    assert handler.isValidAction("KC Load IMGHERE", ctx), \
        "Select an unloaded off-image address first"

    ctx.context = None
    handler.executeAction("KC Load IMGHERE", ctx)

execute_on_main_thread_and_wait(repro_bnkc4)
```

This should fix issue: https://github.com/Vector35/binaryninja/issues/1898

### DIFF
--- a/view/kernelcache/ui/KernelCacheUINotifications.cpp
+++ b/view/kernelcache/ui/KernelCacheUINotifications.cpp
@@ -108,17 +108,25 @@ void UINotifications::OnViewChange(UIContext* context, ViewFrame* frame, const Q
 		uint64_t addr = 0;
 		if (GetAddressInput(addr, "Address", "Address"))
 		{
-			BackgroundThread::create(ctx.context->mainWindow())
-				->thenBackground([ctx, addr]() {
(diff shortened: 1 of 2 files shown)


zznop (approved):
These changes look good to me. Looks like all cache action paths are addressed and it prevents dereferencing a null context pointer. When the context is NULL I assume that it's possible work can outlive the UI. But, I think we permit that elsewhere.

bdash (approved):
This seems fine as a defensive fix, but we should file a follow-up issue about the fact that a context menu item invoked an action with a null UIContext.

PR #8524, opened 2026-09-06
[ObjC] Enable removal of more memory management functions

AngeloD2022:
- Added `_objc_claimAutoreleasedReturnValue` to the list of ignorable memory management functions.
- Added normalization for the name of register-specific ARC functions, like `_objc_release_x19`.
- Added a more robust runtime function check for dyld shared cache binaries.

### DIFF
--- a/plugins/workflow_objc/src/activities/remove_memory_management.rs
+++ b/plugins/workflow_objc/src/activities/remove_memory_management.rs
@@ -29,6 +29,7 @@ const IGNORABLE_MEMORY_MANAGEMENT_FUNCTIONS: &[&[u8]] = &[
     b"_objc_retainAutoreleasedReturnValue",
     b"_objc_retainBlock",
     b"_objc_unsafeClaimAutoreleasedReturnValue",
+    b"_objc_claimAutoreleasedReturnValue",
 ];
 
 fn is_call_to_ignorable_memory_management_function<'func>(
--- a/plugins/workflow_objc/src/activities/remove_memory_management.rs
+++ b/plugins/workflow_objc/src/activities/remove_memory_management.rs
@@ -1,3 +1,5 @@
+use crate::activities::util;
+use crate::{error::ILLevel, metadata::GlobalState, Error};
 use binaryninja::{
     architecture::{Architecture as _, CoreRegister, Register as _, RegisterInfo as _},
     binary_view::BinaryView,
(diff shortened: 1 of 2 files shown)


plugins/workflow_objc/src/activities/remove_memory_management.rs
```diff
@@ -29,8 +30,51 @@ const IGNORABLE_MEMORY_MANAGEMENT_FUNCTIONS: &[&[u8]] = &[
     b"_objc_retainAutoreleasedReturnValue",
     b"_objc_retainBlock",
     b"_objc_unsafeClaimAutoreleasedReturnValue",
+    b"_objc_claimAutoreleasedReturnValue",
 ];
 
+fn is_objc_rt_symbol_dscview(view: &BinaryView, symbol: &Symbol) -> bool {
```
bdash, line 36: Can you clarify what the purpose of this function is? What is it guarding against?

plugins/workflow_objc/src/activities/remove_memory_management.rs
```diff
@@ -57,7 +101,17 @@ fn is_call_to_ignorable_memory_management_function<'func>(
     // Remove any j_ prefix that the shared cache workflow adds to stub functions.
     let symbol_name = symbol_name.strip_prefix(b"j_").unwrap_or(symbol_name);
 
-    IGNORABLE_MEMORY_MANAGEMENT_FUNCTIONS.contains(&symbol_name)
+    // Normalize the name to also include register-specific functions (e.g. _objc_release_x19).
+    let symbol_name = util::strip_arc_reg_suffix(symbol_name);
```
bdash, line 105: This results in incorrect handling of `objc_retain_xN` calls. Converting them to a no-op will result in `x0` having the incorrect value.

Consider this example:

Image

This change results in:

Image


It should be `result->_internal = _initWithCFURLResponse[1]`

plugins/workflow_objc/src/activities/remove_memory_management.rs
AngeloD2022: I added that check just to be as precise as possible and reduce the reliance on only symbol names– because if I'm not mistaken, it's possible in principle for multiple different symbols with the same name to exist in a binary view. So although it's unlikely– if Apple includes a function in one of their shared cache libraries with the same name as a runtime memory management function, it won't be hidden like one.

AngeloD2022:
Fixed by #8535

PR #8525, opened 2026-09-06
[Pseudo-ObjC] Reconstruct `@autoreleasepool` and `@synchronized(...)` blocks

AngeloD2022:
Translate autorelease pool runtime function calls to `@autoreleasepool {...}` blocks.

Original Objective-C snippet:
```objectivec
for (NSUInteger itemNo = 1; itemNo <= 3; itemNo++) {
    @autoreleasepool {
        NSString *item = [NSString stringWithFormat:@"batch %lu, item %lu",
                            (unsigned long)batchNo,
                            (unsigned long)itemNo];
        NSString *displayItem = [item uppercaseString];
        [items addObject:displayItem];
    }
}
```

Pseudo-C snippet:

```c
for (int64_t i = 1; i <= 3; i += 1)
{
    void* context = _objc_autoreleasePoolPush();
    int64_t x0;
    int64_t var_80_1 = x0;
    i_1 = i;
    id location_2 = _objc_retainAutoreleasedReturnValue(_objc_msgSend(
        _OBJC_CLASS_$_NSString, "stringWithFormat:", &cfstr_batch_%lu,_item_%lu));
    int64_t location_1 = _objc_retainAutoreleasedReturnValue(_objc_msgSend(
        location_2, "uppercaseString"));
    _objc_msgSend(var_20, "addObject:", location_1);
    _objc_storeStrong(&location_1, nullptr);
    _objc_storeStrong(&location_2, nullptr);
    _objc_autoreleasePoolPop(context);
}
```

Equivalent Pseudo-Objective-C:

```objectivec
for (int64_t i = 1; i <= 3; i += 1)
{
    @autoreleasepool
    {
        int64_t x0;
        int64_t var_80_1 = x0;
        i_1 = i;
        id location_2 =
            [[NSString stringWithFormat:@"batch %lu, item %lu"] retain];
        int64_t location_1 = [[location_2 uppercaseString] retain];
        [var_20 addObject:location_1];
        _objc_storeStrong(&location_1, nullptr);
        _objc_storeStrong(&location_2, nullptr);
    }
}
```

### DIFF
--- a/lang/c/pseudoc.cpp
+++ b/lang/c/pseudoc.cpp
@@ -751,6 +751,161 @@ void PseudoCFunction::GetExprText(const HighLevelILInstruction& instr, HighLevel
 }
 
 
+void PseudoCFunction::GetExpr_VAR_INIT(
+	const HighLevelILInstruction& instr, HighLevelILTokenEmitter& tokens, DisassemblySettings* settings, bool statement)
+{
+	const auto srcExpr = instr.GetSourceExpr<HLIL_VAR_INIT>();
+	const auto destExpr = instr.GetDestVariable<HLIL_VAR_INIT>();
+
+	const auto variableType = GetHighLevelILFunction()->GetFunction()->GetVariableType(destExpr);
+	const auto platform = GetHighLevelILFunction()->GetFunction()->GetPlatform();
+	const auto prevTypeTokens = variableType.GetValue() ?
+		GetTypePrinter()->GetTypeTokensBeforeName(variableType.GetValue(), platform, variableType.GetConfidence()) :
+		vector<InstructionTextToken> {};
+	const auto postTypeTokens = variableType.GetValue() ?
+		GetTypePrinter()->GetTypeTokensAfterName(variableType.GetValue(), platform, variableType.GetConfidence()) :
+		vector<InstructionTextToken> {};
+
+	// Check to see if the variable appears live
+	bool appearsDead = false;
+	if (const auto ssaForm = instr.GetSSAForm(); ssaForm.operation == HLIL_VAR_INIT_SSA)
+	{
+		const auto ssaDest = ssaForm.GetDestSSAVariable<HLIL_VAR_INIT_SSA>();
+		appearsDead = !GetHighLevelILFunction()->IsSSAVarLive(ssaDest);
+	}
+
+	// If the variable does not appear live, show the assignment as zero confidence (grayed out)
+	if (appearsDead)
+		tokens.BeginForceZeroConfidence();
+
+	if (variableType.GetValue())
+	{
+		for (auto typeToken : prevTypeTokens)
+		{
+			typeToken.context = LocalVariableTokenContext;
+			typeToken.address = destExpr.ToIdentifier();
+			tokens.Append(typeToken);
+		}
+		tokens.Append(TextToken, " ");
+	}
+	tokens.AppendVarTextToken(destExpr, instr, instr.size);
+	if (variableType.GetValue())
+	{
+		for (auto typeToken : postTypeTokens)
+		{
+			typeToken.context = LocalVariableTokenContext;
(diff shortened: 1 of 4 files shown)


lang/c/pseudoc.cpp
```diff
@@ -751,6 +751,161 @@ void PseudoCFunction::GetExprText(const HighLevelILInstruction& instr, HighLevel
...
+		const size_t consumedStmts = TryEmitBlockRegion(instr, exprs, i, tokens, settings);
+		if (consumedStmts != 0)
+		{
+			i += consumedStmts;
+			tokens.NewLine();
+			continue;
+		}
+
+		const auto& statement = exprs[i];
+		EmitStandardBlockStatement(instr, statement, isFirst, isLast, needSeparator, tokens, settings);
```
bdash: The code within `EmitStandardBlockStatement` assumes that `isLast` means the statement is the final statement in the block. This pairs it with the `block.exprIndex == root` check to detect and suppress a trailing void `return`. After this refactoring, `isLast` now only means it's the final statement in the span that was passed to `EmitBlockStatements` as `exprs`. `PseudoObjCFunction::TryEmitBlockRegion` passes a subspan of the block's instructions, which violates that assumption.

I'm not entirely sure how this would manifest in practice, but it makes the code harder to reason about.

lang/c/pseudoc.cpp
```diff
@@ -751,6 +751,161 @@ void PseudoCFunction::GetExprText(const HighLevelILInstruction& instr, HighLevel
...
+		tokens.AppendSemicolon();
+}
+
+
+void PseudoCFunction::EmitStandardBlockStatement(
+	const HighLevelILInstruction& block,
+	const HighLevelILInstruction& statement,
+	bool isFirst,
+	bool isLast,
+	bool& needsSeparator,
```
bdash: This being an output parameter is easy to miss.

bdash (changes requested):
Thanks for the PR! The `@autoreleasepool` rendering is a nice addition. Factoring the handling of `HLIL_BLOCK` so it can be overridden by Obj-C rendering seems like generally the right shape.

The autorelease region matching needs some work, though.

The fundamental issue is that `TryEmitBlockRegion` only ever looks at the immediate children of a single block. Nothing requires that a pool's pops are all siblings of the push. An autorelease pool containing a conditional `return`, or a conditional `break` or `continue` when the pool is inside a loop, will result in multiple pops for a single push, with at least one at a deeper level. Those nested pops are missed by the current approach, so they'll survive into the output despite referencing a variable that is no longer visible:

```objc
while (true)
{
    @autoreleasepool
    {
        ...
        if (cond)
        {
            // What is `context` that this refers to?
            _objc_autoreleasePoolPop(context);
            break;
        }
    }
    ...
}
```

I think the shape you want is to pair pops with pushes using the pool handle rather than by position, and then elide any pop of a handle whose region is currently open, wherever it appears inside that region.

bdash:
This snippet, compiled with `cc -fobjc-arc -Os -o test test.m`, demonstrates the issue I described with pops that are not siblings of the push:
```objc
#import <Foundation/Foundation.h>

int main(int argc, char** argv)
{
    for (id obj in NSProcessInfo.processInfo.environment)
    {
        @autoreleasepool {
            if (!obj)
                break;
            NSLog(@"%@", obj);
        }
    }
}
```

AngeloD2022:
Appreciate the review, Mark! I'll investigate these tomorrow as soon as I can.

AngeloD2022:
I made `EmitBlockStatements` focus solely on the span of statements provided  to it, and moved the void return check up from `EmitStandardBlockStatement` to the loop inside of `EmitBlockStatements`. Part of the test happens in `GetExpr_BLOCK`, where the block instruction is in scope, and is passed through the `EmitBlockStatements` `isBlockRoot` parameter.

PR #8535, opened 2026-09-09
[ObjC] Handle objc_{retain,release}_xYY when removing reference counting calls

bdash:
These are slightly tricky as `objc_retain_xN` takes its argument in `xN` and returns it in `x0`. This means it cannot simply be replaced with a `nop` like we do with `objc_retain`. It is instead replaced with an assignment.

### DIFF
--- a/plugins/workflow_objc/src/activities/remove_memory_management.rs
+++ b/plugins/workflow_objc/src/activities/remove_memory_management.rs
@@ -1,5 +1,5 @@
 use binaryninja::{
-    architecture::{Architecture as _, CoreRegister, Register as _, RegisterInfo as _},
+    architecture::{Architecture as _, CoreRegister, Register as _, RegisterId, RegisterInfo as _},
     binary_view::BinaryView,
     low_level_il::{
         expression::{ExpressionHandler, LowLevelILExpressionKind},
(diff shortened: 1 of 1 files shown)


plugins/workflow_objc/src/activities/remove_memory_management.rs
```diff
@@ -11,91 +11,158 @@ use binaryninja::{
...
+    // The `objc_retain_xN` / `objc_release_xN` variants take their object in `xN` rather than
+    // `x0`. The target's parameter and return value locations name the registers.
+    let target_function = view.function_at(platform, target)?;
+    let arch = target_function.arch();
+    let object_register = target_function
+        .parameter_variables()
+        .contents
+        .first()
+        .filter(|var| var.ty == VariableSourceType::RegisterVariableSourceType)
+        .and_then(|var| arch.register_from_id(RegisterId::from(var.storage as u32)))?;
```
emesare, line 95: For later we should add some helper to make this pattern of getting the backing register from a variable easier. Not apart of this PR.

emesare (approved):
lgtm, small comment about potentially making the "get arch register from var" pattern a helper

PR #8537, opened 2026-09-09
[MIPS] Expand floating-point, coprocessor, and shuffle lifting

zznop:
- Add scalar and paired-single FP arithmetic and conditional moves
- Lift TRUNC/ROUND/CEIL/FLOOR.W/L.S/D with fixed rounding semantics
- Model paired-single arithmetic and per-lane conditional moves with intrinsics
- Decode and lift ALIGN/DALIGN and BITSWAP/DBITSWAP
- Lift JALX, coprocessor zero-test branches, and PREFX
- Model COP2 operations, control/high-word transfers, and shadow GPR access with intrinsics
- Preserve branch inputs across conditional and split-register delay-slot writes
- Expand lifting regressions for endianness, register modes, aliasing, zero registers, and delay slots

After/Before:
Screenshot 2026-09-09 at 4 02 07 PM


zznop:
Here's a binary I made for testing / validating these changes (in addition to the regression tests)

[mips32-lifting-be.elf.zip](https://github.com/user-attachments/files/32026145/mips32-lifting-be.elf.zip)

rssor (approved):

Issue #1502, opened 2020-01-03
Add clickable references to comments (Feature request)


Martyx00:
It would be awesome if we can have clickable references in inline comments in graph/linear views as we have in log.

Ideally it should allow some notation like @0x123456, @_main or something similar.

jeffli678:
closed in favor of #1845

jeffli678:
emm, we are still missing clickable symbol/function names

PR #8538, opened 2026-09-09
Fix multiple PE/COFF parsing bugs (#1502)

Weitao-Sun:
Fixes three PE/COFF parsing issues reported in Vector35/binaryninja#1502 (referenced there as Ghidra #9169, #9170, #9171):
- COFF symbol table OOM via unbounded symbol count and string names
- PointerToRawData not rounded to 0x200 per Windows loader behavior
- SizeOfRawData not rounded up to FileAlignment, hiding code caves

This is the api-repo counterpart of Vector35/binaryninja#1625, which will be updated to just bump the `api` submodule pointer once this merges.

## Review feedback addressed

An earlier version of this change went through a round of review. Summary of what changed as a result:

1. `virtualSize` is grown to `max(virtualSize, sizeOfRawData)` so segments, sections, RVA characteristics, and symbol placement all agree with the corrected raw-data extent.
2. The symbol-name budget is checked before each string-table read instead of after, names are cached by string-table offset so repeated offsets are read once, and the running total accounts for the extra copies a resolved name ends up retained in.
3. The aligned raw-data size is clamped to `UINT32_MAX` before narrowing to `uint32_t`.
4. `coffSymbolCount * sizeofCOFFSymbol` arithmetic is widened to 64-bit in both files, and the COFF section-name-resolution path uses the runtime symbol-record size instead of a hardcoded 18.
5. Section-name and symbol-name string table offsets are validated against the declared table length before being read.
6. All six loader settings (PE + COFF, count / name-length / total-budget) now treat 0 as uncapped.
7. Magic numbers moved into named constants in `peview.h`, shared by both files.
8. `header.coffSymbolCount` is no longer mutated. Every raw symbol-table slot still gets a data variable and marker symbol, which relocation resolution looks up by address; the configurable count limit only gates the more expensive per-symbol work (name resolution, typing, aux record definitions).

## Follow-up round

A second look raised four more points:

- Sector rounding was unconditional and didn't account for low-alignment PE images. Fixed in the latest commit: when `SectionAlignment < page size` and `SectionAlignment == FileAlignment` (the PE spec's low-alignment case), raw offsets map directly to RVAs without sector rounding. Checked against a synthetic low-alignment image (`SectionAlignment = FileAlignment = 0x20`, `PointerToRawData = 0x220`) — the section's file offset and content now match the declared layout instead of getting rounded down to `0x200`.
- Symbol-count limits discarding valid relocations: with `maxCoffSymbolCount = 1`, 2 symbols, and a relocation referencing symbol index 1, the marker for index 1 is still created and the relocation resolves — this is what point 8 above covers.
- Rounded raw bytes ending up outside the mapped section: covered by the `virtualSize = max(virtualSize, sizeOfRawData)` change above.
- Raw-size rounding wrapping past 4 GiB to zero: the clamp to `UINT32_MAX` happens before narrowing (point 3 above), so this doesn't wrap.

If any of the last three still show up against this branch's current tip, a repro would help track it down further.

## Known follow-up (not blocking)

The per-slot marker-symbol loop in the COFF symbol table walk (point 8) still iterates the full declared `coffSymbolCount` even when the count limit caps the heavier per-symbol work, so a very large declared count still costs a linear pass just to place markers. A bound derived from the file's actual size (a legitimate count can't exceed what fits in the remaining file) would close this without reintroducing the relocation-lookup issue point 8 fixed. Can follow up separately if useful.

## Testing

This changes PE section-layout logic, so part of the local analysis-output test suite differs from its stored oracles — expected, since section boundaries legitimately shift for several PE test corpus binaries. The full suite gives 81 differences against this branch; running the same suite against plain `dev` (no commits from this branch) narrows it down:

- 23 already show up on plain `dev` — unrelated to this change (`test_Architecture`, `test_memory_leaks`, `test_struct_return_win32`/`win64`, PDB tests, plus a handful of non-layout tests on `acpipagr.sys`/`acpitime.sys`).
- 58 are specific to this branch, on the PE/COFF corpus binaries the change touches (`acpipagr.sys`, `acpitime.sys`, `jumptable_multiple_indirect`, `partial_register_dataflow`(`_rebasing`), `pe_thumb`, `quick3dcoreplugin.dll`, `win32_x86.dll`, `win32_x86_warp.dll`). Same set with or without the latest (low-alignment) commit. These are the layout-shift kind — uniform address/length changes and/or new, well-formed content in the newly-exposed byte range.

Will regenerate oracles for the 58 once the code here settles, scoped to just those binaries.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/view/pe/coffview.cpp
+++ b/view/pe/coffview.cpp
@@ -8,6 +8,7 @@
 #include <mutex>
 #include <sstream>
 #include <type_traits>
+#include <unordered_map>
 #include <utility>
 #include "coffview.h"
 
@@ -258,21 +259,25 @@ bool COFFView::Init()
 				if (errno == 0 && offset > 0)
 				{
 					BinaryReader stringReader(GetParentView(), LittleEndian);
-					uint64_t stringTableBase = header.coffSymbolTable + (header.coffSymbolCount * 18);
+					// Compute the string table offset using 64-bit arithmetic and the
+					// actual per-symbol record size (18 bytes normally, 20 for BigCOFF).
+					uint64_t stringTableBase = header.coffSymbolTable + ((uint64_t)header.coffSymbolCount * sizeofCOFFSymbol);
 					stringReader.Seek(stringTableBase);
 					uint32_t stringTableLen = stringReader.Read32();
 					if ((stringTableBase + stringTableLen) > GetParentView()->GetEnd())
 					{
 						m_logger->LogError("Cannot resolve section name \"%s\": String table is invalid length", name);
 					}
-					else if (stringTableBase + offset < GetParentView()->GetEnd())
+					else if (offset < stringTableLen)
 					{
 						sectionNameReader.Seek(stringTableBase + offset);
-						resolvedName = sectionNameReader.ReadCString();
+						// Section names longer than 1024 bytes are not meaningful; cap the read
+						// to bound the allocation.
+						resolvedName = sectionNameReader.ReadCString(1024);
 					}
 					else
 					{
-						m_logger->LogError("Cannot resolve section name \"%s\": Offset is past end of string table", name);
+						m_logger->LogError("Cannot resolve section name \"%s\": Offset %u exceeds the string table size %u", name, offset, stringTableLen);
 					}
 				}
 			}
@@ -866,6 +871,29 @@ bool COFFView::Init()
 			// TODO: combine the aux symbol record struct types into a union:
 			// StructureBuilder coffAuxSymbolRecordBuilder(UnionStructureType);
 
+			// A limit of 0 disables the corresponding check.
+			uint64_t maxSymCount = PE_DEFAULT_MAX_COFF_SYMBOL_COUNT;
+			uint64_t maxSymNameLen = PE_DEFAULT_MAX_COFF_SYMBOL_NAME_LENGTH;
+			uint64_t maxTotalSymNameBytes = PE_DEFAULT_MAX_TOTAL_COFF_SYMBOL_NAME_MB * 1024 * 1024;
+			if (settings && settings->Contains("loader.coff.maxCoffSymbolCount"))
+				maxSymCount = settings->Get<uint64_t>("loader.coff.maxCoffSymbolCount", this);
+			if (settings && settings->Contains("loader.coff.maxCoffSymbolNameLength"))
+				maxSymNameLen = settings->Get<uint64_t>("loader.coff.maxCoffSymbolNameLength", this);
+			if (settings && settings->Contains("loader.coff.maxTotalCoffSymbolNameBytes"))
+				maxTotalSymNameBytes = settings->Get<uint64_t>("loader.coff.maxTotalCoffSymbolNameBytes", this)
+				                       * 1024 * 1024;
+			// A name length limit of 0 means no limit; ReadCString takes an actual byte count,
+			// so map it to the largest representable value instead of reading zero bytes.
+			if (!maxSymNameLen)
+				maxSymNameLen = UINT64_MAX;
+
+			// Every symbol table slot gets a data variable and a marker symbol below, so that
+			// relocations can resolve any symbol index the file declares. maxCoffSymbolCount only
+			// bounds how many of those slots also get full name resolution, typing, and aux
+			// record definitions, which are the more expensive per-symbol steps.
+			uint64_t symbolAnnotationLimit =
+				maxSymCount ? std::min<uint64_t>(header.coffSymbolCount, maxSymCount) : header.coffSymbolCount;
+
 			size_t symbolTableSize = header.coffSymbolCount * sizeofCOFFSymbol;
 			auto lastSection = m_sections.back();
 			symbolTableAdjustedOffset = header.coffSymbolTable - lastSection.pointerToRawData + lastSection.virtualAddress;
@@ -887,11 +915,11 @@ bool COFFView::Init()
 			DefineAutoSymbol(new Symbol(DataSymbol, "__symtab", coffSymbolTableBase, NoBinding));
 
 			BinaryReader stringReader(GetParentView(), LittleEndian);
-			uint64_t stringTableBaseRaw = header.coffSymbolTable + ((uint64_t) header.coffSymbolCount * sizeofCOFFSymbol);
+			uint64_t stringTableBaseRaw = header.coffSymbolTable + ((uint64_t)header.coffSymbolCount * sizeofCOFFSymbol);
 
 			stringReader.Seek(stringTableBaseRaw);
-			uint32_t stringTableSize = stringReader.Read32();
-			if ((stringTableBaseRaw + stringTableSize) > GetParentView()->GetEnd())
+			uint32_t stringTableSize;
+			if (!stringReader.TryRead32(stringTableSize) || (stringTableBaseRaw + stringTableSize) > GetParentView()->GetEnd())
 			{
 				throw COFFFormatException("invalid COFF string table size");
 			}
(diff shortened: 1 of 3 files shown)


view/pe/peview.cpp
```diff
@@ -1394,10 +1478,29 @@ bool PEView::Init()
 						stringReader.Seek(header.coffSymbolTable + (i * 18));
 						symbolName = stringReader.ReadCString(8);
 					}
-					else
+					else if (e_offset < stringTableLen)
 					{
-						stringReader.Seek(stringTableBase + e_offset);
-						symbolName = stringReader.ReadCString();
+						auto cached = symbolNameCache.find(e_offset);
```
zznop, line 1489: We should account for retained names on every symbol creation, including cache hits and check the projected allocation against the remaining 1 MiB limit.

With a 1 MiB limit, 300 symbols referencing one 4,096-byte name retains 1,228,800 bytes of raw names, while we track only 16,384 bytes

view/pe/coffview.cpp
```diff
@@ -972,68 +1029,74 @@ bool COFFView::Init()
...
 
-				if (e_sclass == IMAGE_SYM_CLASS_STATIC && e_value == 0)
+				if (annotate && e_sclass == IMAGE_SYM_CLASS_STATIC && e_value == 0)
 				{
 					size_t sectionHeaderOffset = sectionHeadersOffset + (e_scnum - 1) * sizeof(COFFSectionHeader);
 					(void)sectionHeaderOffset;
 					DEBUG_COFF(AddDataReference(m_imageBase + symbolVirtualAddress, m_imageBase + sectionHeaderOffset));
 				}
-				else if (e_sclass == IMAGE_SYM_CLASS_EXTERNAL && e_value == 0 && e_scnum == IMAGE_SYM_UNDEFINED)
+				else if (annotate && e_sclass == IMAGE_SYM_CLASS_EXTERNAL && e_value == 0 && e_scnum == IMAGE_SYM_UNDEFINED)
```
zznop, line 1147: Annotation limits discard relocations to undefined external COFF symbols. Creating a marker for a symbol-table entry preserves relocations to defined section symbols, but undefined exteranl relocations also require a matching ExternalSymbol (which is suppressed when annotation is disabled).

view/pe/peview.cpp
```diff
@@ -833,11 +844,44 @@ bool PEView::Init()
...
+					"rounding down to 0x%x per Windows loader behavior.",
+					i, section.pointerToRawData, section.pointerToRawData & ~(PE_SECTION_RAW_DATA_ALIGNMENT - 1));
+				section.pointerToRawData &= ~(PE_SECTION_RAW_DATA_ALIGNMENT - 1);
+			}
+			// Windows rounds SizeOfRawData up to the nearest FileAlignment multiple for PE32/PE32+.
+			// Without this, bytes between the raw value and the rounded value are invisible to
+			// analysis even though the Windows loader maps them. Skip this for low-alignment
+			// images for the same reason as the PointerToRawData rounding above.
+			// Cap at the remaining file bytes to avoid mapping data past the end of the file.
+			if (!lowAlignmentImage
```
zznop, line 864: Expanded PE sections make directory parsing disagree with the BinaryView's mapped bytes. Core mapping gives a later overlapping section precedence. RVAToFileOffset and GetRVACharacteristics instead choose the first matching section. Rounding can newly extend an earlier section over a directory belonging to a later section. We need to make RVA conversion and characteristics use the same effective mapping precedence as core reads.

view/pe/peview.cpp
```diff
@@ -1394,10 +1478,29 @@ bool PEView::Init()
...
-						stringReader.Seek(stringTableBase + e_offset);
-						symbolName = stringReader.ReadCString();
+						auto cached = symbolNameCache.find(e_offset);
+						if (cached != symbolNameCache.end())
+						{
+							symbolName = cached->second;
+						}
+						else if (!maxTotalSymNameBytes || totalSymNameBytesRead < maxTotalSymNameBytes)
+						{
+							stringReader.Seek(stringTableBase + e_offset);
```
zznop, line 1501: String-table bounds checks still allow names to consume bytes outside the table. We should require a table length of at least 4 bytes, validate payload offsets, and then cap reads to `min(configuredLimit, tableLength - offset)`

zznop (changes requested):

Weitao-Sun:
Fixed all four in bc9cad877: name-budget now charges every retained reference including cache hits; relocations to undefined-external symbols beyond the annotation limit are resolved on demand under the same budget; RVAToFileOffset/GetRVACharacteristics now keep the last matching section on overlap; string-table reads are bounds-checked and capped to the table's remaining length in all three lookup sites in these files (also found and fixed the same missing bounds check in the section-name lookup, which wasn't flagged but has the same issue).

view/pe/peview.cpp
Weitao-Sun: Fixed in bc9cad877 (peview.cpp:1505-1520, and the equivalent path in coffview.cpp). Every symbol that retains a name reference now computes its projected total -- including on a cache hit -- and charges it against the budget before the name is handed back to the caller, so N symbols sharing one string-table offset are bounded by N charges instead of 1. Verified locally: 300 symbols sharing a 4,096-byte name with a 1 MiB budget now resolve exactly 64 before the limit trips and the rest come back empty, on both files.

view/pe/coffview.cpp
Weitao-Sun: Fixed in bc9cad877 (coffview.cpp:1467-1505). The relocation loop still looks for an existing marker symbol first; when that lookup comes back empty because the entry was outside the annotation limit, it now re-reads that symbol-table slot directly, resolves its name, and creates the ExternalSymbol on demand, keyed and cached by symbol-table index so repeats are free. This goes through the same resolveSymbolName budget as the initial pass, so it can't be used to sidestep the byte cap. Verified locally: a relocation targeting a symbol table index beyond maxCoffSymbolCount now gets its ExternalSymbol and the relocation applies.

view/pe/peview.cpp
Weitao-Sun: Fixed in bc9cad877 (RVAToFileOffset / GetRVACharacteristics, peview.cpp:3645-3684). Both now iterate the full m_sections list and keep overwriting the result on every match instead of returning on the first one, so the last-added section wins when ranges overlap. Read through core/binaryview.cpp to confirm this actually matches core's behavior: MemoryRegion::DefaultOrderingStrategy sorts by orderId descending, GenerateRangeMap assigns orderId in the order ranges are passed in, and AddSegmentsToList appends new segments to the end without re-sorting -- so core does resolve overlapping mapped bytes in favor of the most recently added segment, which is what m_sections' insertion order already tracks. Verified locally with two fully-overlapping sections (earlier one executable, later one data-only): the export now resolves to the later section's name and gets classified as data, not a function.

view/pe/peview.cpp
Weitao-Sun: Fixed in bc9cad877. All three string-table-relative name lookups in these two files (COFF symbol names, PE symbol names, COFF section names) now require a table length of at least 4 bytes, reject offsets that fall inside that length field or past the table's end, and cap the read to tableLength - offset. The section-name lookup in coffview.cpp had the identical gap and wasn't called out here, so it got the same fix. Verified locally: an entry with no null terminator placed right at the table's declared end, followed by marker bytes past it, no longer pulls those marker bytes into the resolved name.

view/pe/coffview.cpp
```diff
@@ -690,6 +701,77 @@ bool COFFView::Init()
 	// The offset of the symbol table after adjusting for the alignment of the sections that precede it
 	uint64_t symbolTableAdjustedOffset = 0;
 
+	// Symbol-name resolution and its retained-bytes budget are shared between the initial
+	// pass over the symbol table below and the relocation pass further down, which needs to
```
zznop: COFF relocation markers bypass the total-name budget. So for example with the default 32 KiB name cap, 65,535 relocations would retain around 2 GiB of raw marker names (exceeding the 1 GiB budget).

I don't think this is worth addressing because this PR is already growing quite a bit and there's probably other ways to cause OOM with a malformed PE.

view/pe/peview.cpp
```diff
@@ -849,6 +893,14 @@ bool PEView::Init()
 			{
 				section.virtualSize = section.sizeOfRawData;
 			}
+			// Segments, sections, RVA characteristics, and symbol placement are all bounded by
+			// virtualSize elsewhere in this file, while file-backed reads are bounded by
+			// sizeOfRawData. Keep virtualSize at least as large as sizeOfRawData so a section
+			// whose raw data extends past its declared virtual size is still fully mapped.
+			if (section.sizeOfRawData > section.virtualSize)
```
zznop, line 900: Growing virtualSize to sizeOfRawData can show bytes beyond the rounded nonzero virtual region. For example, a VirtualSize of 0x100 and SizeOfRawData of 0x3000. The section will become 0x3000 bytes long. This is probably harmless, but looking at ReactOS's loader it constrains to a single page (making the mapped size 0x1000). Not sure what the best route is here or even if ReactOS's loader is to be trusted. But I think we probably just want to align the section size up to a page boundary instead of extending all the way to SizeOfRawData.

view/pe/peview.cpp
zznop, line 864: The new rounding expands an earlier section's raw-backed range into a later zero-fill section. The RVA helper then reads the earlier section's hidden file bytes, despite the BV mapping zeros there.

I think we want to fix the section-selection logic in RVAToFileOffset. select the last section covering the RVA by its virtual extent, then check whether that winning section has file backing at that RVA.

zznop (changes requested):
Getting closer. Just a couple issues. I'm less concerned about addressing more adversarial PE issues. I just want to make sure we're accurately mapping the binary into the binary view.

PR #8543, opened 2026-09-10
[aarch64] Disassemble and lift FEAT_CPA instructions

bdash:
`FEAT_CPA` (Checked Pointer Arithmetic) adds pointer arithmetic instructions that, on overflow out of the address bits, force bit 54 to disagree with bit 55. This makes the result non-canonical so it faults when dereferenced.

For readable decompilation these are lifted as if the instructions are supported but enforcement is disabled (i.e., as if the CPU supports `FEAT_CPA` but not `FEAT_CPA2`).


saagarjha (approved):
Checked against kernelcache.release.mac18g and these instructions seem to be lifted now as described

PR #8544, opened 2026-09-10
[Pseudo-ObjC] Render `@throw` statements

AngeloD2022:
Replaces calls to `objc_exception_throw` inserted by the compiler for `@throw` statements in Objective-C.

https://developer.apple.com/documentation/objectivec/objc_exception_throw(_:)

### DIFF
--- a/lang/c/pseudoobjc.cpp
+++ b/lang/c/pseudoobjc.cpp
@@ -126,7 +126,8 @@ struct RuntimeCall
 		Class,
 		Self,
 		RespondsToSelector,
-		IsKindOfClass
+		IsKindOfClass,
+		ExceptionThrow,
 	};
 
 	Type type;
@@ -152,6 +153,7 @@ constexpr std::array RUNTIME_CALLS = {
 	std::make_pair("_objc_retainAutoreleasedReturnValue", RuntimeCall::Retain),
 	std::make_pair("_objc_retainAutoreleaseReturnValue", RuntimeCall::RetainAutorelease),
 	std::make_pair("_objc_retainBlock", RuntimeCall::Retain),
+	std::make_pair("_objc_exception_throw", RuntimeCall::ExceptionThrow),
 	std::make_pair("j__objc_alloc_init", RuntimeCall::AllocInit),
 	std::make_pair("j__objc_alloc", RuntimeCall::Alloc),
 	std::make_pair("j__objc_autorelease", RuntimeCall::Autorelease),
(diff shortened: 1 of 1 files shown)


lang/c/pseudoobjc.cpp
```diff
@@ -278,6 +281,16 @@ void PseudoObjCFunction::GetExpr_CALL_OR_TAILCALL(const BinaryNinja::HighLevelIL
...
+	case RuntimeCall::ExceptionThrow:
+		// https://developer.apple.com/documentation/objectivec/objc_exception_throw(_:)
+		tokens.Append(KeywordToken, "@throw ");
+		if (parameterExprs.size() == 1)
+		{
+			GetExprText(parameterExprs[0], tokens, settings);
+			if (statement)
+				tokens.AppendSemicolon();
+			return;
+		}
```
bdash, line 292: This is missing a `break` and will fall through into the `RuntimeCall::Alloc` case below.

lang/c/pseudoobjc.cpp
```diff
@@ -278,6 +281,16 @@ void PseudoObjCFunction::GetExpr_CALL_OR_TAILCALL(const BinaryNinja::HighLevelIL
 			return;
 		}
 		break;
+	case RuntimeCall::ExceptionThrow:
+		// https://developer.apple.com/documentation/objectivec/objc_exception_throw(_:)
+		tokens.Append(KeywordToken, "@throw ");
+		if (parameterExprs.size() == 1)
```
bdash: It seems surprising to render this as `@throw ` if the parameter count is incorrect. I'd think it should continue to be rendered as a call if the arity doesn't match expectations.

lang/c/pseudoobjc.cpp
```diff
@@ -278,6 +281,16 @@ void PseudoObjCFunction::GetExpr_CALL_OR_TAILCALL(const BinaryNinja::HighLevelIL
...
+		{
+			GetExprText(parameterExprs[0], tokens, settings);
+			if (statement)
```
bdash: Does it even make sense to handle `objc_exception_throw` like this in an expression context? In Obj-C code, `@throw` is only valid as a statement. That makes me think we should skip rendering this as `@throw` if we're not in a statement context.

bdash (changes requested):
Thanks for the PR! There's a few issues that need to be addressed before it can be merged. Let me know if you'd prefer that I make the necessary changes.

lang/c/pseudoobjc.cpp
AngeloD2022: Good catch, I totally thought I put that under the condition.

lang/c/pseudoobjc.cpp
AngeloD2022: Good point; I don't think that would emerge with normal ObjC.

AngeloD2022:
Would you prefer I squash the commits?

bdash:
Yes please!

bdash (approved):
Thanks for the fixes! I'll get this merged tomorrow.

PR #8550, opened 2026-09-12
can't uninstall deprecated plugin fix 

### LINKED ISSUE #8545: cannot uninstall deprecated plugins
Post 6.0, the deprecation support unfortunately can put plugins in a state where they cannot be uninstalled. It's the one feature that should still be enabled for them!

UvuvDev:
API fix for deprecated plugins being impossible to uninstall from UI. Fixes https://github.com/Vector35/binaryninja-api/issues/8545

### DIFF
--- a/binaryninjaapi.h
+++ b/binaryninjaapi.h
@@ -20915,6 +20915,11 @@ namespace BinaryNinja {
 		bool Uninstall();
 		bool CancelUninstall();
 		bool Install(std::string versionID);
+		// Like Install, but also permits installing a version of a plugin that has since been marked
+		// deprecated. Intended only for migrating a plugin the user already had installed under a previous
+		// extension manager: deprecation should block fresh installs, but must never strand a
+		// previously-installed plugin where it can't be registered as installed (and so can't be uninstalled).
+		bool InstallForMigration(std::string versionID);
 		bool InstallDependencies();
 		bool InstallDependencies(const std::string& versionID);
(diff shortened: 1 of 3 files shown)


binaryninjaapi.h
```diff
@@ -20915,6 +20915,11 @@ namespace BinaryNinja {
 		bool Uninstall();
 		bool CancelUninstall();
 		bool Install(std::string versionID);
+		// Like Install, but also permits installing a version of a plugin that has since been marked
```
fuzyll, line 20918: Don't leave these AI-generated comments here. These don't do us any good. If this information is necessary (and I think it is, I would have no idea what this is for), we should have this documented in each externally-facing API we provide (e.g. in `pluginmanager.cpp`) so that it shows up in user-facing documentation.

pluginmanager.cpp
```diff
@@ -375,6 +375,15 @@ bool Extension::Install(std::string versionID)
 }
 
 
+bool Extension::InstallForMigration(std::string versionID)
```
fuzyll, line 378: This seems fine, but where are the Python and Rust equivalents? Also, see other note about moving documentation to here.

fuzyll (changes requested):
Assuming the internal PR is accepted, these are the changes I would want to this API one.

binaryninjaapi.h
UvuvDev: Comments moved to plugin manager.cpp

PR #8556, opened 2026-09-16
[ARMv7] Expand ARM and Thumb SIMD and VFP lifting

zznop:
This PR expands lifting for vector comparisons, bitwise operations, widening/narrowing moves, and more. Expansion includes all the ARMv7 instructions from this issue: https://github.com/Vector35/binaryninja/issues/1910

In addition to expanding lifting, I had to make fixes to the thumb-2 decoder. Here is a short summary:

- `vorr`/`vbic`/`vmvn` immediates: derive `.i16`/`.i32` types from `cmode` instead of register bits, immediate payload bits, or a constant
   - Ensures correct immediate replication during lifting
- `vmovl`: preserve the encoded signedness instead of unconditionally marking all widening moves as unsigned
- `vcge #0`: include the F bit when selecting the data type so `.f32` is distinguished from `.s32`
- `vmull.p64`: add the ARMv8 polynomial-multiple decoding
- Legacy VFP transfers: recognize odd-imm8 `fldm*x`/`fstm*x` encodings before selecting `vpop`/`vpush` aliases
   - Preserves the extra four bytes in addressing and writeback

I've uploaded my test binaries that I used for verification in addition to the binaries in this repository: https://github.com/Aarno-Labs/iot-binary-dataset (that are responsible for uncovering the issues)

[armv7-dh-lifting-expansion.zip](https://github.com/user-attachments/files/32307912/armv7-dh-lifting-expansion.zip)

Before:

Screenshot 2026-09-16 at 4 11 33 PM

After:

Screenshot 2026-09-16 at 4 11 58 PM


rssor (approved):

arch/armv7/armv7_disasm/armv7.c
```diff
@@ -5027,9 +5036,11 @@ uint32_t armv7_three_register_different(uint32_t instructionValue, Instruction*
 			break;
 		case 14: //op = 1
 			{
-				static DataType dataType[4] = {DT_P8, DT_P16, DT_P32, DT_NONE};
+				// ARMv8 crypto uses size=2 for a single 64x64 polynomial product.
+				if (decode.vcom.u != 0 || (decode.vcom.size != 0 && decode.vcom.size != 2) || (decode.vcom.vd & 1))
```
bdash, line 5042: I noticed that this correctly models the `if Vd[0] == '1' then Undefined(); end;` check in the decode pseudocode. The decode for the two other `ARMV7_VMULL` encodings, later in the file, don't have this check, despite their decode pseudocode having the same check.

zznop: Good catch, I updated the other two existing encodings with the same check

PR #8560, opened 2026-09-17
[PPC] Lift lswi/stswi as intrinsics and fix blrl return handling

zznop:
Adds PowerPC `lswi` and `stswi` lifting as memory intrinsics. Fixes `blrl` handling so GOT helpers return without decoding trailing data as code, while preserving ordinary indirect calls. I also started a PPC `test_lift.py` with regression tests for these instructions.

`blrl` fix before / after:

Screenshot 2026-09-17 at 2 34 58 PM

Screenshot 2026-09-17 at 2 34 25 PM

`lswi` and `stswi` lifting before / after:

Screenshot 2026-09-17 at 2 37 42 PM

Screenshot 2026-09-17 at 2 37 06 PM


rssor (changes requested):
I think there are actually issues that need fixing here, will need to dig a bit deeper. I don't think we can mark BLRL FunctionReturn that aggressively.

The TEMP thing makes me suspect this might warrant something in the core to help with conditional return analysis

zznop:
Updated the PR to directly lift `lswi` and `stswi` when they are used in isolation. But use an intrinsic and lift them together as a string copy when they are used consecutively (with the same registers) 

Screenshot 2026-09-18 at 10 01 15 AM

zznop:
> I think there are actually issues that need fixing here, will need to dig a bit deeper. I don't think we can mark BLRL FunctionReturn that aggressively.
> 
> The TEMP thing makes me suspect this might warrant something in the core to help with conditional return analysis

The `FunctionReturn` serves as a hint to the core's contextual-return analysis. This pattern is also used by ARM's `blx lr` instruction. The arch is lifting `blrl` as `LLIL_CALL`; it's not lifting it as a `LLIL_RET`.

The core initially treats it as a return, then checks the target through dataflow. If the target is anything other than `ReturnAddressValue` it restores call handling and reanalyzes the function to recover the fallthrough. The TEMP is needed to preserve the instructions _actual_ ordering. 

I get what you're saying in that there is no architectural guarantee that the callee can't branch back through the updated LR and take the fallthrough path, but I don't think that's likely in the case where the call target is `RetrunAddressValue`.

